PR #16, opened 2023-03-16
Update code of isValidHttpurl() in util.js

aasim-syed:
- [x] Updated Simplified version of the previous code of ```isValidHttpurl()```

We don't need to check the protocol property of the URL object explicitly, as the constructor automatically assigns a default value of "http:" to the protocol property if the input string doesn't specify one. Therefore, we can rely on the constructor to set the protocol property correctly and simply return true if no error occurs.


aasim-syed:
> Thanks for the PR! However, as stated in the stackoverflow post, your code will incorrectly match urls like `javascript:void(0)`
> 
> image
> 
> The function is only used when checking if we must use the fetch method (if file system disabled or it is an HTTP url) or whether we must load from a local file (in node)
> 
> So, is there a reason you would need to load non-HTTP urls?

Oh alrightyyy got it!

 Loved the repo imma keep tryna best to contribute 🚀🫡

PR #24, opened 2023-03-18
Update Util.js softmax() and log_softmax()

aasim-syed:
- [x] used a typed array for the exps ```Float64Array()``` array to save memory and potentially improve performance


xenova:
Just a note on the typed arrays:
using .map() on a typed array will return another typed array (of the same type). So, I don't imagine your approach will be much faster. If you think it will be, feel free to run some benchmarks. 👍

kungfooman:
PR also breaks `isValidHttpUrl` for no obvious reason:

image

Please keep PR's focused on one issue and explain why something needs to be changed.

xenova:
Closing since it breaks core functionality

PR #35, opened 2023-03-20
Smart execution providers

### LINKED ISSUE #30: Current use of execution providers is suboptimal
Currently the library only operates with the WASM backend even in nodejs. However the better approach should be use [node bindings](https://www.npmjs.com/package/onnxruntime-node) in node and add the `webgl` EP to automatically fallback to `wasm` if `webgl` is not supported by the model.

Currenly in code:
```js
let session = await InferenceSession.create(buffer, {
  // executionProviders: ["webgl"]
  executionProviders: ["wasm"]
});
```

Hack to at least improve in the web
```js
let session;

try {
    session = await InferenceSession.create(buffer, {
        executionProviders: ['webgl', 'wasm']
    });
} catch (err) {
    session = await InferenceSession.create(buffer, {
        executionProviders: ['wasm']
    });
}
```

I can prepare a PR to improve the execution in node and web 😃

DavidGOrtega:
- closes #30 

The purpose of this PR is:
 - Best effort to use the gpu over cpu if not fallbacks to wasm
 - Use `onnxruntime-node` in case is installed as dep in any node app using transformers.js (5x faster than WASM). If not installed it will falback to WASM provider.
 - Adds the hack for [this issue](https://github.com/microsoft/onnxruntime/issues/10311) setting `numThreads` to one 
 - Use onnx executorProviders fallback chain to improve faster inference in browser if the model supports it.
 - Wraps multiple `onnxruntime-web`  requirement dependancy  under one file

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,10 @@ env.remoteModels = false;
 
 // Set parent path of .wasm files. Defaults to use a CDN.
 env.onnx.wasm.wasmPaths = '/path/to/files/';
+
+#### Node.js
+
+This project uses `onnxruntime-web` as default backend. However if you add the `onnxruntime-node` dependency it will preferred over the web backend. `onnxruntime-node` is nearly 5X faster than WASM executor provider probably due to [this issue](https://github.com/microsoft/onnxruntime/issues/10311).
 ```
 
 ## Usage
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ env.onnx.wasm.wasmPaths = '/path/to/files/';
 
 #### Node.js
 
-This project uses `onnxruntime-web` as default backend. However if you add the `onnxruntime-node` dependency in your project it will preferred over the web backend. `onnxruntime-node` is nearly 5X faster than WASM executor provider probably due to [this issue](https://github.com/microsoft/onnxruntime/issues/10311).
+This project uses `onnxruntime-web` as default backend. However if you add the `onnxruntime-node` dependancy in your project it will preferred over the web backend. `onnxruntime-node` is nearly 5X faster than WASM executor provider probably due to [this issue](https://github.com/microsoft/onnxruntime/issues/10311).
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ env.onnx.wasm.wasmPaths = '/path/to/files/';
 
 #### Node.js
 
-This project uses `onnxruntime-web` as default backend. However if you add the `onnxruntime-node` dependency it will preferred over the web backend. `onnxruntime-node` is nearly 5X faster than WASM executor provider probably due to [this issue](https://github.com/microsoft/onnxruntime/issues/10311).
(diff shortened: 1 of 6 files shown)


DavidGOrtega:
@xenova This should do the trick. I need to test the web and check that the fallback is proceeding fine.

README.md
xenova, line 1: dependency is correct :)

src/tensor_utils.js
xenova, line 1: Does the execution provided switch to wasm if the webgl backend fails? If so, then this is alright. If not, then I am slightly worried about using backends (webgl/cuda/webgpu) that do not fully support the necessary operations.

Can you confirm?

src/env.js
xenova, line 1: I see you're accessing ONNX through the tensor utils file, but I think it would be better if we create a separate file (e.g., backend.js or onnx.js) which handles the loading and fallbacks of the various imports.

xenova (changes requested):
Looks good overall - just some questions about how fallbacks are handled and organization details

src/tensor_utils.js
DavidGOrtega: ONNX will fallback to the next one if its fails. However this behaviour is flaky and for that I have not included all backends. With the ones selected we should be ok.

DavidGOrtega:
@xenova Im not totally happy with the PR unless we remove all the backends and allow the user to install `node` bindings, which are much faster than wasm.

Also the fallback is not working if the model do not support a layer 🤦 

I think that we should provide a way to expose the desired  executor.
So I would change the PR to do this.

```js
let ONNX;
let executionProviders = [ 'wasm' ];

try {
    ONNX = require('onnxruntime-node');
    executionProviders = [ 'cuda', 'cpu' ];
} catch (err) {
    ONNX = require('onnxruntime-web');
    if(typeof process === 'object') {
        // https://github.com/microsoft/onnxruntime/issues/10311
        ONNX.env.wasm.numThreads = 1;
    }
}
```

With the code below we have at least fixed a rough edge and allow the user to use node bindings if desired.

What do you think?

xenova:
Yeah I agree 👍 Once WebGPU releases, I'll be more focused on getting GPU support working (both for node and browser).

I'll merge this main into this PR, make some edits, then merge the PR back into main (hopefully soon haha). Thanks again for your contributions!

xenova:
I ran some tests to see what kind of speedup these changes make, and it's amazing!

| Task                           | Speedup |
| ------------------------------ | ------: |
| Text classification            | 1400%   |
| Question answering             | 600%    |
| Image-to-text                  | 400%    |
| Text-to-text generation        | 350%    |
| Code generation                | 350%    |
| Embeddings                     | 325%    |
| Masked language modelling      | 300%    |
| Translation                    | 225%    |
| Summarization                  | 200%    |
| Zero-shot image classification | 200%    |
| Image classification           | 175%    |
| Text generation                | 150%    |

🎉 

I'm doing some final merging and will hopefully get it published soon :)

xenova:
@DavidGOrtega Can you grant me write access to your fork? I would like to push the changes without having to create a new fork and make a PR. (I think you can just add me as a collaborator?)

xenova:
Got it working, PR merged! 🎉 Thanks again for your contributions!

PR #46, opened 2023-03-22
Smart execution providers (Merges #35 into main)

xenova:
Since I don't have write access to the smart-execution fork by @DavidGOrtega, I had to create a new branch from his fork.


xenova:
> Great work getting this to run in node! Works super well.

Glad to hear! Next step regarding node is to implement a decent disk-caching system (like how HF's python library handles downloading new models)

PR #51, opened 2023-03-23
added DIRNAME_AVAILABLE check in env for electron

biw:
Hello, thanks for a great package! In electron on the web process `const fs = require('fs');` returns an object:
```
{ default: { /** empty */ } }
```
so it exists but cannot be used. I added another check:
```
const DIRNAME_AVAILABLE = typeof __dirname === 'string'
```

I could also test for some functions used like [`fs. existsSync()`](https://github.com/xenova/transformers.js/blob/fa0eb38c4966272480a62d5b090e052b3d86bbcd/src/utils.js#L12) if you'd prefer to do that over checking for `__dirname`. Let me know!

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -5,7 +5,7 @@ const { env: onnx_env } = require('./backends/onnx.js').ONNX;
 
 // check if various APIs are available (depends on environment)
 const CACHE_AVAILABLE = typeof self !== 'undefined' && 'caches' in self;
-const FS_AVAILABLE = !isEmpty(fs); // check if file system is available
+const FS_AVAILABLE = !isEmpty(fs) && typeof fs.existsSync === 'function'; // check if file system is available
 const PATH_AVAILABLE = !isEmpty(path); // check if path is available
 
 const RUNNING_LOCALLY = FS_AVAILABLE && PATH_AVAILABLE;


xenova:
Is this check not sufficient:
https://github.com/xenova/transformers.js/blob/e628c0a957d4b40810e73ea50a786142327998a2/src/env.js#L8

which calls `isEmpty` (checks if the object contains at least one key):
```js
function isEmpty(obj) {
    return Object.keys(obj).length === 0;
}
```

biw:
From https://github.com/electron/electron/issues/2288#issuecomment-337858978, it looks like the user agent might be the best place to detect it:
```
var userAgent = navigator.userAgent.toLowerCase();
if (userAgent.indexOf(' electron/') > -1) {
   // Electron-specific code
}
```

however, if I understand correctly, this has the disadvantage that if `nodeIntegration` is enabled for an electron app, they may still be able to use `fs`.

xenova:
> From https://github.com/electron/electron/issues/2288#issuecomment-337858978, it looks like the user agent might be the best place to detect it:
> 
> ```
> 
> var userAgent = navigator.userAgent.toLowerCase();
> 
> if (userAgent.indexOf(' electron/') > -1) {
> 
>    // Electron-specific code
> 
> }
> 
> ```
> 
> 
> 
> however, if I understand correctly, this has the disadvantage that if `nodeIntegration` is enabled for an electron app, they may still be able to use `fs`.

Ahh yes, good point. Perhaps it might be best to simply just check if `fs` has one of its methods (like `existsSync`), as you mentioned above.

How would one use the filesystem if `nodeIntegration` is disabled?

biw:
> Ahh yes, good point. Perhaps it might be best to simply just check if fs has one of its methods (like existsSync), as you mentioned above.

Will do now

> How would one use the filesystem if nodeIntegration is disabled?

They would need to use the `contextBridge` and then call it in node from their main electron process or they'd need to use a web worker on the electron web process

xenova:
I wrote an electron application to make sure the library works, and didn't run into this issue. This might be because I was running from a web worker (which is recommended to avoid blocking the UI). See here: https://github.com/xenova/transformers.js/tree/main/examples/electron

Closing the issue for now.

PR #55, opened 2023-03-24
Add Logits Processors

spencekim:
This is still a WIP as I need to clean up the code and formatting, but opening for comments as a quick sanity check.

@xenova from what I understand the `input_ids` parameter is a flattened array where you can infer the shape by reading from `logits.dims[0]`?

Let me know if I'm on the right track.

### DIFF
--- a/src/generation.js
+++ b/src/generation.js
@@ -207,13 +207,38 @@ class NoRepeatNGramLogitsProcessor extends LogitsProcessor {
     );
     for (const [hypo, bannedTokens] of bannedBatchTokens.entries()) {
       for (const token of bannedTokens) {
-        console.log("banned token", token);
         logits.data[hypo * logits.dims[1] + token] = -Infinity;``
       }
     }
     return logits;
   }
 }
+
+
+class RepetitionPenaltyLogitsProcessor extends LogitsProcessor {
+  constructor(penalty) {
+    super();
+    this.penalty = penalty;
+  }
+
+  _call(input_ids, logits) {
+      const numBatches = logits.dims[0]
+      const batchSize = logits.dims[1]
+      for (let batch = 0; batch < numBatches; ++batch) {
+          for (const inputId of input_ids) {
+              const index = batch * batchSize + inputId
+              const currentScore = logits.data[index]
+              if (currentScore < 0) {
+                  logits.data[index] = currentScore * this.penalty
+              } else {
+                  logits.data[index] = currentScore / this.penalty
+              }
+          }
+      }
+      return logits
+    }
+}
+
 
 class GenerationConfig {
     constructor(kwargs = {}) {
@@ -286,4 +311,5 @@ module.exports = {
     WhisperTimeStampLogitsProcessor,
     ForceTokensLogitsProcessor,
     NoRepeatNGramLogitsProcessor,
+    RepetitionPenaltyLogitsProcessor
 };
--- a/src/generation.js
+++ b/src/generation.js
@@ -135,6 +135,86 @@ class WhisperTimeStampLogitsProcessor extends LogitsProcessor {
     }
 }
 
+const getNgrams = (ngramSize, prevInputIds, numHypos) => {
+  const curLen = prevInputIds.length / numHypos;
+  const generatedNgrams = [];
+  for (let i = 0; i < numHypos; ++i) {
+    const genTokens = prevInputIds.slice(i * curLen, (i + 1) * curLen);
+    const ngrams = [];
+    for (let j = 0; j < curLen + 1 - ngramSize; j++) {
+      const ngram = [];
+      for (let k = 0; k < ngramSize; ++k) {
+        ngram.push(genTokens[j + k]);
+      }
+      ngrams.push(ngram);
+    }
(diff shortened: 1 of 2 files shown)


xenova:
Just one quick comment - it's quite difficult to see what was changed when the whole file changes due to some formatting. Can you try match the source formatting? It's probably just your prettier, and we are working to get a standard format setup throughout the repo.

xenova:
> @xenova from what I understand the `input_ids` parameter is a flattened array where you can infer the shape by reading from `logits.dims[0]`?

I'd have to double check this, but I think that both `input_ids` and `logits` are flat 1D arrays.
```js
    _call(input_ids, batchedLogits) {
        // Apply logits processor to each item in the batch:
        for (let logits of batchedLogits) {
            // Modifies logits inplace
            this.processors.forEach(
                func => func(input_ids, logits)
            )
        }
    }
```

The main reason for doing this is because JS doesn't support vectorized operations like numpy, so, we have to perform it to each batch separately.

src/generation.js
```diff
@@ -135,6 +135,86 @@ class WhisperTimeStampLogitsProcessor extends LogitsProcessor {
...
+
+class NoRepeatNGramLogitsProcessor extends LogitsProcessor {
+  constructor(no_repeat_ngram_size) {
+    super();
+    this.no_repeat_ngram_size = no_repeat_ngram_size;
+  }
+
+  _call(input_ids, logits) {
+    const numBatchHypotheses = logits.dims[0];
+    const curLen = input_ids.length / numBatchHypotheses;
```
xenova, line 201: `curLen` is just `input_ids.length` (see comment on main thread for why)

xenova (commented):
Looks like a pretty decent translation from python code.

I'll do proper testing tomorrow.

src/generation.js
```diff
@@ -135,6 +135,86 @@ class WhisperTimeStampLogitsProcessor extends LogitsProcessor {
...
+  const batchBannedTokens = [];
+  if (curLen + 1 < ngramSize) {
+    for (let i = 0; i < numHypos; ++i) {
+      batchBannedTokens.push([]);
+    }
+  } else {
+    const generatedNgrams = getNgrams(ngramSize, prevInputIds, numHypos);
+
+    for (let i = 0; i < numHypos; ++i) {
+      const hypoPrevInputIds = prevInputIds.slice(i * curLen, (i + 1) * curLen);
```
spencekim, line 180: > curLen is just `input_ids.length` (see comment on main thread for why)

Given what you said ^, this line will have to change to just `prevInputIds` without slicing into it. I was referencing this line in the HuggingFace implementation: https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L468

spencekim: also on this thread, I'm trying to implement the EncoderNoRepeatNGramLogitsProcessor with this change in mind.

Looking at this line in the HuggingFace implementation: https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L530
what would the `self.batch_size` be if we're only getting a flat array?

spencekim: ok, I have an attempt at NoRepeatNGramLogitsProcessor and RepetitionPenaltyLogitsProcessor

xenova (commented):
Nice! This is a big one.

Looks good for now :+1: I will do a thorough review once near the end of this week (as I am currently quite busy).

Thanks again!

src/generation.js
```diff
@@ -207,14 +207,39 @@ class NoRepeatNGramLogitsProcessor extends LogitsProcessor {
...
+class RepetitionPenaltyLogitsProcessor extends LogitsProcessor {
+  constructor(penalty) {
+    super();
+    this.penalty = penalty;
+  }
+
+  _call(input_ids, logits) {
+      const numBatches = logits.dims[0]
+      const batchSize = logits.dims[1]
+      for (let batch = 0; batch < numBatches; ++batch) {
```
xenova, line 227: I'll double check my original implementation, but if I remember correctly, you shouldn't need to iterate over batches (this is is handled by the LogitsProcessorList, which, when called, does the batching there.

This is because JS doesn't allow vectorized operations.

src/generation.js
```diff
@@ -207,14 +207,39 @@ class NoRepeatNGramLogitsProcessor extends LogitsProcessor {
...
+          for (const inputId of input_ids) {
+              const index = batch * batchSize + inputId
+              const currentScore = logits.data[index]
+              if (currentScore < 0) {
+                  logits.data[index] = currentScore * this.penalty
```
xenova, line 232: Will just have to check whether score is in log probability format or not. This might change to + and -, depending on that.

xenova:
While I was busy merging this into main, I discovered a nasty generation bug in my original code (essentially, in the first step, it would not just look at the last set of logits). Fortunately, your PR helped find it since you were looping over the batches, which confused me because I thought I moved the batching into the `LogitsProcessorList` class! 🥳 

So, I'll hopefully be merging everything (including the bug fix) into main soon.

xenova:
Merged 🙌  Thanks again for the PR!

PR #72, opened 2023-04-06
Define ReadableStream from node:stream/web if it doesn't exist

ekolve:
`ReadableStream` is not defined as a global under Node 16 (this causes an error within the FileResponse constructor), this pulls the class in from `node:stream/web`.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -3,7 +3,7 @@ const fs = require('fs');
 
 const { env } = require('./env.js');
 
-if (global.ReadableStream  === undefined && typeof process !== 'undefined') {
+if (global.ReadableStream === undefined && typeof process !== 'undefined') {
     try {
         global.ReadableStream  = require('node:stream/web').ReadableStream; // ReadableStream is not a global with Node 16
     } catch(err) {
--- a/src/utils.js
+++ b/src/utils.js
@@ -3,6 +3,14 @@ const fs = require('fs');
 
 const { env } = require('./env.js');
 
+if (global.ReadableStream  === undefined && typeof process !== 'undefined') {
+    try {
+        global.ReadableStream  = require('node:stream/web').ReadableStream; // ReadableStream is not a global with Node 16
+    } catch(err) {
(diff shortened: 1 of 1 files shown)


src/utils.js
```diff
@@ -3,6 +3,14 @@ const fs = require('fs');
 
 const { env } = require('./env.js');
 
+if (global.ReadableStream  === undefined && typeof process !== 'undefined') {
```
xenova, line 6: Does this work in other environments with process defined? Like electron?

xenova:
Thanks for the PR! Can you also provide the error message you received (if any)?

src/utils.js
xenova, line 6:
```suggestion
if (global.ReadableStream === undefined && typeof process !== 'undefined') {
```

xenova (changes requested):

ekolve:
This is the error message I received when running under Node 16.20.0:

```
ReferenceError: ReadableStream is not defined
    at new FileResponse (node_modules/@xenova/transformers/src/utils.js:27:29)
    at getFile (node_modules/@xenova/transformers/src/utils.js:177:16)
    at getModelFile (node_modules/@xenova/transformers/src/utils.js:231:26)
    at fetchJSON (node_modules/@xenova/transformers/src/utils.js:288:24)
    at Function.from_pretrained (node_modules/@xenova/transformers/src/models.js:2143:28)
    at pipeline (node_modules/@xenova/transformers/src/pipelines.js:1224:24)
```

src/utils.js
ekolve: I'm not sure about whether this would work under electron - since the require is performed in a try/catch it shouldn't cause a failure if for some reason the package is not available.

ekolve: I just did a quick test of a simple electron app and everything worked fine.

xenova: Sweet!

xenova (approved):

chelouche9:
@xenova works like a charm :)

PR #76, opened 2023-04-08
[docs] Automatically generate documentation from JSDoc

xenova:
Making a draft PR for comments and suggestions!


package.json
```diff
@@ -11,7 +11,8 @@
     "typegen": "tsc ./src/transformers.js --outDir dist/types --allowJs --declaration --emitDeclarationOnly",
     "dev": "webpack serve --no-client-overlay",
     "test": "node tests/index.js",
-    "build": "npm run webpack && npm run typegen"
+    "build": "npm run webpack && npm run typegen",
+    "docs": "npx jsdoc-to-markdown --files ./src/**.js"
```
kungfooman, line 15: 1) I was running into caching issues, it just wouldn't update (`--clear` clears the cache).
2) Creating `docs.md`, so the console isn't spammed

```suggestion
    "docs": "npx jsdoc-to-markdown --clear && npx jsdoc-to-markdown --files ./src/**.js > docs.md"
```

kungfooman:
Currently the output looks like:

image

I dislike the dot syntax in the so called "type application": `Promise.<Array.<unknown>>`

Should we just remove the dots (they are applied in both `jsdoc` and `catharsis` package)?

And then about the output... right now it's just one big file, does it need to be many smaller files? I couldn't find an option for that, probably that would require some API scripting.

xenova:
> I dislike the dot syntax in the so called "type application": Promise.<Array.<unknown>>
> Should we just remove the dots (they are applied in both jsdoc and catharsis package)?

Agreed. Not too sure why they do that. Apparently it's an older syntax. That sounds like the easiest solution, yeah.

> And then about the output... right now it's just one big file, does it need to be many smaller files? I couldn't find an option for that, probably that would require some API scripting.

I was looking at this yesterday, and I also couldn't find a way to do it from the CLI. Do you think we should do it on a file basis or a class basis? I'm trying to make it similar to [huggingface.js' docs](huggingface.js' docs), which splits each class into a separate file.

kungfooman:
> Do you think we should do it on a file basis or a class basis? I'm trying to make it similar to [huggingface.js' docs](https://github.com/huggingface/huggingface.js/tree/main/docs), which splits each class into a separate file.

Hm, many projects already assume that one file = one class, but I guess you want to keep them all in one huge file?

Per-class-separation sounds better to me too.

Also worked out the PR that at least `catharsis` is not adding its strange syntax: https://github.com/xenova/catharsis/pull/1

xenova:
> but I guess you want to keep them all in one huge file?

For the most part, I'd actually prefer it to be separated... especially for the really large files. What about something like this:

docs/
 - index.md
 - _toctree.yml
 - file1.js/
   - classes/
     - class1.md
     - class2.md
     - ///
   - functions/
     - ...
 - file2.js/
 - ...

takoyaro:
> For the most part, I'd actually prefer it to be separated... especially for the really large files. What about something like this:
> 
> docs/
> * index.md
> * _toctree.yml
> * file1.js/
>   * classes/
>     * class1.md
>     * class2.md
>     * ///
>   * functions/
>     * ...
> * file2.js/
> * ...
> 
> Not too sure if "functions"/"classes" are the way to break it up more though.

I personally prefer that approach as well. Keeping this library's docs consistent with the HF docs will be better for the Dev experience overall.

takoyaro:
A lot is going on here, should I wait for this to be merged before putting more time on #91 ?

xenova:
> A lot is going on here, should I wait for this to be merged before putting more time on #91 ?

Haha 😅 ... yeah this PR is getting quite big 😬  

It will probably be best to wait, otherwise there will be a ton of conflicts :') I'm hoping I can get everything done, and then merged this weekend.

xenova:
> A lot is going on here, should I wait for this to be merged before putting more time on #91 ?

You can continue from this branch. All that needs to be done is get it working in the browser :) For some reason, I'm unable to import the script from a `Worker` (of type module).

I'll continue working on it tomorrow though.

PR #81, opened 2023-04-11
fix: reference the types in package json

chelouche9:


chelouche9:
@xenova added the types to the package json file so the types could be used.

package.json
```diff
@@ -1,8 +1,9 @@
 {
-  "name": "@xenova/transformers",
+  "name": "@chelouche9/transformers",
```
takoyaro, line 2:
```suggestion
  "name": "@xenova/transformers",
```

chelouche9: THank you for noticing

xenova:
Is there a way to do this that allows users to use `npm link`? If I'm not mistaken, this will cause an issue since the types won't be updated?

chelouche9:
> Also, this main.d.ts file won't be updated each time we build.

That is right, the main.d.ts file is not updated. I will try to find a solution that won't require us to change the build process.

xenova:
Changes live on main 👍

PR #91, opened 2023-04-16
Es6 rewrite

takoyaro:
That's just a WIP, tackling the issues I can, one at a time. Contributions welcome

Also, someone else mentioned using prettier the other day, I'm aware this should be a different PR but I think it'd make contributions easier for everybody, including this one.

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -1,6 +1,6 @@
-import fs from 'fs';
-import path from 'path';
-import { fileURLToPath } from 'url';
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 
 import { ONNX } from './backends/onnx.js';
 const { env: onnx_env } = ONNX;
--- a/src/image_utils.js
+++ b/src/image_utils.js
@@ -1,5 +1,5 @@
 
-import fs from 'fs';
+import fs from 'node:fs';
 import { getFile, isString } from './utils.js';
 import { env } from './env.js';
 
@@ -8,7 +8,7 @@ let ImageClass = typeof Image !== 'undefined' ? Image : null;
 
 let ImageDataClass;
 let loadImageFunction;
-if (typeof self !== 'undefined') {
+if (typeof OffscreenCanvas !== 'undefined') {
     CanvasClass = OffscreenCanvas;
     loadImageFunction = self.createImageBitmap;
     ImageDataClass = ImageData;
@@ -18,6 +18,7 @@ if (typeof self !== 'undefined') {
     CanvasClass = Canvas;
     loadImageFunction = async (/**@type {Blob}*/ b) => await loadImage(Buffer.from(await b.arrayBuffer()));
     ImageDataClass = ImageData;
+    // @ts-ignore
     ImageClass = Image;
 }
 
--- a/src/transformers.js
+++ b/src/transformers.js
@@ -29,7 +29,6 @@ export {
 
 // environment variables
 export { env } from './env.js';
-
 // other
 export {
     pipeline
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,5 +1,5 @@
 
-import { existsSync, statSync, promises } from 'fs';
+import { existsSync, statSync, promises } from 'node:fs';
 
 import { env } from './env.js';
 


kungfooman:
> That's just a WIP, tackling the issues I can, one at a time. Contributions welcome

You can just mark it as draft and concentrate on real info?

> Also, someone else mentioned using prettier the other day, I'm aware this should be a different PR but I think it'd make contributions easier for everybody, including this one.

This sentence is highly confusing. I don't know what you wanted to say, but please don't overlap PR's with different concerns (linting + ES6 modules + ...), it just makes it cumbersome to review.

takoyaro:
> You can just mark it as draft and concentrate on real info?

I'm not really familiar with contributions, doing the best I can. I converted the PR to draft for the time being.
Can you explain what you mean by "_concentrate on real info_" ?

> This sentence is highly confusing. I don't know what you wanted to say, but please don't overlap PR's with different concerns (linting + ES6 modules + ...), it just makes it cumbersome to review.

Sorry about my poor wording and english. I got it, will be careful.

kungfooman:
> Can you explain what you mean by "_concentrate on real info_" ?

Yes, just some explanations why and what you did:
 - Why replace `fs` with `node:fs`, is it just the linter you wrote about?
 - What does this PR fix?
   - Does Webpack build work now?
   - Does it work via import-maps without a bundler?
   - Does it play nice with ES5 ONNX?
   - Does it work in node?
 - To what extend did you test it?
 - Which issues were you running into (which you mentioned in the first post)?

kungfooman:
`src/transformers.js` can be simplified into... not sure if it needs comments any longer, because it feels self-explaining:

```
export * from './tokenizers.js';
export * from './models.js';
export * from './processors.js';
export * from './env.js';
export * from './pipelines.js';
export * from './tensor_utils.js';
```

And usually that is called the entry point, named `src/index.js`.

It simply exports everything, which I like for debugging. I rewrote @do-me's SemanticFinder into ES6 using ES6 transformers.js and it just makes it nice to quickly test around with:

image

kungfooman:
Can you also rewrite:

```js
if (global.ReadableStream === undefined && typeof process !== 'undefined') {
    try {
        // @ts-ignore
        global.ReadableStream = (await import('node:stream/web')).ReadableStream; // ReadableStream is not a global with Node 16
    } catch (err) {
        console.warn("ReadableStream not defined and unable to import from node:stream/web");
    }
}
```

into code which works in a browser + node at the same time (simply replacing `global` with `globalThis`):

```js
if (!globalThis.ReadableStream && typeof process !== 'undefined') {
    try {
        // @ts-ignore
        globalThis.ReadableStream = (await import('node:stream/web')).ReadableStream; // ReadableStream is not a global with Node 16
    } catch (err) {
        console.warn("ReadableStream not defined and unable to import from node:stream/web");
    }
}
```

This kind of code is also called a polyfill and is often separated into e.g. `polyfills/ReadableStream.js` and included in `index.js`:

```js
import 'polyfills/ReadableStream.js';
export * from './tokenizers.js';
export * from './models.js';
export * from './processors.js';
export * from './env.js';
export * from './pipelines.js';
export * from './tensor_utils.js';
```

takoyaro:
I was planning to get back on it this weekend...not sure why you speak to me with this tone but I hope you're ok.

kungfooman:
> I was planning to get back on it this weekend...not sure why you speak to me with this tone but I hope you're ok.

Thank you! Unfortunately no, I'm not, I'm sitting on a burning stone flying with high-speed through the universe and people have no other issues than "increasing casualties" of each other.

I just want to check-in with you since I heard nothing from you for some days and people open issues etc. I'm sorry for the tone and can you forgive me?

xenova:
You should be able to continue from the latest version of the [docs branch](https://github.com/xenova/transformers.js/tree/docs). It seems to work in node.js, but doesn't work in the browser for some reason. Still working that out. It's quite later here, so, I'll try figure it out tomorrow :)

xenova:
So you're suggesting the problem is occurring in https://github.com/xenova/transformers.js/blob/docs/src/backends/onnx.js? I'll look into it.

kungfooman:
Yep, I just found this issue, they seem to have it on their radar:

https://github.com/microsoft/onnxruntime/issues/14529

And their ES6 issue: https://github.com/microsoft/onnxruntime/issues/10913

The ES6 issue has been open for a year, but eventually it should be fixed. Pretty much two options:

1) Fork / fix it ourselves - possibly creating a PR which could be merged, while using repo-fork in the meantime
2) Wait till the devs realize that a name like `ort.es6.min.js` should be a proper ES6 module with an export 😅

@fs-eire Do you happen to have any ETA for ES6 support? I mean... it kinda works, just not in the way ES6 wants it (using `globalThis.ort`).

Also ES6 standard is to name it `ort.min.mjs` (`mjs` meaning an ES6 module)

fs-eire:
sorry, I don't have an ETA for ES6 yet. If you would like to contribute, it would be great!

The major reason is that we did a lot of "optimizations" for the file size of ort.min.js (UMD), and those are not compatible with webpack with ES6 format. Those including:
- duplicated contents from
    - ort-wasm.js vs. ort-wasm-simd.js 
    - ort-wasm-threaded.js vs. ort-wasm-simd-threaded.js 
    - ort-wasm-threaded.worker.js vs. ort-wasm-simd-threaded.worker.js 
- non-standard way dealing inline worker for proxy feature to avoid CORS issue
- non-standard way dealing inline threaded worker for wasm multithreading to avoid CORS issue
- not able to tree-shaking in ES6 as all imports are top layer, as we also deploy trimed files like ort.webgl.min.js

I actually worked on this last month, but it turns out to be much more difficult than expected. I may come back some day to revisit this issue.

kungfooman:
> sorry, I don't have an ETA for ES6 yet. If you would like to contribute, it would be great!

Thank you for the quick response! It feels like you are drowning in complexity there. ES6 should be as simple as renaming `*.ts` to `*.js` and using:

```js
import * from './tensor.js';
```
Instead of:
```js
import * from './tensor';
```

This is what every modern browser speaks natively.

You can keep the non-standard-ways for CORS etc. to build a bundle while not messing with TypeScript and their AST rewriting first aswell? The interface types just go into `*.d.ts` and normal types go into JSDoc (which exist anyway for documentation).

Without payment I cannot deep-dive into this, even though I would like to take on this challange. This is also something that probably doesn't take weeks, but maybe two months aswell to get everything right, figured out and tested in multiple browsers/environments etc.

fs-eire:
> > sorry, I don't have an ETA for ES6 yet. If you would like to contribute, it would be great!
> 
> Thank you for the quick response! It feels like you are drowning in complexity there. ES6 should be as simple as renaming `*.ts` to `*.js` and using:
> 
> ```js
> import * from './tensor.js';
> ```
> 
> Instead of:
> 
> ```js
> import * from './tensor';
> ```
> 
> This is what every modern browser speaks natively.
> 
> You can keep the non-standard-ways for CORS etc. to build a bundle while not messing with TypeScript and their AST rewriting first aswell? The interface types just go into `*.d.ts` and normal types go into JSDoc (which exist anyway for documentation).
> 
> Without payment I cannot deep-dive into this, even though I would like to take on this challange. This is also something that probably doesn't take weeks, but maybe two months aswell to get everything right, figured out and tested in multiple browsers/environments etc.

I did some study and find that it seems no way for me to create a NPM package which allows user to import in both ESM/CJS way:

```ts
// this is Typescript code
import { a, b, b } from 'onnxruntime-common';
```

It generates error like this:
> //The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("onnxruntime-common")' call instead.
 >  To convert this file to an ECMAScript module, change its file extension to '.mts', or add the field `"type": "module"` to 'package.json'.

[This article](https://evertpot.com/universal-commonjs-esm-typescript-packages/) works for JavaScript but still so far no way to make it works for TypeScript.

xenova:
Hi @fs-eire - Yes, this has also been a massive headache for me too (see [dev branch](https://github.com/xenova/transformers.js/tree/docs)). Fortunately, for each of my use-cases, I have found a workaround so that ES6 modules work in node... which was way more difficult the other way around (getting node modules working in ES6).

Perhaps @kungfooman can give some advice in this regard, as I am way less experienced haha.

fs-eire:
> Hi @fs-eire - Yes, this has also been a massive headache for me too (see [dev branch](https://github.com/xenova/transformers.js/tree/docs)). Why doesn't `"type": "module"` work?
> 
> Fortunately, for each of my use-cases, I have found a workaround so that ES6 modules work in node... which was way more difficult the other way around (getting node modules working in ES6).
> 
> Perhaps @kungfooman can give some advice in this regard, as I am way less experienced haha.

`"type": "module"` works. However this is not what I want. converting an existing JS/TS project from commonjs to module may take a lot of efforts. My goal is to make `onnxruntime-web` able to be imported by either ESM/CJS modules.

I already make `onnxruntime-common` a ESM/CJS dual package, with a very tricky hack.  The article I mentioned [above](https://evertpot.com/universal-commonjs-esm-typescript-packages/) works by setting up CJS/ESM source codes with duplicated types declarations in different folders, which containing a package.json with different type.

This is because TypeScript tries to determine whether the current code is CJS/ESM by looking for a nearest "package.json" file, and the solution is to prepare such a file to make it happy. Not sure whether TypeScript will change this behavior in future, but it works for now at least.

The change is here: https://github.com/microsoft/onnxruntime/pull/15772, which is far more complicated than just appending the ".js". And I expect doing that for `onnxruntime-web` need take a while....

kungfooman:
> [This article](https://evertpot.com/universal-commonjs-esm-typescript-packages/) works for JavaScript but still so far no way to make it works for TypeScript.

> The change is here: [microsoft/onnxruntime#15772](https://github.com/microsoft/onnxruntime/pull/15772), which is far more complicated than just appending the ".js". And I expect doing that for `onnxruntime-web` need take a while....

From your referenced article:

image

Once you realize that a project/language basically actively blocks you, it's time to identify it as technical debt and get rid of it. I used TypeScript for years, but it became so annoying that I just don't want to touch it anymore (in its "pure" form with all the peculiar design decisions of TS purists who basically dislike JS while barely knowing it).

I take the good parts (the type system with turing complete types, the TS language server) and `*.d.ts` files while otherwise keeping everything `*.js` with JSDoc. It just makes everything a tad easier.

I will probably not convince you, but I think it should be pointed out (before more people apply these kinds of `find/xargs/sed` rewrite hacks or referring to `*.ts` files as `*.js` for example).

The TypeScript stance doesn't even make sense... they happily introduced enums that transpile to different JS code or public-constructor-assign-AST rewrites, but adding an import-rewrite-feature requested by all kinds of people is ideologically blocked for purity reasons... oh well, glad I'm outta there 😂

fs-eire:
@kungfooman I know that Typescript has some limitations, but I don't want to jump into the debate of whether or not to use Typescript in my project. The important thing is, as a library developer, I want to make sure my library can be consumed by both JavaScript & Typescript users. Even if I write pure JavaScript with hand-written .d.ts in my package, I still need to deal with the problems and do the same hack to make sure TypeScript users to consume the package happily.

kungfooman:
@fs-eire No problem, lets cease that debate. Do you still run into problems with that PR? I'm not on the right computer for testing your PR, but the way you are using `"exports"` field in `package.json` is exactly how it works for me aswell. But you mentioned the ugly hack, which I don't need to make it work for me.

My `playcanvas` package simply contains this in `package.json`:

```js
  "exports": {
    "import": "./build/playcanvas.mjs/index.mjs",
    "require": "./build/playcanvas.js"
  },
```

Testing with node:

image

Funny thing is that it only works for me when the files actually are named `*.mjs`, which I wrote a quick/simple PHP script for:

```php

<?php
  function renameFilesIn($here) {
    $files = glob($here);
    $n = count($files);
    print("rename files in $here n=$n\n");
    foreach ($files as $file) {
        $mjs_file_name = str_replace(".js", ".mjs", $file);
        rename($file, $mjs_file_name);
        print("$file -> $mjs_file_name\n");
    }
  }
  renameFilesIn("**/*.js");
  renameFilesIn("**/**/*.js");
  renameFilesIn("**/**/**/*.js");
  renameFilesIn("*/*/*/*/*.js");
  renameFilesIn("*/*/*/*/*/*.js");
  renameFilesIn("*/*/*/*/*/*/*.js");
?>

```

And the ES5 version is simply generated from the ES6 files (using `rollup`, but you can use whatever you want).

I'm not sure what your current problems are or if you solved it so far. As test I simply made `src/test.ts`:

```ts
import { Entity } from "playcanvas";
export function lol() {
  const entity = new Entity("test");
  console.log("lol entity name:", entity.name);
}
lol();
```

JS code is simply generated using `npx tsc` and runs fine:

image

So if you summarize your current state that would be helpful and I applaud and congratulate you for all of your efforts 👍 💯 🥇

fs-eire:
using .mjs / .cjs is a good option, but that turns out to be not compatible with some tools (eg. Clang-format). Clang-format only supports `.mjs .js .ts` for JavaScript language. (see [doc](https://clang.llvm.org/docs/ClangFormat.html)). Need to use the "--assumeFileName" flag to trick clang-format otherwise it treats .mts/.cts/.cjs as C++ source code. I added a PR to the VSCode extension but seems no response yet: https://github.com/xaverh/vscode-clang-format/pull/150

The lucky thing is that, the one-liner package.json is working. The PR should be good to consume by TS/JS & CJS/ESM users.

kungfooman:
> The lucky thing is that, the one-liner package.json is working. The PR should be good to consume by TS/JS & CJS/ESM users.

Glad to hear that 👍 I tested in node and nodes module resoulation respects it well. And for browsers people should use import-maps anyway, specifying the ES6 path. Some people of course also prefer to create giant files which contain everything and that should work aswell via `webpack`/`rollup`.

> Clang-format only supports `.mjs .js .ts` for JavaScript language.

Why do you need `.mts`? Wouldn't it be possible to just stick to `.ts` and apply your `$importpath` -> `$importpath.mjs` TypeScript hack? It should also be possible to just apply that to the distributed "dist"/"build" folder and keep everything else as it is.

kungfooman:
Thanks to all the work of @xenova in the docs branch, we can load transformers.js and onnxruntime in a browser like this now:

```js
<body></body>
<script>
  function importFile(content) {
    return "data:text/javascript;base64," + btoa(content);
  }
  const imports = {
    "transformers": "./src/transformers.js",
    "fs": importFile("export default {};"),
    "url": importFile("export default {};"),
    "path": importFile("export default {};"),
    "stream/web": importFile("export default {};"),
    "sharp": importFile("export default {};"),
    "onnxruntime-node": importFile("export default {};"),
    "onnxruntime-web": importFile(`
      await import("https://cdnjs.cloudflare.com/ajax/libs/onnxruntime-web/1.14.0/ort.es6.min.js");
      let ONNX = globalThis.ort;
      export default ONNX;
      export {
        ONNX
      };
    `),
  };
  const importmap = document.createElement("script");
  importmap.type = "importmap";
  importmap.textContent = JSON.stringify({imports});
  document.body.appendChild(importmap);
</script>
<script type="module">
  import * as transformers from "./src/transformers.js";
  Object.assign(window, {...transformers});
</script>
<audio id="SPEECH2TEXT_AUDIO" src="./examples/demo-site/assets/audio/jfk.wav" controls="true"></audio>
```

To test, you can just hack code into F12/devtools, e.g. Whisper example:

```js
const pipe = await pipeline("automatic-speech-recognition");
const audioCTX = new AudioContext({
sampleRate: 16000
});
const arrayBuffer = await (await fetch(SPEECH2TEXT_AUDIO.currentSrc)).arrayBuffer();
const decoded = await audioCTX.decodeAudioData(arrayBuffer);
const audio = decoded.getChannelData(0);
const result = await pipe(audio);
console.log("result", result);
```

Result:

image

Once onnxruntime is a proper ES6 module, we can simplify the import map code for it, but otherwise it works nice for a quick dev cycle already.

xenova:
Great! I also think I've found a way to do it without import maps - will keep you updated :)

xenova:
Changes live on main. 👍

PR #103, opened 2023-05-05
Update models.js, replace python ??= with javascript ||

cpravetz:
models.js had a python logical operator, and this replaces it with equivalent javascript.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -1601,7 +1601,7 @@ class WhisperForConditionalGeneration extends WhisperPreTrainedModel {
 
 
         // Whisper has additional options for returning timestamps
-        generation_config.return_timestamps ??= false;
+        generation_config.return_timestamps = generation_config.return_timestamps || false;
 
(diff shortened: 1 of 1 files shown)


xenova:
This is intended. `??` is the null-coalescing operator, which "returns the first operand if it exists and is not null or undefined; otherwise, it returns the second operand" (see [here](https://www.tutorialspoint.com/What-is-a-null-coalescing-operator-in-JavaScript#:~:text=The%20null%20coalescing%20operator%20(%3F%3F,it%20returns%20the%20second%20operand.))

`||` is less-strict, and will return the second operator if the first is "falsy". For example:
```js
console.log(0 ?? 1);  // prints 0
console.log(null ?? 1); // prints 1
console.log(undefined ?? 1); // prints 1

console.log(0 || 1);  // prints 1
console.log(null || 1);  // prints 1
console.log(undefined || 1);  // prints 1

```

PR #106, opened 2023-05-07
Fix tensor normalization in EmbeddingsPipeline

chrislee973:
I was directly using the quantized all-MiniLM-L6-v2 onnx model to precompute some embeddings from Python, and I noticed some slight differences between those embeddings and the ones transformers.js generated. I found that the vector norm calculation wasn't squaring the first value of the vector: ```let norm = Math.sqrt(batch.data.reduce((a, b) => a + b * b))```.

Adding an initial value of 0 to the ```reduce``` fixes this. I've verified that this fixed normalization calculation results in embeddings equal to the ones generated directly from the onnx model!

### DIFF
--- a/src/pipelines.js
+++ b/src/pipelines.js
@@ -628,7 +628,7 @@ class EmbeddingsPipeline extends Pipeline {
      */
     _normalize(tensor) {
         for (let batch of tensor) {
-            let norm = Math.sqrt(batch.data.reduce((a, b) => a + b * b))
+            let norm = Math.sqrt(batch.data.reduce((a, b) => a + b * b, 0))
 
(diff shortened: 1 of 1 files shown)


kungfooman:
Simple test for validating the difference:

```js
function aPlusBsquared(a, b) {
  console.log({a, b});
  return a + b * b;
}
[10, 20, 30, 40].reduce(aPlusBsquared);
[10, 20, 30, 40].reduce(aPlusBsquared, 0);
```

Output:

image

Basically when no initial value is given, it doesn't start with `undefined`, but using index 0 and 1 instead.

xenova:
Whoops! Thanks for that!

PR #112, opened 2023-05-16
[Whisper] Implement/use getMelFilters

kungfooman:
Attempting to fix https://github.com/xenova/transformers.js/issues/107

Testing:

```js
url = "https://huggingface.co/Xenova/whisper-tiny.en/raw/main/preprocessor_config.json";
res = await fetch(url);
config = await res.json();
original_mel_filters = config.mel_filters;
calculated_mel_filters = getMelFilters(config.sampling_rate, config.n_fft, config.feature_size);
differences = [];
for (var i=0; i<80; i++) {
  for (var j=0; j<201; j++) {
    const expected = original_mel_filters[i][j];
    const calculated = calculated_mel_filters[i][j];
    if (expected !== calculated) {
        const diff = Math.abs(expected - calculated);
        differences.push({i, j, expected, calculated, diff});
    }
  }
}
console.table(differences);
```

There are differences, but they are all minuscule:

image

However, based on real transcribing, these differences do matter quite a lot... so this is a draft right now, until we figure out how to fine-tune the calculation to be exactly like the Python ones.

### DIFF
--- a/src/processors.js
+++ b/src/processors.js
@@ -38,6 +38,7 @@ import {
 import { Tensor, transpose, cat, interpolate } from './utils/tensor.js';
 
 import { CustomImage } from './utils/image.js';
+import { getMelFilters } from './utils/audio.js';
 
 
 /**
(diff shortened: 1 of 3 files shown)


xenova:
Thanks for this! Those outputs are well-within the necessary bounds. For example, ONNX exports only require floating point precision within 1e-5... and yours are within 1e-9.

xenova:
Also - could I have write access? I'll add some unit tests and fix some of the type annotations (jsdoc is complaining lol)

PR #123, opened 2023-05-23
Returning local filename in node.js in getModelFile

dakenf:
A slight optimization to pass filename instead of Buffer to onnxruntime in node.js

### DIFF
--- a/src/utils/hub.js
+++ b/src/utils/hub.js
@@ -1,7 +1,7 @@
 
 /**
  * @file Utility functions to interact with the Hugging Face Hub (https://huggingface.co/models)
- * 
+ *
  * @module utils/hub
  */
 
@@ -18,7 +18,7 @@ if (!globalThis.ReadableStream) {
 }
 
 /**
- * @typedef {Object} PretrainedOptions Options for loading a pretrained model.     
+ * @typedef {Object} PretrainedOptions Options for loading a pretrained model.
  * @property {boolean?} [options.quantized=true] Whether to load the 8-bit quantized version of the model (only applicable when loading model files).
  * @property {function} [options.progress_callback=null] If specified, this function will be called during model construction, to provide the user with progress updates.
  * @property {Object} [options.config=null] Configuration for the model to use instead of an automatically loaded configuration. Configuration can be automatically loaded when:
@@ -154,7 +154,7 @@ class FileResponse {
     /**
      * Reads the contents of the file specified by the filePath property and returns a Promise that
      * resolves with a parsed JavaScript object containing the file's contents.
-     * 
+     *
      * @returns {Promise<Object>} A Promise that resolves with a parsed JavaScript object containing the file's contents.
      * @throws {Error} If the file cannot be read.
      */
@@ -243,7 +243,7 @@ function handleError(status, remoteURL, fatal) {
 class FileCache {
     /**
      * Instantiate a `FileCache` object.
-     * @param {string} path 
+     * @param {string} path
      */
     constructor(path) {
         this.path = path;
@@ -251,7 +251,7 @@ class FileCache {
 
     /**
      * Checks whether the given request is in the cache.
-     * @param {string} request 
+     * @param {string} request
      * @returns {Promise<FileResponse | undefined>}
      */
     async match(request) {
@@ -268,8 +268,8 @@ class FileCache {
 
     /**
      * Adds the given response to the cache.
-     * @param {string} request 
-     * @param {Response|FileResponse} response 
+     * @param {string} request
+     * @param {Response|FileResponse} response
      * @returns {Promise<void>}
      */
     async put(request, response) {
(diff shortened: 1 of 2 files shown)


src/utils/hub.js
```diff
@@ -428,15 +428,19 @@ export async function getModelFile(path_or_repo_id, filename, fatal = true, opti
...
-    const buffer = await readResponse(response, data => {
-        dispatchCallback(options.progress_callback, {
-            status: 'progress',
-            ...data,
-            name: path_or_repo_id,
-            file: filename
+    // in node.js it is inefficient to read big model files into memory and pass to ONNX runtime,
+    // so we'll just return filename
+    let buffer = response.filePath
+    if (!process.env) {
```
kungfooman, line 434:
```suggestion
    if (!globalThis.process?.env) {
```
A browser doesn't expose `process`, so accessing it directly will throw an Error via ES6 `<source type="module">`

kungfooman:
Is this ready for review or a draft? Right now it crashes for me in node.js:

```js
> const { pipeline } = await import("./src/transformers.js");
```
```js
> const pipe = await pipeline("automatic-speech-recognition", "Xenova/whisper-tiny");
```

image

dakenf:
sorry, my bad. is there an easy way to run tests for the browser?

kungfooman:
> sorry, my bad. is there an easy way to run tests for the browser?

Sorry, not that I know, I just tested this PR against the code that I use anyway:

**test.html**

```html
<body></body>
<script>
  function importFile(content) {
    return "data:text/javascript;base64," + btoa(content);
  }
  const imports = {
    "transformers": "../transformers.js_es6/src/transformers.js",
    "fs": importFile("export default {};"),
    "url": importFile("export default {};"),
    "path": importFile("export default {};"),
    "stream/web": importFile("export default {};"),
    "sharp": importFile("export default {};"),
    "onnxruntime-node": importFile("export default {};"),
    "onnxruntime-web": importFile(`
      await import("https://cdnjs.cloudflare.com/ajax/libs/onnxruntime-web/1.14.0/ort.es6.min.js");
      //http://127.0.0.1/transformer/onnxruntime/js/common/dist/
      //await import("http://127.0.0.1/transformer/onnxruntime/js/common/dist/ort-common.js");
      //await import("http://127.0.0.1/transformer/onnxruntime/js/web/dist/ort-webgl.js");
      //await import("https://cdnjs.cloudflare.com/ajax/libs/onnxruntime-web/1.14.0/ort.es6.min.js");
      let ONNX = globalThis.ort;
      export default ONNX;
      export {
        ONNX
      };
    `),
  };
  const importmap = document.createElement("script");
  importmap.type = "importmap";
  importmap.textContent = JSON.stringify({imports});
  document.body.appendChild(importmap);
</script>
<script type="module">
  import * as transformers from "transformers";
  Object.assign(window, {...transformers});
</script>
<!--
  <audio id="SPEECH2TEXT_AUDIO" src="./examples/demo-site/assets/audio/jfk.wav" controls="true"></audio>
  <audio id="SPEECH2TEXT_AUDIO" src="./examples/demo-site/assets/audio/minner_fra_krigen.wav" controls="true"></audio>
-->
<audio id="SPEECH2TEXT_AUDIO" src="./audio/minner_fra_krigen_short.wav" controls="true"></audio>

```

And then you can just test around via F12/devtools for quick testing:

```js
const pipe = await pipeline("automatic-speech-recognition", "Xenova/whisper-tiny");
const audioCTX = new AudioContext({
  sampleRate: 16000
});
const arrayBuffer = await (await fetch(SPEECH2TEXT_AUDIO.currentSrc)).arrayBuffer();
const decoded = await audioCTX.decodeAudioData(arrayBuffer);
const audio = decoded.getChannelData(0);
  return_timestamps: true,
  //chunk_length_s: 30,
  chunk_callback: console.log
});
console.log("result", result);
```

I hope that helps, quick testing without any build tool etc.

dakenf:
@kungfooman i've fixed cache behaviour in node and browser

and got to an unrelated issue. i've fixed it by replacing "wasm" backend to "cpu" in node env, looks like you had something similar. related to loading wasm files. i've double-checked it first by removing node_modules (to see if it had my patched runtime) but it did not help
image

dakenf:
here you go. benchmarks look great!

dakenf:
These changes are going to be obsolete as I'm going to add passing the Response object directly to ONNX runtime, so it can read data with chunks from web/cache directly into webassembly memory. With current implementation it allocates memory for the model twice and UInt8Array cannot hold >4gb of data, so closing for now and will get back to it after ONNX pull request is merged

xenova:
> These changes are going to be obsolete as I'm going to add passing the Response object directly to ONNX runtime, so it can read data with chunks from web/cache directly into webassembly memory.

Awesome! Does the current version of ORT support passing response objects, or is that something you will add in your fork? If it's your fork, maybe the microsoft folks would be interested in adding support for it!

dakenf:
> Does the current version of ORT support passing response objects, or is that something you will add in your fork?

In mine. I've created a draft pull request to have some discussion. It cannot be merged and built yet because only 2 of 4 of my emscripten (webassembly compiler toolchain) pull requests are merged

xenova:
Gotcha 👌 You're doing so much amazing work! 🔥 Just out of curiosity, do you know if there's a way to skip this duplication of memory, not by passing a response object, but simply using the same array buffer (or uint8array)? I find it very strange that this isn't handled correctly by ORT itself.

dakenf:
There is no way of doing that, because WASM has a separate isolated memory (basically an array buffer). You can check how I do it now https://github.com/dakenf/onnxruntime/blob/0ddc05b978c42c1e29868239297266d5e9a579f0/js/web/lib/wasm/wasm-core-impl.ts#L86

So here can be only two ways: passing a response/some kind of readable stream or exporting WASM memory from ORT runtime but I don't think maintainers would agree to do that

> You're doing so much amazing work!

I just wanted to generate some pictures in the browser 😄 But after realizing its not possible yet, I decided to see how far I can go to fix it in my free time. For now its very interesting and would be a great case for resume

PR #160, opened 2023-06-21
Set chunk timestamp to rounded time

### LINKED ISSUE #159: [Bug] Chunk Timestamp [1] is not rounded
**Describe the bug**

While using **Whisper Web** and trying to export to JSON i have encountered a bug.
The timestamp[1] does not have proper rounder values like the timestamp[0].

I have verified this is not a bug due to exporting to JSON, I believe that wrong variable is used when setting chunk.timestamp[1] 

Please check this line [here](https://github.com/xenova/transformers.js/blob/480417118044f407bd5f0feca9008d1f31e82888/src/tokenizers.js#L2615)
```js
chunk.timestamp[1] = time;
```

This should be equal to  `rounded_time`
```diff
- chunk.timestamp[1] = time;
+ chunk.timestamp[1] = rounded_time;
```

**How to reproduce**

You can reproduce this bug by using the default audio link provided and then exporting to JSON

[Website](https://huggingface.co/spaces/Xenova/whisper-web)
[Audio](https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/ted_60.wav)

**Expected behavior**

timestamp[1] should have rounder values like the timestamp[0]

**Logs/screenshots**

image

**Environment**

- Transformers.js version: 2.2.0

**Additional context**

I have tried to make changes to the code.
```diff
- chunk.timestamp[1] = time;
+ chunk.timestamp[1] = rounded_time;
```
This change does fix the problem.

image

You can assign this bug to me, and I can raise a PR with this fix.

PushpenderSaini0:
Fixes #159

### DIFF
--- a/src/tokenizers.js
+++ b/src/tokenizers.js
@@ -2612,7 +2612,7 @@ export class WhisperTokenizer extends PreTrainedTokenizer {
                             // This is an issue in the underlying model output
                             // Let's just skip it so it becomes
                         } else {
-                            chunk.timestamp[1] = time;
+                            chunk.timestamp[1] = rounded_time;
 
                             // Handling merges
                             previous_tokens.push(current_tokens)


xenova:
Thanks again!

PR #169, opened 2023-06-29
[Doc] misc doc suggestions, feel free to cherry pick the commits you want and skip the ones you don't like!

julien-c:
main suggestions are to document a bit:
- how to find pretrained checkpoints on the Hub
- how to convert your model to ONNX (more simple) if you want to


xenova:
Amazing! Thanks so much @julien-c! 🤗

julien-c:
awesome, thanks! for the conversion Space i do think we should push it more / add more features to it

This would 🚀 the number of supported models quite a lot

Or maybe we do a `spaces/xenova/convert-to-transformers.js-onnx` custom Space tailored for this?

PR #174, opened 2023-06-30
Update onnx.js

lsb:
I tried out bert base cased using wasm simd on a 16.5.1 iOS device and I got the desktop result: https://jsfiddle.net/3uakbhes/

### DIFF
--- a/src/backends/onnx.js
+++ b/src/backends/onnx.js
@@ -43,7 +43,7 @@ if (typeof process !== 'undefined' && process?.release?.name === 'node') {
     // SIMD for WebAssembly does not operate correctly in some recent versions of iOS (16.4.x).
     // As a temporary fix, we disable it for now.
     // For more information, see: https://github.com/microsoft/onnxruntime/issues/15644
-    const isIOS = typeof navigator !== 'undefined' && /iP(hone|od|ad) OS 16_4_/.test(navigator.userAgent);
+    const isIOS = typeof navigator !== 'undefined' && /iP(hone|od|ad).+16_4.+AppleWebKit/.test(navigator.userAgent);
     if (isIOS) {
         ONNX.env.wasm.simd = false;
     }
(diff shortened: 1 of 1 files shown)


lsb:
Hm, let me test further. I haven't seen any release notes that detailed. I'm imagining that this turns into something like the WebGL conformance test for making sure that models executing in CPU mode run approximately equal to executing in WASM/WebGL/WebGPU modes

lsb:
Not sure about ORT 1.15 on iOS 16.4.x. ORT 1.15 on iOS 16.5.1 works, not only with bert-base-cased-web but also with the sentence transformer all-minilm-l6-v2

xenova:
Did a bit of testing and it looks like you're right! It's working now 👀. According to [this blog post](https://webkit.org/blog/14154/webkit-features-in-safari-16-5/#:~:text=Fixed%20WASM%20SIMD%20breaking%20WebP%20decoding%20applications), there was a bug with "Fixed WASM SIMD" which they mention has been fixed.

src/backends/onnx.js
```diff
@@ -40,10 +40,10 @@ if (typeof process !== 'undefined' && process?.release?.name === 'node') {
     // Running in a browser-environment
     ONNX = ONNX_WEB.default ?? ONNX_WEB;
 
-    // SIMD for WebAssembly does not operate correctly in recent versions of iOS (>= 16.4).
+    // SIMD for WebAssembly does not operate correctly in some recent versions of iOS (16.4.x).
     // As a temporary fix, we disable it for now.
     // For more information, see: https://github.com/microsoft/onnxruntime/issues/15644
-    const isIOS = typeof navigator !== 'undefined' && /iP(hone|od|ad)/.test(navigator.userAgent);
+    const isIOS = typeof navigator !== 'undefined' && /iP(hone|od|ad) OS 16_4_/.test(navigator.userAgent);
```
xenova, line 46: Are you 100% sure this matches iPod and iPad user agents? I have tested with my iPhone and it matches that, but a quick search (https://developer.apple.com/forums/thread/119186) suggests the that other devices won't match.

xenova, line 46:
```suggestion
    const isIOS = typeof navigator !== 'undefined' && /iP(hone|od|ad).+16_4.+AppleWebKit/.test(navigator.userAgent);
```

What do you think about this?

PR #195, opened 2023-07-16
Fixed an issue with an error during pipeline execution

tosh99:
The error I am getting is the following - 

```
ReferenceError: Response is not defined
 at getModelFile /node_modules/@xenova/transformers/src/utils/hub.js:462:42
```

The environment I am using is NodeJS v16.15.0 with TypeScript with the default CommonJS configuration. 

This Commit fixes the issue.

### DIFF
--- a/src/utils/hub.js
+++ b/src/utils/hub.js
@@ -459,7 +459,7 @@ export async function getModelFile(path_or_repo_id, filename, fatal = true, opti
         }
 
 
-        if (cache && response instanceof Response && response.status === 200) {
+        if (cache && response.status === 200) {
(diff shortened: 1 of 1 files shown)


xenova:
This unfortunately breaks the caching system as `FileResponse` objects will now also be cached (leading to duplicated storage). According to the 16.15.0 (LTS) [release notes](https://nodejs.org/en/blog/release/v16.15.0), you should be able to resolve this by adding the `--experimental-fetch` flag. Alternatively, you can update to a more up-to-date version of NodeJS (18+) or use a polyfill

tosh99:
Thanks for the clarification 👍

PR #242, opened 2023-08-08
Improve onnx conversion instructions in readme

do-me:
- Creation and activation of venv onnxconversion
- Installation of requirements.txt
- Short explanation of huggingface URL model identifiers

I thought a short note on how to use the conversion script might be useful for non-Python users. I used venv (and not e.g. conda) as it's built-in.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -446,10 +446,8 @@ async function encoderForward(self, model_inputs) {
     for (let key of self.session.inputNames) {
         encoderFeeds[key] = model_inputs[key];
     }
-    if (self.session.inputNames.includes('token_type_ids')) {
-        const token_type_ids = model_inputs.attention_mask.clone();
-        token_type_ids.data.fill(0n);
-        encoderFeeds.token_type_ids = token_type_ids
+    if (!encoderFeeds.token_type_ids && self.session.inputNames.includes('token_type_ids')) {
+        add_token_types(encoderFeeds);
     }
     return await sessionRun(self.session, encoderFeeds);
 }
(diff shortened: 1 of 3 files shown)


xenova:
Thanks for the PR! I think it's useful to mention the requirements in the requirements.txt, but I'm a little hesitant to increase the complexity otherwise (since it should be up to the user whether they want to create a venv). Also, the `/bin/activate` is only available on linux/mac. Afaik, it's `/Scripts/activate.exe` on windows, so if we will add something like this, we'd have to account for both cases...

I think a better option would be to just create a HF space for this (which is something I've just put off doing since I haven't really had time yet).

do-me:
I'm not too familiar with the concept of HF spaces but if it offers space for some more detailed explanation why not. Feel free to edit as you need or close this PR!

On Windows it should be: 

`onnxconversion/scripts/activate`

(last time I actually used venv on Windows) 

But there might be more ways to activate it. Usually I try to avoid Windows and instead prefer WSL (if I'm forced to work on a Windows machine) as it's less error-prone. I can verify if you need, just let me know.

src/models.js
```diff
@@ -446,6 +446,11 @@ async function encoderForward(self, model_inputs) {
     for (let key of self.session.inputNames) {
         encoderFeeds[key] = model_inputs[key];
     }
+    if (self.session.inputNames.includes('token_type_ids')) {
+        const token_type_ids = model_inputs.attention_mask.clone();
+        token_type_ids.data.fill(0n);
+        encoderFeeds.token_type_ids = token_type_ids
+    }
```
xenova, line 453: To avoid overwriting existing `token_type_ids`, you should add a check here:

```suggestion
    if (!encoderFeeds.token_type_ids && self.session.inputNames.includes('token_type_ids')) {
        add_token_types(encoderFeeds);
    }
```

You can import the `add_token_types` function from [tokenizers.js](https://github.com/xenova/transformers.js/blob/b8719b12dd7b61f99340a7d27e32f105538daf2e/src/tokenizers.js#L2581)

xenova (changes requested):
Thanks for the PR. I see you reused an existing PR, which is fine (can you just update the title?)

do-me:
Of course, when pushing to my main branch I didn't consider the other open PR. Also, feel free to close this one here too to unclutter the PRs. 

I'll go ahead now and create a PR with a different branch, that's probably the easiest solution.

PR #261, opened 2023-08-25
Rename how-to guides to developer guides

perborgen:
There was an inconsistency in the naming, as it said "HOW-TO GUIDES" in the paragraph whereas they are actually called "DEVELOPER GUIDES" in the sidebar.


xenova:
Thanks! 🤗

PR #278, opened 2023-09-03
Make // @ts-ignore obsolete for _call overrides by respecting LSP

kungfooman:
Fixes https://github.com/xenova/transformers.js/issues/275

Right now we violate the Liskov Substitution Principle:

> Simply put, the Liskov Substitution Principle (LSP) states that objects of a superclass should be replaceable with objects of its subclasses without breaking the application. In other words, what we want is to have the objects of our subclasses behaving the same way as the objects of our superclass.

Minified example:

```js
class Callable {
    /**
     * This method should be implemented in subclasses to provide the
     * functionality of the callable object.
     *
     * @param {any[]} args
     * @throws {Error} If the subclass does not implement the `_call` method.
     */
    _call(...args) {
        throw Error('Must implement _call method in subclass')
    }
}

class Pipeline extends Callable {
    /**
     * Executes the task associated with the pipeline.
     * @param {any} texts The input texts to be processed.
     * @returns {Promise<any>} A promise that resolves to an array containing the inputs and outputs of the task.
     */
    async _call(texts) {
        // ...
    }
}

class QuestionAnsweringPipeline extends Pipeline {
    /**
     * Executes the question answering task.
     * @param {string|string[]} question The question(s) to be answered.
     * @param {string|string[]} context The context(s) where the answer(s) can be found.
     * @param {Object} options An optional object containing the following properties:
     * @param {number} [options.topk=1] The number of top answer predictions to be returned.
     * @returns {Promise<any>} A promise that resolves to an array or object containing the predicted answers and scores.
     */
    async _call(question, context, {
        topk = 1
    } = {}) {
        // ...
    }
}

```

[TS Playground](https://www.typescriptlang.org/play?ts=4.6.2&filetype=js#code/MYGwhgzhAEDCYnAIxAU2gbwFDV9A9AFSE57SHQAqAFgJYwC2qALtQPYAm0E7AriFyTpaDAA5omAO2aoutSd15JQkCKhjM20UQCc2AN1od0rVKTwUAZr0nBmtNpIS1mAT2htL0U9GAJkaB5IAFaodgB05rgkZOTQAAKiYDpgDJhgkq4A2gC6AL7QyQDmEFFx8ax6AO4wGACiOno6BQCSXj4QSipQ0Bxs6tCSbMzQIuKoUiM+AAYA+n6I09BMrJyRsYT4UfP+ABThB8UQAJSYZXiVbFXQDU27AOQAsrwQI2MSqNLQO4jLLOxyBSdZTgKD3Y5RPJYKFYbowAAKtFEqBA8nQqAAHjJJBwYPBEGAUOhsGQiDEyBQ6hiwrwZBpqCZIABrQpQNjAWhgGRcKouajeBnaJEotHrCkJJIpNIYDKuAoyLEwGjCSSiWneTHMDRaITaPTAdRqDhiiwJHQsXg6SS1eF6Bj0VAAHllAD4CgBBPVse1qAVc6DmiBsED6AaaQoKZIpdzARzMMDyeRFAUqtVaiNcNi0tMwTwp7zMk3RLZkSCuWzfBYgXYKrWnEmxXD4fDQA5FqEwuHQACKvHU9kc7utVVQOiTiORqMk6Kxn1x0AnIunZ1JxDKlOpwFpYcFAEc+68HJHh6OkwWIEyi+VJalMK8x5IigAfe9J3IFff9o9UPcHgeSXYTm8HV0AyCAR3NY11wlZJbwwV9HxfZgHyKd9fDjTUf3QWNpE1QDTiqBlzXzMCIPw3wMmgXVLCzHErwoRJYOlAB5EIwmYAo2FEf8YCHDxuKPBAglCOx0OkBNJDPHwaMQK4z10LjR3sdQAC5oMYqVMEkXgGCEZpoCyLiePCTRRCZABeABGHIsMGHS9I8douIjcDRz1WRaDsI9tSo9BzWYS1pygjYzQtK0bTtB1nUyN1oE9BSfRMah-UDYNQx8yiozAdw2B0YT2LE+NE0ffNdA8uxZBciCYAyLgIFjQN6JLPAywrH5q0-Q9HAAGkKzVeobRtTJZczoEsyFoFGjA8nrc4mxbNtIWhLAgA)

Solution? We just need default values of the extra arguments (e.g. `context` for `QuestionAnsweringPipeline#call` to make the function signatures compatible. Should work like that with the other cases too... only task is: picking sensible default values

So now we can run it without context:

image

However, I could also imagine this being a nice solution:

image

Basically just `context = 'missing context'` default argument. Any thoughts / reflections?

### DIFF
--- a/src/pipelines.js
+++ b/src/pipelines.js
@@ -308,7 +308,7 @@ export class TokenClassificationPipeline extends Pipeline {
  */
 
 /**
- * @typedef {Promise<QuestionAnsweringResult|QuestionAnsweringResult[]|undefined>} QuestionAnsweringReturnType
+ * @typedef {Promise<QuestionAnsweringResult|QuestionAnsweringResult[]>} QuestionAnsweringReturnType
  */
 
 /**
@@ -334,8 +334,8 @@ export class QuestionAnsweringPipeline extends Pipeline {
      * @param {string|string[]} context The context(s) where the answer(s) can be found.
      * @param {Object} options An optional object containing the following properties:
      * @param {number} [options.topk=1] The number of top answer predictions to be returned.
-     * @returns {QuestionAnsweringReturnType} A promise that resolves to an array or object containing the
-     * predicted answers and scores. Can also return undefined in case of `topk = 1`.
+     * @returns {QuestionAnsweringReturnType} A promise that resolves to an array or object
+     * containing the predicted answers and scores.
      */
     async _call(question, context = '', {
         topk = 1
--- a/src/pipelines.js
+++ b/src/pipelines.js
@@ -302,6 +302,16 @@ export class TokenClassificationPipeline extends Pipeline {
 }
 
 /**
+ * @typedef {object} QuestionAnsweringResult
+ * @property {string} answer - The answer.
+ * @property {number} score - The score.
+ */
+
+/**
+ * @typedef {Promise<QuestionAnsweringResult|QuestionAnsweringResult[]|undefined>} QuestionAnsweringReturnType
+ */
+
+/**
  * Question Answering pipeline using any `ModelForQuestionAnswering`.
  * 
  * **Example:** Run question answering with `Xenova/distilbert-base-uncased-distilled-squad`.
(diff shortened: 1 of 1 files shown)


xenova:
Thanks for the PR! Considering that these are required inputs for the different pipelines, perhaps it would be best to either
1. set them to `null` and throw and error if not present
2. add a `...args` (or similar) to the base class's call method?

WDYT?

kungfooman:
@xenova Thank you for testing, I like option (2) most, since (1) would introduce `null` types, even though we don't "really" need them.

With (2) we can just add `...unused`:

```js
export class Pipeline extends Callable {
// ...
    async _call(texts, ...unused) {
        // ...
    }
// ...
}
```

*will update PR to test*

kungfooman:
I like your (2) solution, even less changes in PR now without introducing arbitrary default values :+1:

I don't really see the point of LSP anyway in this case.

xenova:
Great - thanks for this! It might also be a good idea to define return types for the rest of the pipelines, and somehow make it work when the user does:

```js
const pipe = pipeline('task', 'model');
const output = pipe(...); // Typing works, and depends on the task above.
```

but we can leave that for another PR.

Issue #263, opened 2023-08-28
Headers is not defined


utiq:
**Describe the bug**
*A clear and concise description of what the bug is.*

I tried to create a nodeJS express app, I get this error when I make a call to the server:
```
Unable to load from local path "/node_modules/@xenova/transformers/models/Xenova/distilbert-base-uncased-finetuned-sst-2-english/tokenizer.json": "ReferenceError: Headers is not defined"
Unable to load from local path "/node_modules/@xenova/transformers/models/Xenova/distilbert-base-uncased-finetuned-sst-2-english/tokenizer_config.json": "ReferenceError: Headers is not defined"
Unable to load from local path /node_modules/@xenova/transformers/models/Xenova/distilbert-base-uncased-finetuned-sst-2-english/config.json": "ReferenceError: Headers is not defined"
```

xenova:
Hi there. What version of node are you using? We recommend using at least 18.

utiq:
> Hi there. What version of node are you using? We recommend using at least 18.

16.15.1

xenova:
> 16.15.1

Yes, that is the reason then. You might be able to get around this by just importing from node-fetch and updating the global `fetch` and `Headers` objects.

For example:
```js
import fetch, {Headers} from 'node-fetch'

if (!globalThis.fetch) {
  globalThis.fetch = fetch
  globalThis.Headers = Headers
}
```
^^ from https://stackoverflow.com/a/71626669/13989043

xenova:
Oh yes, thanks for that! I think I just copied the versions of Vite from another tutorial. I will fix that.

Kiranism:
I'm using Next.js 13 with Node.js version 19, and I'm still encountering this error. Interestingly, when I perform a certain function, I receive a response. However, when I call another function that does the same thing, it returns an error. Everything was fine untill i changed the model. Unable to load from local path
im new to this transformer thing so please help

xenova:
@Kiranism does [this](https://github.com/xenova/transformers.js/issues/263#issuecomment-1695867568) not work for you? However, node 19 should have `Headers` defined globally, no?

PR #293, opened 2023-09-11
docs: update using minimal node 18+

rubiagatra:
See #263 

As @xenova we need start to using node 18+ instead of 16+


xenova:
Thanks!

PR #301, opened 2023-09-12
make loading, results and input text field visible

Bilalkhan4086:
There was an issue in nextjs examples ( in both  server and client) that we cannot able to see the text for loading, results and what we type in input field. Its due to there white text color, so to make it visible I make in black.

### DIFF
--- a/examples/next-client/src/app/page.js
+++ b/examples/next-client/src/app/page.js
@@ -53,7 +53,7 @@ export default function Home() {
       <h2 className="text-2xl mb-4 text-center">Next.js template (client-side)</h2>
       <input
         type="text"
-        className="w-full max-w-xs p-2 border border-gray-300 rounded mb-4"
+        className="w-full text-black max-w-xs p-2 border border-gray-300 rounded mb-4"
         placeholder="Enter text here"
         onInput={e => {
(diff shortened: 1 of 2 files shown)


Bilalkhan4086:
Please review the changes

xenova:
Hi there 👋 Could you send a screenshot of what you were seeing?

This is what it shows for me:
image

Perhaps this is an issue with a browser extension, or maybe a default colour scheme?

Bilalkhan4086:
Don't know why, for me its working like this (with default code)...
https://github.com/xenova/transformers.js/assets/63542485/ecb6863d-9332-4980-9670-c8db65002228

xenova:
This appears to be an issue with either your browser or system being dark mode by default. Your fix won't affect the light mode version, so if you'd like to reopen, I'd be happy to merge 👍

Bilalkhan4086:
Sure! Thanks,
Actually, I want to apologise, because I created next app use create-next-app and then copy and paste the components.
So default theme of next app is dark, so that is why, it's happening.

But if this fix doesn't effect the light mode version, then please merge it, because If someone else repeat the same path, he may encounter with same issue.
BTW Thanks for your response. 👍

PR #306, opened 2023-09-15
readme: add feature extraction docs and models

happysalada:
hey, I was just a bit surprised to see feature extraction missing in the list of tasks.
That said I noticed the sentence-similarity points to the feature extraction pipeline.
not sure what you want to do with this, happy to just close this, I just thought I would propose it.


xenova:
Hi there! 👋 The feature extraction task is listed in the [multimodal](https://github.com/xenova/transformers.js#multimodal) section! This is because feature extraction can also be used for image, audio, etc. tasks.

However, as you suggest, the main use-case is for computing text embeddings - which is correct! I do plan to extend the functionality of the feature extraction pipeline to make it much easier to work with multimodal inputs.

happysalada:
Perfect, i had missed it, thanks !

PR #325, opened 2023-09-21
Remove `CustomCache` requirement

xenova:
I revisted this today, and while testing, I didn't run into any issues with the Web Cache API this time. 🤷 

@do-me @VarunNSrivastava If possible, could you test this example project locally too? Just to make sure it doesn't break anything.


VarunNSrivastava:
It's working locally for me.

xenova:
I think it was crashing before because I was doing a local check before downloading from the Hub. Either that, or there was a Chrome bug fix in the past 2 months (which I doubt 😅).

@VarunNSrivastava Thanks for the info! I'll get these changes merged soon.

PR #389, opened 2023-11-13
Add depth estimation pipeline

### LINKED ISSUE #350: [Feature request] Depth Estimation pipeline
[This demo](https://huggingface.co/spaces/ilumine-AI/Insta-3D) has got me interested in playing around with depth stuff on the web, so it'd be cool if it were available as a pipeline in Transformers.js!

 * https://huggingface.co/tasks/depth-estimation
 * https://huggingface.co/models?pipeline_tag=depth-estimation

xenova:
This PR adds support for the `depth-estimation` pipeline w/ DPT and GLPN

Closes #350 

### Example usage:

```js
import { pipeline } from '@xenova/transformers';

let url = 'https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/cats.jpg';
let depth_estimator = await pipeline('depth-estimation', 'Xenova/dpt-hybrid-midas');
let out = await depth_estimator(url);
// {
//   predicted_depth: Tensor {
//     dims: [ 384, 384 ],
//     type: 'float32',
//     data: Float32Array(147456) [ 542.859130859375, 545.2833862304688, 546.1649169921875, ... ],
//     size: 147456
//   },
//   depth: RawImage {
//     data: Uint8Array(307200) [ 86, 86, 86, ... ],
//     width: 640,
//     height: 480,
//     channels: 1
//   }
// }
```

### Input
image

### Output:
1. Python (baseline):
image

Code used: https://huggingface.co/Intel/dpt-hybrid-midas#how-to-use

2. JavaScript (quantized):
depth_quantized

3. JavaScript (unquantized):
depth_unquantized


josephrocca:
Xenova you need to STOP. enough is Enough. "i'm xenova and along with the ort web and optimium teams i'm going to get literally all the <1gb sota ML models quantized and running in the browser, and make it really easy for web devs to use"-- NO. This is not Natural. Web devs are meant to struggle with op support on a tflite model for 2 weeks before realising that they really can't get around the TF Select op requirements, and then steel themselves for creating a custom tflite build, but then find out that the build tooling required for adding TF Select ops is Google-internal, and then try to port the model to tfjs instead, but end up with weird errors, which they eventually fix by breaking the model apart into multiple models and implementing some operations in JS to glue the model back together, with one component actually being executed via the tflite runtime because it has support for a particular op that tfjs didn't have, but then the results for one part of the tfjs model are different from the Python model, so they post an issue with a Colab that isolates and minimally replicates it, but that isn't replied to until 6 months later and that reply is actually just the stale bot.

josephrocca:
I should probably post this as a separate issue, but is there a way to get a trimmed down version of transformers.js that only has the depth estimation pipeline? I.e. a "treeshaking" type feature?

xenova:
@josephrocca ❤️ and we're just getting started 😏 

> is there a way to get a trimmed down version of transformers.js that only has the depth estimation pipeline?

This has actually been requested quite a lot by others, like for use-cases where only tokenizers are used (so, we don't need onnxruntime-web to be bundled). So, please feel free to open up a feature request. Perhaps some others might be able to contribute to the discussion.

For the most part, I think this responsibility should be transferred to build tools like webpack or rollup, as they have much better and advanced support for this. In the worst case, you can just fork the repo and remove everything you don't need :)

PR #404, opened 2023-11-19
Move tensor clone to fix Worker ownership NaN issue

### LINKED ISSUE #390: [Bug] `NaN`s produced when `env.backends.onnx.wasm.proxy = true`?
I think I'm doing something silly here, but setting `env.backends.onnx.wasm.proxy = true` makes this produce `NaN`s:
```js
let { pipeline, env } = await import('https://cdn.jsdelivr.net/npm/@xenova/transformers@2.8.0');
env.allowLocalModels = false;
// env.backends.onnx.wasm.proxy = true; // uncomment this line to produce NaNs
let extractor = await pipeline('feature-extraction', 'Xenova/bge-base-en-v1.5');
let result = await extractor("test 123", { pooling: 'mean', normalize: true });
console.log(result.data);
```
I thought maybe it was a `2.8.0` regression, but I tried all versions back to `2.5.4` and it's didn't work in any of them, so that makes me lean toward assuming that I'm doing something wrong. If that's the case, then an informative error message would be helpful :)

kungfooman:
Fixes #390

(this is probably a temporary fix, depending on what ONNXRuntime will be doing, but at least we prevent invalid tensor NaN bugs in the meantime)

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -166,7 +166,7 @@ function validateInputs(session, inputs) {
             continue;
         }
         // NOTE: When `env.wasm.proxy is true` the tensor is moved across the Worker
-        // boundary, the ownership is transferred to the worker, invalidating the tensor.
+        // boundary, transferring ownership to the worker and invalidating the tensor.
         // So, in this case, we simply sacrifice a clone for it.
         checkedInputs[inputName] = env.wasm.proxy ? tensor.clone() : tensor;
     }
--- a/src/models.js
+++ b/src/models.js
@@ -82,7 +82,9 @@ import {
 
 import { executionProviders, ONNX } from './backends/onnx.js';
 import { medianFilter } from './transformers.js';
-const { InferenceSession, Tensor: ONNXTensor } = ONNX;
+const { InferenceSession, Tensor: ONNXTensor, env } = ONNX;
+
+/** @typedef {import('onnxruntime-web').InferenceSession} InferenceSession */
 
 //////////////////////////////////////////////////
 // Model types: used internally
@@ -142,20 +144,27 @@ async function constructSession(pretrained_model_name_or_path, fileName, options
 /**
  * Validate model inputs
  * @param {InferenceSession} session The InferenceSession object that will be run.
- * @param {Object} inputs The inputs to check.
+ * @param {Record<string, Tensor>} inputs The inputs to check.
  * @returns {Promise<Object>} A Promise that resolves to the checked inputs.
  * @throws {Error} If any inputs are missing.
  * @private
  */
 async function validateInputs(session, inputs) {
-    // NOTE: Only create a shallow copy
+    // NOTE: Create either a shallow or deep copy based on `onnx.wasm.proxy`
     const checkedInputs = {};
     const missingInputs = [];
-    for (let inputName of session.inputNames) {
-        if (inputs[inputName] === undefined) {
+    for (const inputName of session.inputNames) {
+        const tensor = inputs[inputName];
+        if (!tensor) {
             missingInputs.push(inputName);
         } else {
-            checkedInputs[inputName] = inputs[inputName];
+            if (env.wasm.proxy) {
+                // Moving the tensor across Worker boundary moves ownership to the worker,
+                // which invalidates the tensor. So we simply sacrifize the clone for it.
+                checkedInputs[inputName] = tensor.clone();
+            } else {
+                checkedInputs[inputName] = tensor;
+            }
         }
     }
     if (missingInputs.length > 0) {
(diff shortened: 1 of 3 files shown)


kungfooman:
Hey @xenova, there is nothing in `async function validateInputs(session, inputs)` that justifies it being `async` while returning a `Promise`, shall we just simplify it?

xenova:
@kungfooman Oh yes, good spot 👍 Sure, we can simplify it :)

src/models.js
```diff
@@ -142,20 +144,27 @@ async function constructSession(pretrained_model_name_or_path, fileName, options
...
- * @param {Object} inputs The inputs to check.
+ * @param {Record<string, Tensor>} inputs The inputs to check.
  * @returns {Promise<Object>} A Promise that resolves to the checked inputs.
  * @throws {Error} If any inputs are missing.
  * @private
  */
 async function validateInputs(session, inputs) {
-    // NOTE: Only create a shallow copy
+    // NOTE: Create either a shallow or deep copy based on `onnx.wasm.proxy`
     const checkedInputs = {};
```
xenova, line 154:
```suggestion
    const checkedInputs = Object.create(null);
```
Another thing I've noticed, there could be a rare case where one of the model's input names corresponds to a built-in object name (e.g., `toString`), which would cause the later `(!tensor)` check to fail. Probably good to fix it now too.

xenova (changes requested):

src/models.js
kungfooman: Hm, that's interesting, normally a `Map` is used then, but `ONNXRuntime` only supports an object...

Given this code and putting "extern" in front of `validateInputs` for testing:

```js
const session = {
    inputNames: ['a', 'b', 'toString', 'valueOf']
};
const inputs = {
    a: new Tensor("float32", [1, 2, 3, 4], [2, 2]),
    b: new Tensor("float32", [1, 2, 3, 4], [1, 4]),
    //toString: new Tensor("float32", [1, 2, 3, 4], [1, 4]),
    valueOf: new Tensor("float32", [1, 2, 3, 4], [4, 1]),
};
const checkedInputs = validateInputs(session, inputs);
console.log(Object.keys(checkedInputs));
```

I tested two scenarios:

image

This happens with or without the `Object.create(null)` change, because `tensor` comes from the `inputs` arguments.

This should probably throw, but it doesn't:

```js
env.backends.onnx.wasm.proxy = false;
const session = {
    inputNames: ['toString', 'valueOf']
};
const inputs = {};
const checkedInputs = validateInputs(session, inputs);
console.log(Object.keys(checkedInputs));
```

So I just tested a `tensor instanceof Tensor` check and that detects the missing inputs:

image

Thank you for your quick input and review!

src/models.js
```diff
@@ -142,21 +144,31 @@ async function constructSession(pretrained_model_name_or_path, fileName, options
 /**
  * Validate model inputs
  * @param {InferenceSession} session The InferenceSession object that will be run.
- * @param {Object} inputs The inputs to check.
- * @returns {Promise<Object>} A Promise that resolves to the checked inputs.
+ * @param {Record<string, Tensor>} inputs The inputs to check.
+ * @returns {Record<string, Tensor>} A Promise that resolves to the checked inputs.
```
xenova, line 148:
```suggestion
 * @returns {Record<string, Tensor>} The checked inputs.
```

src/models.js
```diff
@@ -142,21 +144,31 @@ async function constructSession(pretrained_model_name_or_path, fileName, options
...
+        if (!(tensor instanceof Tensor)) {
             missingInputs.push(inputName);
-        } else {
-            checkedInputs[inputName] = inputs[inputName];
+            continue;
         }
+        // NOTE: When `env.wasm.proxy is true` the tensor is moved across the Worker
+        // boundary, the ownership is transferred to the worker, invalidating the tensor.
+        // So, in this case, we simply sacrifice a clone for it.
+        checkedInputs[inputName] = env.wasm.proxy ? tensor.clone() : tensor;
```
xenova, line 171:
```suggestion
        // NOTE: When `env.wasm.proxy is true` the tensor is moved across the Worker
        // boundary, transferring ownership to the worker and invalidating the tensor.
        // So, in this case, we simply sacrifice a clone for it.
        checkedInputs[inputName] = env.wasm.proxy ? tensor.clone() : tensor;
```

xenova (approved):
LGTM. Minor nits.

src/models.js
```diff
@@ -82,7 +82,9 @@ import {
 
 import { executionProviders, ONNX } from './backends/onnx.js';
 import { medianFilter } from './transformers.js';
-const { InferenceSession, Tensor: ONNXTensor } = ONNX;
+const { InferenceSession, Tensor: ONNXTensor, env } = ONNX;
+
+/** @typedef {import('onnxruntime-web').InferenceSession} InferenceSession */
```
xenova, line 87:
```suggestion
```

PR #408, opened 2023-11-21
Add support for chat templates

xenova:
This PR adds support for [chat templates](https://huggingface.co/docs/transformers/main/en/chat_templating), making it possible to generate well-formatted prompts in a generic fashion. This will also be useful to users of other libraries, like TGI, as this means they can generating prompts on the client-side ([link](https://github.com/huggingface/text-generation-inference/issues/1088) to some discussion).

Since the python library's implementation uses [Jinja](https://github.com/pallets/jinja), it required a minimalistic JavaScript reimplementation of the templating engine. Special thanks to [Tyler Laceby](https://github.com/tlaceby) for his amazing ["Guide to Interpreters"](https://github.com/tlaceby/guide-to-interpreters-series) tutorial series, which provided the basis for this implementation.

**Example:** Applying a chat template to a chat history

```javascript
import { AutoTokenizer } from '@xenova/transformers';

const tokenizer = await AutoTokenizer.from_pretrained("mistralai/Mistral-7B-Instruct-v0.1");

const chat = [
  { "role": "user", "content": "Hello, how are you?" },
  { "role": "assistant", "content": "I'm doing great. How can I help you today?" },
  { "role": "user", "content": "I'd like to show off how chat templating works!" },
]

const text = await tokenizer.apply_chat_template(chat, { tokenize: false });
// "<s>[INST] Hello, how are you? [/INST]I'm doing great. How can I help you today?</s> [INST] I'd like to show off how chat templating works! [/INST]"

const input_ids = await tokenizer.apply_chat_template(chat, { tokenize: true, return_tensor: false });
// [1, 733, 16289, 28793, 22557, 28725, 910, 460, 368, 28804, 733, 28748, 16289, 28793, 28737, 28742, 28719, 2548, 1598, 28723, 1602, 541, 315, 1316, 368, 3154, 28804, 2, 28705, 733, 16289, 28793, 315, 28742, 28715, 737, 298, 1347, 805, 910, 10706, 5752, 1077, 3791, 28808, 733, 28748, 16289, 28793]
```

cc @OlivierDehaene (original requester of the feature)

## TODOs

- [x] Add support for built-in string functions (e.g., `.strip()`)
- [x] Add support for logical operators
- [x] Add all loop variables (like `loop.last`)
- [x] Add full support for nested for-loops (currently, the loop index is reused)
- [x] Auto-generate unit tests
- [x] Add support for array slicing (e.g., `messages[1:]`)
- [x] Do `trim_blocks=True, lstrip_blocks=True` (needed by [HuggingFaceH4/zephyr-7b-beta](https://huggingface.co/HuggingFaceH4/zephyr-7b-beta))
- [x] Add test cases for invalid templates


xenova:
The larger this PR gets, the more I wonder whether it's better to extract this logic into a separate library, and then make it an optional dependency for transformers.js. Chat templates are most likely not going to be used by all users, and I'm considering whether the new features justify the increase to the bundle size (783KB -> 801KB).

Would love to hear others' thoughts 👇

gary149:
> The larger this PR gets, the more I wonder whether it's better to extract this logic into a separate library, and then make it an optional dependency for transformers.js

Side note: it would be great to have chat templates into https://github.com/huggingface/huggingface.js/blob/main/packages/inference/README.md 👀 btw

kungfooman:
> Would love to hear others' thoughts 👇

Isn't it decided already? Personally I would rather stick to JSDoc, since you can't load it easily in the browser for quick PR testing any longer and you have to maintain another package now.

But yes, bundle size matters for non-tree-shaked builds/examples, so rather have an extra HTTP script request for examples that deal with chat templates.

PR #409, opened 2023-11-22
Webpack: add/build RTI target

kungfooman:
I'm excited to introduce [RuntimeTypeInspector.js](https://runtimetypeinspector.org/), which I believe will benefit Transformers.js as a whole and our debugging workflows in specific. RTI is designed to type-check the arguments of every function and method. It detects type errors while the engine is running in a way that goes beyond the capabilities of TypeScript, as demonstrated in [this video](https://www.youtube.com/watch?v=o5ipQe2rVKQ) (there is another one on the website debugging the Microsoft Holometric Video PC project).

During the development of RTI it already helped me to understand and fix things, for example various JSDoc types and this NaN bug: https://github.com/xenova/transformers.js/issues/390

I believe that RTI will help improve productivity and efficiency, by catching errors early on and reporting them in detail.

Runtime Type Inspector is implemented using Babel AST traversal with custom JSDoc parsing (powered by TypeScript itself for highest compatibility).

TLDR: It adds type assertions for runtime type checking, for example:

image

Being able to set breakpoints exactly where the errors are happening for the first time helps to track down problems that are otherwise difficult to pinpoint (and saves a lot of time :sweat_smile:).

Try for yourself: https://pc.runtimetypeinspector.org/#/animation/blend-trees-2d-cartesian

Just make sure to pick the right JS context - the exampleIframe:

image

I'm working on this for quite some time by now, a mix of fun and frustration - rewriting code on the AST level first seemed tricky, but given enough time, it was a rather straightforward way to automate type validations. I'm excited to read feedback and suggestions. :pray:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "sharp": "^0.32.0"
       },
       "devDependencies": {
+        "@runtime-type-inspector/plugin-webpack5": "^1.1.2",
         "@types/jest": "^29.5.1",
         "catharsis": "github:xenova/catharsis",
         "copy-webpack-plugin": "^11.0.0",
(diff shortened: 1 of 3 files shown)


kungfooman:
I wanted the examples to be build-free while also being able to use RTI type checking on each of them. Because that was quite a challenge in the beginning, I decided to make a custom repo for it: https://github.com/kungfooman/transformers-examples.js

The first main challenge was loading a worker containing ESM code depending on import map resolution... I found some brittle solutions, but ended up making my own NPM package for it: https://github.com/kungfooman/worker-with-import-map/

Still working on some examples and making everything easier to use, but most of them already work build-free and with RTI :see_no_evil:

PR #451, opened 2023-12-11
Fix tensor inheritance

### LINKED ISSUE #437: error TS2510: Base constructors must all have the same return type.
### System Info

Node: v20.9.0
tsc: The TypeScript Compiler - Version 5.2.2

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension
- [X] Server-side (e.g., Node.js, Deno, Bun)

### Description

I've been building my app with `tsc` for quite a while. Today, I tried to upgrade `@xenova/transformers` from `2.7.0` to `2.10.1` but if tailed due to the error below. `2.8.0` works fine, but it doesn't work `2.9.0` and afterward.

### Reproduction

```
npm install @xenova/transformers@2.10.1
echo "import { pipeline, Pipeline } from '@xenova/transformers'" > index.ts
tsc index.ts
```

```
node_modules/@xenova/transformers/types/utils/tensor.d.ts:75:29 - error TS2510: Base constructors must all have the same return type.

75 export class Tensor extends Tensor_base {
                               ~~~~~~~~~~~
Found 1 error in node_modules/@xenova/transformers/types/utils/tensor.d.ts:75
```

xenova:
fixes #437

### DIFF
--- a/src/utils/maths.js
+++ b/src/utils/maths.js
@@ -130,9 +130,9 @@ export function transpose_data(array, dims, axes) {
 
 /**
  * Compute the softmax of an array of numbers.
- *
- * @param {number[]} arr The array of numbers to compute the softmax of.
- * @returns {number[]} The softmax array.
+ * @template {TypedArray|number[]} T
+ * @param {T} arr The array of numbers to compute the softmax of.
+ * @returns {T} The softmax array.
  */
 export function softmax(arr) {
     // Compute the maximum value in the array
@@ -142,18 +142,20 @@ export function softmax(arr) {
     const exps = arr.map(x => Math.exp(x - maxVal));
 
     // Compute the sum of the exponentials
+    // @ts-ignore
     const sumExps = exps.reduce((acc, val) => acc + val, 0);
 
     // Compute the softmax values
     const softmaxArr = exps.map(x => x / sumExps);
 
-    return softmaxArr;
+    return /** @type {T} */(softmaxArr);
 }
 
 /**
  * Calculates the logarithm of the softmax function for the input array.
- * @param {number[]} arr The input array to calculate the log_softmax function for.
- * @returns {any} The resulting log_softmax array.
+ * @template {TypedArray|number[]} T
+ * @param {T} arr The input array to calculate the log_softmax function for.
+ * @returns {T} The resulting log_softmax array.
  */
 export function log_softmax(arr) {
     // Compute the softmax values
@@ -162,7 +164,7 @@ export function log_softmax(arr) {
     // Apply log formula to each element
     const logSoftmaxArr = softmaxArr.map(x => Math.log(x));
 
-    return logSoftmaxArr;
+    return /** @type {T} */(logSoftmaxArr);
 }
 
 /**
@@ -178,8 +180,7 @@ export function dot(arr1, arr2) {
 
 /**
  * Get the top k items from an iterable, sorted by descending order
- *
- * @param {Array} items The items to be sorted
+ * @param {any[]|TypedArray} items The items to be sorted
  * @param {number} [top_k=0] The number of top items to return (default: 0 = return all)
  * @returns {Array} The top k items, sorted by descending order
  */
--- a/src/utils/maths.js
+++ b/src/utils/maths.js
@@ -252,8 +252,8 @@ export function min(arr) {
 
 /**
  * Returns the value and index of the maximum element in an array.
- * @param {number[]|TypedArray} arr array of numbers.
- * @returns {number[]} the value and index of the maximum element, of the form: [valueOfMax, indexOfMax]
+ * @param {number[]|AnyTypedArray} arr array of numbers.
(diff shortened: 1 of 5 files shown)


xenova:
cc @kungfooman - I'd appreciate a review before I merge.

src/utils/tensor.js
```diff
@@ -15,8 +15,10 @@ import {
 } from './maths.js';
 
 
-// @ts-ignore
-const DataTypeMap = new Map([
+/**
+ * @typedef {'bool'|'float32'|'float64'|'string'|'int8'|'uint8'|'int16'|'uint16'|'int32'|'uint32'|'int64'} DataType 
+ */
+const DataTypeMap = new Map( /** @type {[DataType, any][]} */([
```
kungfooman, line 21: This is creating so much ASCII soup, why not just:

```js
const DataTypeMap = {
    bool: Uint8Array,
    float32: Float32Array,
    float64: Float64Array,
    string: Array, // string[]
    int8: Int8Array,
    uint8: Uint8Array,
    int16: Int16Array,
    uint16: Uint16Array,
    int32: Int32Array,
    uint32: Uint32Array,
    int64: BigInt64Array,
};
```

You get better typing for free without any casting hacks:

image

Probably just went overboard with "Map is new holy grail" suggestions? We don't need any of that, the order doesn't matter, it won't contain keys like `toString`, we only need strings as key etc.

src/utils/tensor.js
```diff
@@ -28,27 +30,40 @@ const DataTypeMap = new Map([
...
     ['uint32', Uint32Array],
     ['int64', BigInt64Array],
-])
+]))
 
 /**
  * @typedef {import('./maths.js').AnyTypedArray | any[]} DataArray
  */
 
 const ONNXTensor = ONNX.Tensor;
```
kungfooman, line 39: Only used as type for the constructor, but stripped from type definitions, so it will be unresolved/any, we can just remove the line

src/utils/tensor.js
```diff
@@ -28,27 +30,40 @@ const DataTypeMap = new Map([
...
+    /** @type {DataArray} The data stored in the tensor. */
+    data;
+
+    /** @type {number[]} Dimensions of the tensor. */
+    dims;
+
     /**
      * Create a new Tensor or copy an existing Tensor.
-     * @param {[string, DataArray, number[]]|[ONNXTensor]} args
+     * @param {[DataType, DataArray, number[]]|[ONNXTensor]} args
```
kungfooman, line 53:
```suggestion
     * @param {[DataType, DataArray, number[]]|[import('onnxruntime-common').Tensor]} args
```

src/utils/tensor.js
```diff
@@ -28,27 +30,40 @@ const DataTypeMap = new Map([
...
 
 const ONNXTensor = ONNX.Tensor;
 
-export class Tensor extends ONNXTensor {
+export class Tensor {
+    /** @type {DataType} Type of the tensor. */
+    type;
+
+    /** @type {DataArray} The data stored in the tensor. */
+    data;
```
kungfooman, line 46: ONNX has this order:

```js
Object.keys(new onnx.ONNX.Tensor('string', ['a', 'b', 'c', 'd']))
```

image

In short, `dims` goes to the top while adding `size` (a number) at the end.

(this ensures internal object alignment, so `JSON.stringify` stays compatible)

src/utils/maths.js
```diff
@@ -142,18 +142,20 @@ export function softmax(arr) {
...
+    return /** @type {T} */(softmaxArr);
 }
 
 /**
  * Calculates the logarithm of the softmax function for the input array.
- * @param {number[]} arr The input array to calculate the log_softmax function for.
- * @returns {any} The resulting log_softmax array.
+ * @template {TypedArray|number[]} T
+ * @param {T} arr The input array to calculate the log_softmax function for.
+ * @returns {T} The resulting log_softmax array.
```
kungfooman, line 158: Very nice change :+1: I planned to add that to my JSDoc fix PR aswell, sinces `typed arrays` != `number[]`

src/utils/tensor.js
kungfooman: I also suggest to keep "type variables" lower-case, it's the standard:

https://github.com/playcanvas/engine/blob/f9bac39b27f9df00d1ca6285a3c1c36b5af77fd4/src/framework/parsers/glb-parser.js#L151-L166

xenova:
Thanks @kungfooman! Applied the suggestions now. I've left the `DataTypeMap` variable uppercase since that's how it's defined in onnxruntime. I also just added a simple `hasOwnProperty` just to avoid weird cases like the user passing `"toString"` as the data type.

kungfooman:
Very nice changes, thank you @xenova! I looked over it again and run `npm run typegen` produces this `any` type right now:

image

Otherwise I'm very happy about your work here :1st_place_medal:

kungfooman:
> @kungfooman Great :) I've fixed that type now too 👍 lmk if I can merge and put out a new release :)

Very nice, PR looks very good! (famous last words) - I can't see any problems or errors in the type declaration files any longer

PR #454, opened 2023-12-12
Add openvinojs backend

vishniakov-nikolai:
125201


vishniakov-nikolai:
Draft status of PR will removed after openvinojs publication

PR #461, opened 2023-12-19
Fix Next.js Dockerfile HOSTNAME

Lian1230:
HOSTNAME localhost keep the space App waiting for the build.

### DIFF
--- a/examples/next-server/Dockerfile
+++ b/examples/next-server/Dockerfile
@@ -59,7 +59,7 @@ USER nextjs
 EXPOSE 3000
 
 ENV PORT 3000
-ENV HOSTNAME localhost
+ENV HOSTNAME 0.0.0.0
 
 # Allow the running process to write model files to the cache folder.
 # NOTE: In practice, you would probably want to pre-download the model files to avoid having to download them on-the-fly.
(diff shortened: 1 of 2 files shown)


xenova:
Thanks for this! Looks like they recently updated their template: https://github.com/vercel/next.js/pull/54342.
Could you also update the other instance of this [here](https://github.com/xenova/transformers.js/blob/0bf6e6712faddad0f9a913d30bb168906d71ad6b/examples/semantic-image-search/Dockerfile#L62)?

xenova:
> keep the space App waiting for the build.

Just for clarity, are you saying you're running into issues when running from Hugging Face spaces? If so, can you link to that space where it's broken? You may be missing `app_port: 3000` in the README if this is the case. If this is not related to Hugging Face spaces, then I'm happy to merge.

Lian1230:
@xenova yes, it's in Hugging Face spaces.
`app_port` is set to `3000`, but the traffic to the container is blocked.
It's also reproducible in local, if you build and run this container, 3000 is not reachable on host.

xenova:
That's interesting - I've been able to deploy next.js apps in the past, like this one: https://huggingface.co/spaces/Xenova/next-server-example-app. Do you have a link to it so I can check it out?

Lian1230:
@xenova I just created one for you: https://huggingface.co/spaces/phuhui/reproduce_error
maybe it's next14:
image

xenova:
I was able to reproduce this - thanks for reporting and fixing! Merging now :)

PR #485, opened 2023-12-29
Improve typing of `pipeline` helper function

### LINKED ISSUE #484: TypeScript Pipline Types for different models?
### Question

Is there a suggested way to get types for the different models? Right now after I create a pipline, like one of the following:

```
const segmenter = await pipeline('image-segmentation', 'Xenova/face-parsing');
// or 
const extractor = await pipeline(`feature-extraction`, `Xenova/UAE-Large-V1`, {
  quantized: true, // Set this to false to use the full (unquantized) model
});
```

All the methods and returned values are `(...args: any[]) => any` - finding it hard to work with the methods and returned values.

I realize each model returns different outputs, and I'm fairly new to the whole convertion process, but are these types kept somewhere in the Python or the json files with the model that could be used as typescript types? 

Ideally `pipeline` would infer the types, but I'm also ok with importing (or generating the types myself) and using it as a generic:

```
const whateve = pipeline<ReturnType>(`task`, `model`)
```

xenova:
A first attempt at improving the types inferred when creating a new pipeline.

image

Still to do: return actual call parameters instead of `(...args: any[]) => any`.

Closes #484.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -813,7 +813,7 @@ export class PreTrainedModel extends Callable {
     }
 
     /**
-     * @param {GenerationConfig} generation_config 
+     * @param {import('./utils/generation.js').GenerationConfigType} generation_config 
      * @param {number} input_ids_seq_length The starting sequence length for the input ids.
      * @returns {LogitsProcessorList}
      * @private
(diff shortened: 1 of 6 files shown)


xenova:
After a bit of tinkering, I think I've got a good solution, which I've tested with the image segmentation pipeline, since this was requested by @wesbos in the original issue.

typing-demo

This also means when you hover over the class name, you'll get example code to help you out.
typing-demo2

kungfooman:
Wow, very nice work! :100: :+1:

vickkhera:
This changed the interface of `pipeline()` which broke my typescript. It should have caused a bump in minor version number at least, not just the patch number.  See #504

xenova:
Hi there. This was fixed in a subsequent version (2.13.3), could you please upgrade and try again?

vickkhera:
> Hi there. This was fixed in a subsequent version (2.13.3), could you please upgrade and try again?

I updated to 2.13.4 when I noticed the breakage and just narrowed it down to 2.13.1 release.

Even with 2.13.4 I had to rewrite my TS code to work with these changes. They are good changes but they are breaking changes so should have caused a version bump rather than a patch level.

Specifically I had to change the type of the function returned by `pipeline()` and then the type returned by that generated function also changed for some but not all tasks.

PR #488, opened 2023-12-29
wasm does not work on node right now with multiple threads

sroussey:
Reference: https://github.com/microsoft/onnxruntime/issues/17274#issuecomment-1692587686

Also (https://github.com/xenova/transformers.js/pull/488)


sroussey:
I should note that you ran into this as well before, as you have this is tests:

https://github.com/xenova/transformers.js/blob/61459e38d8978b88e4a887a6f2ce3d6f15528fcc/tests/init.js#L16

tianhuil:
Hi -- I ran into the same issue with using bun.  Could we merge this @xenova ?

sroussey:
this is now fixed

PR #508, opened 2024-01-09
fix: token-classification start and end

llmer:
## Motivation
Fix token-classification pipeline https://github.com/xenova/transformers.js/issues/359

## Implementation
Keep track of current position to calculate start and end


llmer:
mm not quite ready. seeing some differences between python and js.

PR #519, opened 2024-01-12
Add zero-shot classification demo

xenova:
Source code for https://huggingface.co/spaces/Xenova/zero-shot-classification-demo

zero-shot-demo-final

### DIFF
--- a/examples/zero-shot-classification/.eslintrc.cjs
+++ b/examples/zero-shot-classification/.eslintrc.cjs
@@ -1,0 +1,20 @@
+module.exports = {
+  root: true,
+  env: { browser: true, es2020: true },
+  extends: [
+    'eslint:recommended',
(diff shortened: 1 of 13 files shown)


sroussey:
How can you get the model to give you the categories to begin with?

xenova:
> How can you get the model to give you the categories to begin with?

I'd recommend using [sentence-transformers](https://www.sbert.net/examples/applications/clustering/README.html#topic-modeling) for this. There are some tutorials online (like [this one](https://towardsdatascience.com/topic-modeling-with-bert-779f7db187e6)), which can also be useful here.  Although, if you're interested in running with transformers.js, you should be able to reimplement some of the helper functions in JS, perhaps those needed as preprocessing or postprocessing steps on top of the model itself (which you can run with transformers.js).

PR #524, opened 2024-01-17
Prevent electron use node backend

hans00:
As title.

Prevent using the node backend when Electron disabled isolation.


hans00:
https://onnxruntime.ai/docs/tutorials/web/build-web-app.html#options-for-deployment-target

I found the official recommendation. I'll change my implement

PR #582, opened 2024-02-10
BertProcessing add_prefix_space, trim_offsets

shavit:
Use config properties to trim space from tokens

https://github.com/xenova/transformers.js/blob/main/src/tokenizers.js#L1626


xenova:
Thanks for working on this! Do you maybe know of any model(s) and input for which these changes correct the current behaviour? The current set of supported models and unit tests do not have any issues with this.

shavit:
No, I just saw that this is missing from RobertaProcessing.

PR #596, opened 2024-02-21
fixes for ort-1.17

guschmue:
the ort tensor object in 1.17 has changed to add support for keeping webgpu tensors on gpu (ie to avoid copies of kv_cache or hidden state). As result the Object.assign() no longer works.
Either we extend the ort tensor or we decouple by keeping it around in the transformers.js tensor ... we think later is the cleanest. 
Take a look and let us know. Using extend is another option that would work.


guschmue:
I think this one is included in the v3 branch - closing.

Issue #551, opened 2024-01-30
Whisper model word-level timestamps broken


BjoernRave:
### System Info

"@xenova/transformers": "^2.14.0",

macbook with M2 chip and MacOs Sonoma

Node.js: 20.11.0

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension
- [X] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

I am running whisper like this:

```
export const speechToText = async (audio: Buffer) => {
  const float32Array = await convertAudioToFloat32Array(audio)
  env.allowLocalModels = false

  const transcriber = await pipeline(
    "automatic-speech-recognition",
    "Xenova/whisper-large-v3",
  )
  const output = await transcriber(float32Array, {
    return_timestamps: "word",
  })

  return output
}
```

However the returned word-level timestamps are all equal to the total duration of the audio file.

During the run my console also gets flooded with this kind of logs:

```
2024-01-30 15:14:37.230 node[87226:3900219] 2024-01-30 15:14:37.230677 [W:onnxruntime:, graph.cc:3490 CleanUnusedInitializersAndNodeArgs] Removing initializer '/model/decoder/layers.20/self_attn_layer_norm/Constant_1_output_0'. It is not used by any node and should be removed from the model.
2024-01-30 15:14:37.230 node[87226:3900219] 2024-01-30 15:14:37.230684 [W:onnxruntime:, graph.cc:3490 CleanUnusedInitializersAndNodeArgs] Removing initializer '/model/decoder/layers.12/self_attn_layer_norm/Constant_1_output_0'. It is not used by any node and should be removed from the model.
2024-01-30 15:14:37.230 node[87226:3900219] 2024-01-30 15:14:37.230693 [W:onnxruntime:, graph.cc:3490 CleanUnusedInitializersAndNodeArgs] Removing initializer '/model/decoder/layers.6/final_layer_norm/Constant_1_output_0'. It is not used by any node and should be removed from the model.
2024-01-30 15:14:37.230 node[87226:3900219] 2024-01-30 15:14:37.230701 [W:onnxruntime:, graph.cc:3490 CleanUnusedInitializersAndNodeArgs] Removing initializer '/model/decoder/layers.11/encoder_attn_layer_norm/Constant_1_output_0'. It is not used by any node and should be removed from the model.
[...]

PR #605, opened 2024-02-25
return stride information in callback from speech recognition

shopped:
In the [whisper web repo](https://github.com/xenova/whisper-web), I noticed that time stamp generation was off while being streamed in - but correct after the transcription is complete. If you look at the demo video, and pause during generation, you will see a 1:02 timestamp for a 1 minute audio clip - which is overwritten upon completion to 0:57.

This is because timestamps are generated correctly when they are passed in as completed chunks, but are off when partial updates are streamed back from the model generation. The callback from completed chunks in ```pipelines.js``` contains stride information, but the callback during partial generation in ```models.js``` does not, and stride information is needed to get correct timestamps from the ```decode_asr``` function in ```tokenizers.js```.

This is not a major issue right now, since the timestamps are only off by 5s, and the error is transient and corrects itself. However I want to make some improvements on whisper-web, like adding [word level timestamps](https://github.com/xenova/whisper-web/issues/17), making decoding more efficient, and adding streaming audio transcriptions from the microphone... and the timestamp discrepancies matter more in those code changes. I can do a workaround to solve this error in a fork of ```whisper-web```, but making the change here is cleaner. If this code change is merged in, the timestamp issue can be mitigated by adding ```last.stride = item.stride;``` after line 134 in ```worker.js``` in ```whisper-web``` - I can open a PR for that

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -1135,6 +1135,7 @@ export class PreTrainedModel extends Callable {
 
             // Run callback
             if (generation_config.callback_function) {
+                beams.stride = generation_config.stride;
                 generation_config.callback_function(beams);
             }
         }
--- a/src/pipelines.js
+++ b/src/pipelines.js
@@ -1427,6 +1427,7 @@ export class ZeroShotAudioClassificationPipeline extends (/** @type {new (option
  * @property {number[][]} [kwargs.forced_decoder_ids] A list of pairs of integers which indicates a mapping from generation indices to token indices
  * that will be forced before sampling. For example, [[1, 123]] means the second generated token will always be a token of index 123.
  * @property {number} [num_frames] The number of frames in the input audio.
+ * @property {number[]} [stride] Audio chunk information needed to generate proper timestamps.
  * @typedef {import('./utils/generation.js').GenerationConfigType & AutomaticSpeechRecognitionSpecificParams} AutomaticSpeechRecognitionConfig
  * 
  * @callback AutomaticSpeechRecognitionPipelineCallback Transcribe the audio sequence(s) given as inputs to text.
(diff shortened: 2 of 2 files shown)


wobbble:
Hey @shopped 
I have tried to implement your changes with my fork of `transformers.js` and then also changes in `whisper-web`, but it didn't help.
From digging in it I have found out that single change of value of `stride_length_s` from 5 to 3 in `whisper-web` helped me.
Here is detailed too and example audio.
https://github.com/xenova/transformers.js/issues/551#issuecomment-2033779889

Do you think that change is can be sustainable? or lowering stride will occasionally help only with some files?
Thanks!

shopped:
> Hey @shopped I have tried to implement your changes with my fork of `transformers.js` and then also changes in `whisper-web`, but it didn't help. From digging in it I have found out that single change of value of `stride_length_s` from 5 to 3 in `whisper-web` helped me. Here is detailed too and example audio. [#551 (comment)](https://github.com/xenova/transformers.js/issues/551#issuecomment-2033779889)
> 
> Do you think that change is can be sustainable? or lowering stride will occasionally help only with some files? Thanks!

Hey thanks for taking a look and trying to reproduce. Yes, I hardcoded a [stride length change too as a hack](https://github.com/xenova/whisper-web/pull/28) to get correct values initially, but decided to dig deeper and that's how I arrived at this code change. It is odd though that you are not able to reproduce - are you sure that your code references the local transformers.js change without hitting a vite cache or CDN cache?

Also, I have been getting correct word level timestamps during testing, but this PR is addressing incorrect chunk level timestamps. I have been busy with other projects but if I get some time and am back working on this, I will take a look at your specific issue. Cheers!

getflourish:
I also realized that the timestamps of chunks in streamed responses don’t add up. 

I end up with timestamps > length of the audio. The timestamps are correct in the final result. 

In general I find working with the `WhisperTextStreamer` not as pleasant because with each new chunk, the reported timestamps start at 0. It would be convenient if I don’t have to keep track of chunks myself :)

PR #610, opened 2024-02-28
Add convert_notebook

D4ve-R:
Add a ipynb notebook for converting models to onnx & uploading to the hub.


D4ve-R:
An [Open in Colab](https://openincolab.com/) button could be added to the Readme, to make it really easy to convert models

PR #614, opened 2024-02-29
filter cache puts to http(s)

patrickkettner:
I am bundeling my model within a chrome extension, and as a result I am getting a bunch of these warnings. Since Cache.put only takes http/s, I thought you may as well filter out other requests


patrickkettner:
the file this patched has since been rewritten and moved, so this is obsolete as written. I've opened https://github.com/huggingface/transformers.js/pull/1723 to fix the same problem in the current code. closing as obsolete

PR #620, opened 2024-03-02
added the missing pip install step

davies-w:
As we're not installing this (typically?) from a pip install itself, the requirements.txt will not be used, resulting in a failure if a different version of ONNX (for example) is installed. Added in the (obvious to everyone but me) step.


davies-w:
I give up. No one has reviewed it.

PR #637, opened 2024-03-10
Support 4bits quantization

hans00:
As title.

Notes:
- Execution require `onnxruntime>=1.16.2`
- Convert require `onnxruntime>=1.17.0`
- 4bits currently support CPU / CUDA / WebGPU(future) ([MatMulNBits](https://github.com/search?q=repo%3Amicrosoft%2Fonnxruntime+MatMulNBits&type=code))
- bnb4 currently support CPU / CUDA ([MatMulBnb4](https://github.com/search?q=repo%3Amicrosoft%2Fonnxruntime+BNB4&type=code))

### DIFF
--- a/scripts/convert.py
+++ b/scripts/convert.py
@@ -573,7 +573,7 @@ def main():
         if conv_args.reduce_range is not None:
             quantize_config['reduce_range'] = conv_args.reduce_range
 
-        quantize(conv_args.quantize_mode, [
+        quantize(QuantMode(conv_args.quantize_mode), [
             os.path.join(output_model_folder, x)
             for x in os.listdir(output_model_folder)
             if x.endswith('.onnx') and not x.endswith('_quantized.onnx')
--- a/scripts/convert.py
+++ b/scripts/convert.py
@@ -16,6 +16,7 @@ from transformers import (
 
 import onnx
 from optimum.exporters.onnx import main_export, export_models
+from optimum.onnx.graph_transformations import check_and_save_model
 from optimum.exporters.tasks import TasksManager
 from onnxruntime.quantization import (
     quantize_dynamic,
@@ -271,17 +272,6 @@ def get_operators(model: onnx.ModelProto) -> Set[str]:
     return operators
 
 
-def save_model(model, save_path):
-    # Should use external data if the model is larger than 2GB
-    if model.ByteSize() >= 2147483648:
-        onnx.external_data_helper.convert_model_to_external_data(
-            model,
-            all_tensors_to_one_file=True,
-            location=Path(save_path).name + "_data",
-        )
-    onnx.save_model(model, save_path)
-
-
 def quantize(mode, model_names_or_paths, **quantize_kwargs):
     """
     Quantize the weights of the model from float32 to int8 to allow very efficient inference on modern CPU
@@ -350,7 +340,7 @@ def quantize(mode, model_names_or_paths, **quantize_kwargs):
                 is_symmetric=quantize_kwargs.get('is_symmetric', True),
             )
             quant.process()
-            save_model(quant.model, save_path)
+            check_and_save_model(quant.model.model, save_path)
             del quant
         
         elif mode == QuantMode.BNB4:
@@ -360,7 +350,7 @@ def quantize(mode, model_names_or_paths, **quantize_kwargs):
                 quant_type=quantize_kwargs.get('quant_type', MatMulBnb4Quantizer.NF4),
             )
             quant.process()
-            save_model(quant.model, save_path)
+            check_and_save_model(quant.model.model, save_path)
             del quant
         
         else:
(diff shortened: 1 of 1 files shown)


xenova:
Thanks so much! This is amazing! I think it would be good to also name/suffix these models slightly differently, so you can export (and use) multiple versions.

bil-ash:
@hans00 Can the bnb4 quantization be used with webgpu backend?

xenova:
@bil-ash ORT recently added support for `MatMulNBit` and will be available in v1.17.3. You can already access it with [this version](https://www.npmjs.com/package/onnxruntime-web/v/1.18.0-dev.20240309-efad5bbc5a).

scripts/convert.py
```diff
@@ -158,6 +165,12 @@ class ConversionArguments:
             "help": "Whether to quantize the model."
         }
     )
+    quantize_mode: str = field(
+        default=QuantMode.BIT8,
+        metadata={
+            "help": "Quantization mode to use. Options are: int4, int8, bnb4"
```
xenova, line 171: mismatched names between this and `QuantMode.*`

hans00:
> Also, to help test, would you like to create a demo model on the HF Hub which contains all these different variants?

Here is my convertion:
https://huggingface.co/BricksDisplay/stablelm-2-1_6b-q4
https://huggingface.co/BricksDisplay/stablelm-2-1_6b-bnb4
https://huggingface.co/BricksDisplay/phi-1_5-q4
https://huggingface.co/BricksDisplay/phi-1_5-bnb4

As my test (Windows with DML, CPU EP) using StableLM-2-1.6b
Q8: 0.95 token/s
Q4: 2.46 token/s
BNB4: 0.44 token/s

hans00:
For benchmark, the quantized `all-MiniLM-L6-v2` model is
https://huggingface.co/BricksDisplay/all-MiniLM-L6-v2-q4
https://huggingface.co/BricksDisplay/all-MiniLM-L6-v2-bnb4

xenova:
> Did you mean the model file name like model_quantized_int4.onnx or model_quantized_nf4.onnx ?

Yes, exactly :) I think this PR can help standardize which suffixes we should use (as `quantized` will be overloaded). It might be worthwhile aligning this with other libraries' naming conversions. Do you have any thoughts on this?

Also, just checking, but the [bnb4 version](https://huggingface.co/BricksDisplay/all-MiniLM-L6-v2-bnb4/tree/main/onnx) has model sizes which are the same as fp16 (only half the size of the original). Is this correct?

hans00:
> Yes, exactly :) I think this PR can help standardize which suffixes we should use (as quantized will be overloaded). It might be worthwhile aligning this with other libraries' naming conversions. Do you have any thoughts on this?

I have no idea.
My naming will be `{orig_name}_quantized_{quant_type}.onnx`

> Also, just checking, but the [bnb4](https://huggingface.co/BricksDisplay/all-MiniLM-L6-v2-bnb4/tree/main/onnx) and [q4](https://huggingface.co/BricksDisplay/all-MiniLM-L6-v2-q4/tree/main/onnx) versions have model sizes which are the same as fp16 (only half the size of the original). Is this correct?

The dtype is not change, just convert from `sentence-transformers/all-MiniLM-L6-v2`.
The `MatMul` op should quantize to INT4 or NF4, others unchanged (maybe FP32)

scripts/convert.py
```diff
@@ -5,6 +5,8 @@
 from dataclasses import dataclass, field
 from typing import Optional, Set
 from tqdm import tqdm
+from enum import Enum
+from pathlib import Path
```
xenova, line 9:
```suggestion
```

scripts/convert.py
xenova: The names are still mismatched 👀 (int4 and int8, specifically)

scripts/convert.py
```diff
@@ -270,11 +286,13 @@ def quantize(model_names_or_paths, **quantize_kwargs):
 
     quantize_config = dict(
         **quantize_kwargs,
+        quantize_mode=mode.value,
```
xenova, line 289: This would prevent multiple quantizations to be produced per model.

xenova:
Thanks for this contribution 🤗! Since 4-bit quantization will only be supported in v3, I've merged it into [that branch](https://github.com/xenova/transformers.js/pull/545/commits/9efbada2b97fdb22fb0775ebbe30d93371cb83a3).

xenova:
@hans00 Have you by any chance experimented with q4f16 formats (model is and runs in fp16, with int4 block weight quantization)? That's the difference between:
- (fp32) https://huggingface.co/Xenova/Phi-3-mini-4k-instruct, and
- (fp16) https://huggingface.co/Xenova/Phi-3-mini-4k-instruct_fp16

PR #640, opened 2024-03-11
Update package.json to latest sharp

casret:
We were having problems installing this package in our environment because of sharp binary dependecies.  Sharp has updated the package to pull in pre-built binaries which fixed that problem for us.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "homepage": "https://github.com/xenova/transformers.js#readme",
   "dependencies": {
     "onnxruntime-web": "1.14.0",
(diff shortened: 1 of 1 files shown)


xenova:
Hi there 👋 We are aware of this, and these updates will all be made to the v3 branch/PR ([here](https://github.com/xenova/transformers.js/pull/545)). You can test it out with
```
npm install xenova/transformers.js#v3
```

Closing in favour of that PR. Also, a duplicate of https://github.com/xenova/transformers.js/pull/463

PR #643, opened 2024-03-12
Return-offset-mapping-using-tokenizer

guptanaman011:
This PR introduces support for offset mapping during tokenization, aiding tasks like Named Entity Recognition by aligning predicted tags with original sentence positions.


xenova:
Unfortunately, this PR introduces breaking changes, since the return types for the functions are now objects instead of `string[]`.

PR #645, opened 2024-03-13
Fix read data from blob URI

hans00:
As title

### DIFF
--- a/src/utils/hub.js
+++ b/src/utils/hub.js
@@ -167,7 +167,7 @@ function isValidHttpUrl(string, validHosts = null) {
     if (validHosts && !validHosts.includes(url.hostname)) {
         return false;
     }
-    return url.protocol === "http:" || url.protocol === "https:";
+    return url.protocol === "http:" || url.protocol === "https:" || url.protocol === "blob:";
 }
 
 /**


xenova (changes requested):
Unfortunately this breaks the purpose of the function this check resides in: to check whether a URL is valid http or https URL.

I've scanned through the code and I don't see why it shouldn't work though - what error do you get on your side?

hans00:
> I've scanned through the code and I don't see why it shouldn't work though - what error do you get on your side?

My self-reviewed the code, I noticed the root cause.
My environment is Electron and for some reason, I'm enable the `nodeIntegration`.
So that it will use FS read.

src/utils/hub.js
```diff
@@ -150,6 +150,22 @@ class FileResponse {
     }
 }
 
+/**
+ * Determines whether the given string is a valid Blob URL.
+ * @param {string|URL} string The string to test for validity as an Blob URL.
+ * @returns {boolean} True if the string is a valid Blob URL, false otherwise.
+ */
+function isValidBlobUrl(string) {
+    // https://stackoverflow.com/a/43467144
```
xenova, line 159:
```suggestion
```

src/utils/hub.js
```diff
@@ -150,6 +150,22 @@ class FileResponse {
     }
 }
 
+/**
+ * Determines whether the given string is a valid Blob URL.
+ * @param {string|URL} string The string to test for validity as an Blob URL.
+ * @returns {boolean} True if the string is a valid Blob URL, false otherwise.
+ */
+function isValidBlobUrl(string) {
```
xenova, line 158: How about abstracting the `isValidHttpURL` function into something like:
```js
function isValidUrl(string, protocols=null) {
  let url;
  
  try {
    url = new URL(string);
  } catch (_) {
    return false;  
  }

  return !protocols || protocols.includes(protocols)
}

function isValidHttpUrl(string) {
    return isValidUrl(string, ["http:", "https:"]);
}
```

and then the other code becomes:
```js
if (env.useFS && !isValidUrl(urlOrPath, ["http:", "https:", "blob:"])) {
```

xenova (changes requested):
```js
!isValidHttpUrl(urlOrPath) && !isValidBlobUrl(urlOrPath)
```
would parse the URL twice.

xenova:
Thanks for the updates. Could you also provide a test-case for this? Perhaps in `tests/utils.test.js`?

xenova:
Just an update: This will be merged and included in the next release. 😇 I'm just finalizing a few other things first.

xenova (changes requested):
Reviewing this fully now, and I think it's best to not expose the `isValidURL` function (via `export`). Instead, the unit tests should rather test the `getFile` function, which is the only place `isValidURL` is used.

For example, a test for blobs can be done with:
```js
const obj = { hello: "world" };
const blob = new Blob([JSON.stringify(obj, null, 2)], {
    type: "application/json",
});
const url = URL.createObjectURL(blob);
const result = await getFile(url);

// Then check obj and result are equivalent
```

xenova (approved):
Thanks! 🚀 Looks good :)

Running unit tests now, and you might just need to rebase off main, since there was a CI issue introduced by mistral gating their models.

PR #680, opened 2024-04-04
Add utils : save_audio_to_wav

Th3G33k:
Example use :
```js
const synthesizer = await pipeline('text-to-speech', 'Xenova/mms-tts-eng');
const output = await synthesizer('Hello, my dog is cute');
save_audio_to_wav(output.audio, output.sampling_rate, "audio.wav")
```

### DIFF
--- a/src/pipelines.js
+++ b/src/pipelines.js
@@ -61,7 +61,8 @@ import {
     round,
 } from './utils/maths.js';
 import {
-    read_audio
+    read_audio,
+    RawAudio
 } from './utils/audio.js';
 import {
     Tensor,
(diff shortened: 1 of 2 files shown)


Th3G33k:
Example use :
```js
const synthesizer = await pipeline('text-to-speech', 'Xenova/mms-tts-eng');
const output = await synthesizer('Hello, my dog is cute');
save_audio_to_wav(output.audio, output.sampling_rate, "audio.wav")
```

xenova:
Thanks for this! 🤗 Could you provide a reference for where this code was adapted from? Another thing which should be updated is that this should work if running from node.js.

Also, it might even be a good idea to wrap the output of these pipelines into a `Audio` class, which has a `.save(path)` method. This will simplify the usage a lot, and will align well with the `RawImage` class.

```js
const synthesizer = await pipeline('text-to-speech', 'Xenova/mms-tts-eng');
const output = await synthesizer('Hello, my dog is cute');
output.save("audio.wav");
```

PR #683, opened 2024-04-05
Add binary & scalar embedding quantization support to Transformers.js

### LINKED ISSUE #681: [feat] Add binary & scalar embedding quantization support to Transformers.js
### Feature request

Add binary & scalar quantization support

We should extract the algorithm to quantize the embeddings from the PR below from sentence-transformers and add it transformers.js so the feature-extraction pipeline can support binary vector search

Either adding a quantize-output or binary-output to the pipeline or we can have helper method that quantizes the tensor so this solution can be applied to other parts of the codebase.

https://github.com/UKPLab/sentence-transformers/pull/2549

### Motivation

Given performance gains for binary vector embeddings additional quantization helper methods can be useful for client-side vector search to reduce memory footprint

### Your contribution

I plan on making a PR, but note, I am slow to open source world still and learning the transformers.js repo as fast as I can.

@xenova may be able to more quickly parse the repo from above and add the methods along with test cases in the style of the repo quicker than me.

However I am working on a poc for using vector embeddings on a side project so I can contribute that work and repo or readme when I finish it.

jonathanpv:
This PR addressed the below issue, I've also added an example project showcasing the binary embeddings and using hamming distance function as @xenova gave an example of earlier in the below discussion.

-  Closes #681

- Let me know if you'd like any changes to this, perhaps we can make another options object in javascript like a QuantizationOptions so it's obvious which parameters are grouped together logically

- I also added a texts === undefined check since I ran into an issue earlier where I was accidentally not passing in any texts and was met with an unfriendly error.

For some reason though I don't get the same similarity numbers as Xenova did on his example but the numbers are at least similar

Hopefully the UI can help debug some of my code if there needs to be changes.

Here's an image of it:

image

### DIFF
--- a/examples/binary-vector-search/.eslintrc.cjs
+++ b/examples/binary-vector-search/.eslintrc.cjs
@@ -1,0 +1,20 @@
+module.exports = {
+  root: true,
+  env: { browser: true, es2020: true },
+  extends: [
+    'eslint:recommended',
+    'plugin:react/recommended',
+    'plugin:react/jsx-runtime',
+    'plugin:react-hooks/recommended',
+  ],
+  ignorePatterns: ['dist', '.eslintrc.cjs'],
+  parserOptions: { ecmaVersion: 'latest', sourceType: 'module' },
+  settings: { react: { version: '18.2' } },
+  plugins: ['react-refresh'],
+  rules: {
+    'react-refresh/only-export-components': [
+      'warn',
+      { allowConstantExport: true },
+    ],
+  },
+}
--- a/examples/binary-vector-search/.gitignore
+++ b/examples/binary-vector-search/.gitignore
@@ -1,0 +1,24 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+lerna-debug.log*
+
+node_modules
+dist
+dist-ssr
+*.local
+
+# Editor directories and files
+.vscode/*
+!.vscode/extensions.json
+.idea
+.DS_Store
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
--- a/examples/binary-vector-search/index.html
+++ b/examples/binary-vector-search/index.html
@@ -1,0 +1,13 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Binary vector search w/ Transformers.js</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="/src/main.jsx"></script>
+  </body>
+</html>
--- a/examples/binary-vector-search/package-lock.json
+++ b/examples/binary-vector-search/package-lock.json
@@ -1,0 +1,5375 @@
+{
+  "name": "adaptive-retrieval",
+  "version": "0.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "adaptive-retrieval",
+      "version": "0.0.0",
+      "dependencies": {
+        "@xenova/transformers": "file:../../",
+        "react": "^18.2.0",
+        "react-dom": "^18.2.0"
+      },
+      "devDependencies": {
+        "@types/react": "^18.2.43",
+        "@types/react-dom": "^18.2.17",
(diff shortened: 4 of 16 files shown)


xenova:
Thanks for the PR! Would you mind separating out the example app from the core quantization logic? We can possibly integrate the example app at a later stage, but I think a good starting point for now is to just provide a bit of sample code.

jonathanpv:
> Thanks for the PR! Would you mind separating out the example app from the core quantization logic? We can possibly integrate the example app at a later stage, but I think a good starting point for now is to just provide a bit of sample code.

Ah, yep forgot to this a while back, here it is! Let me know of any changes you'd like

- https://github.com/xenova/transformers.js/pull/691

xenova:
Feel free to rebase off main (which includes your other PR: https://github.com/xenova/transformers.js/issues/681) 🤗

jonathanpv:
> This is awesome! Maybe a little more explanation in the docs would be good. Like what's the (performance) difference between binary and ubinary [here](https://huggingface.co/docs/transformers.js/main/en/api/utils/tensor#utilstensorquantizeembeddingstensor-precision--code-tensor-code) and what's the default right now (I guess none)? 
> 
> @xenova I'm a little confused that there is no option `normal` / `none`. Or is the idea that if you do not pass the parameter, none of the two options is applied and instead it's simply normal (float) precision?

yep idea is to not pass the param if you don't plan on using it, its set as a default param / named param

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

do-me:
Thanks for your reply. Found the related line [here](https://github.com/xenova/transformers.js/pull/683/files#diff-161ff30745072fdddb33e89b10d0c6ca2bb115bd389dd7499ad11a486ebec2abR1210): 

```javascript
    if (!['binary', 'ubinary'].includes(precision)) {
        throw new Error("The precision must be either 'binary' or 'ubinary'");
    }
```

I understand the idea of the paradigm. However, it causes silightly more work for downstream applications. In my case, I wanted to create a dropdown menu for UI users with 3 values: 
- binary 
- ubinary 
- normal/full/float (however you want to call it) 

I would always pass the `precision` param as it's the easiest to do in client-side JS. However, due to the current logic I need to add more complexity, checking whether a user selected `normal` mode and not pass the `precision` param in the function call in case. 
Is there any advantage in not including the `normal` mode or would you mind adding it?

jonathanpv:
> Thanks for your reply. Found the related line [here](https://github.com/xenova/transformers.js/pull/683/files#diff-161ff30745072fdddb33e89b10d0c6ca2bb115bd389dd7499ad11a486ebec2abR1210):
> 
> ```js
>     if (!['binary', 'ubinary'].includes(precision)) {
>         throw new Error("The precision must be either 'binary' or 'ubinary'");
>     }
> ```
> 
> I understand the idea of the paradigm. However, it causes silightly more work for downstream applications. In my case, I wanted to create a dropdown menu for UI users with 3 values:
> 
> * binary
> * ubinary
> * normal/full/float (however you want to call it)
> 
> I would always pass the `precision` param as it's the easiest to do in client-side JS. However, due to the current logic I need to add more complexity, checking whether a user selected `normal` mode and not pass the `precision` param in the function call in case. Is there any advantage in not including the `normal` mode or would you mind adding it?

Hmm good call out, the reason I added quantize was to make it explicit you'd be quantizing. It technically does add more complexity from a UI / code perspective

Trying to think if quantize yes/no would add another field besides precision, if so then quantizeOptions would need to be created, can't think of anything at the moment.

technically we could flatten out the param to just be 'precision' 

## to be super clear in your suggestion: 
your current scenario makes your UI render the dropdown conditionally based on if the 'quantize embedding' is set to true eg

```
checkbox:
checkbox label: quantize embedding?

dropdown:
if checkbox yes, then render precision options


vs just having precision his ui would be the three dropdown

precision dropdown:
ubinary
binary
float32 (full)
```

So basically right now the cases are:

# cases now: 4
```

quantize: false
precision: binary (this is ignored)

===

quantize: false
precision: ubinary (this is ignored)

===

quantize: true
precision: binary

===

quantize: true
precision: ubinary

```

# cases if flattend: 3
```
precision: binary
===
precision: ubinary
===
precision: float32 / full ?
```

do-me:
## UI

> your current scenario makes your UI render the dropdown conditionally based on if the 'quantize embedding' is set to true eg

Exactly. One dropdown with all 3 flattened options would be much easier to use than conditionally changing the UI and JSON I need to pass to the function.

-- 

By the way: there are now two `quantize/d` options in transformers.js, one for loading a quantized model in `pipeline` and one for your precision settings.

Th3G33k:
+1 with do-me

PR #686, opened 2024-04-06
Update worker.js - fix TTS example

flatsiedatsie:
I added the `dtype: 'fp32',` tip you gave me earlier, so the example should then work again.

### DIFF
--- a/examples/electron/package-lock.json
+++ b/examples/electron/package-lock.json
@@ -3066,9 +3066,9 @@
       }
     },
     "node_modules/ip": {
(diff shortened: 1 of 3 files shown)


xenova:
Please note that this will only work with transformers.js v3, which is still in [development](https://github.com/xenova/transformers.js/pull/545). I'll update the branch for this PR and merge into v3.

flatsiedatsie:
Thanks!

xenova:
Done! :)

PR #737, opened 2024-05-04
Auto rotate image

KTibow:
Tell Sharp to rotate the image based on the EXIF. Allows proper processing on images taken from phones.


xenova:
Thanks!

PR #684, opened 2024-04-06
Add custom task `register_pipeline`

Th3G33k:
Following feature request #659

Example use :

```js
import {
  Pipeline,
  read_audio,
  register_pipeline,
  ClapAudioModelWithProjection,
  AutoProcessor,
  pipeline
} from '@xenova/transformers';

class AudioFeatureExtractionPipeline extends Pipeline {
  constructor(options) {
    super(options);
  }
  async _call(input, kwargs = {}) {
    input = await read_audio(input)
    input = await this.processor(input);
    let { audio_embeds } = await this.model(input);
    return audio_embeds
  }
}

register_pipeline('audio-feature-extraction', {
  pipeline: AudioFeatureExtractionPipeline,
  model: ClapAudioModelWithProjection,
  processor: AutoProcessor,
  model_name: 'Xenova/larger_clap_music_and_speech'
})

let pipe = await pipeline('audio-feature-extraction');
let out = await pipe('https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/jfk.wav')
console.log(out)
```

PR #747, opened 2024-05-09
v3: custom pipelines

Th3G33k:
Add `get_text_embeddings` and `get_similarities` methods to `ImageFeatureExtractionPipeline`


xenova:
Thanks! Are these methods present in the python version of the image feature extraction pipeline? If so, can you please link to the original documentation? If not, please note that we're trying to match the python API as closely as possible, and may make it difficult to merge the PR.

Th3G33k:
I don't think there in an equivalent in Python transformers.

https://github.com/xenova/transformers.js/blob/880cd3ea657b3cd3b7a30fca5067838e312b9eba/src/models.js#L5816-L5821

However, I think it would be really handful to have those fonctions directly into the library transformers.js image feature extraction pipeline.

It's used in hugging face spaces and in examples model code
* [webgpu-mobileclip](https://huggingface.co/spaces/Xenova/webgpu-mobileclip)
* [semantic image search](https://huggingface.co/spaces/Xenova/semantic-image-search-client)
* [clip vit](https://huggingface.co/Xenova/clip-vit-base-patch16)
* [mobileclip](https://huggingface.co/Xenova/mobileclip_s0)

Th3G33k:
To not differ from the Python library, I will add the custom pipeline separetely, with register_pipeline. #684

PR #770, opened 2024-05-22
Update pipelines.js to allow for `token_embeddings` as well

NikhilVerma:
In recent examples of optimum pipeline export the feature extraction pipelines have their output state as `token_embeddings` instead of `last_hidden_state`.

So we should support this as well.

### DIFF
--- a/src/pipelines.js
+++ b/src/pipelines.js
@@ -1255,7 +1255,7 @@ export class FeatureExtractionPipeline extends (/** @type {new (options: TextPip
         // console.log(outputs)
 
         /** @type {Tensor} */
-        let result = outputs.last_hidden_state ?? outputs.logits;
+        let result = outputs.last_hidden_state ?? outputs.logits ?? outputs.token_embeddings;
         if (pooling === 'none') {
             // Skip pooling
         } else if (pooling === 'mean') {
--- a/src/pipelines.js
+++ b/src/pipelines.js
@@ -3283,3 +3283,4 @@ async function loadItems(mapping, model, pretrainedOptions) {
 
     return result;
 }
+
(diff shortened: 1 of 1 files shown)


xenova:
Can you give an example of a model which has this as an output name?

NikhilVerma:
@xenova Sorry I should have provided. this is the command I ran

```
optimum-cli export onnx --model sentence-transformers/all-MiniLM-L6-v2 ./sbert/
```

xenova:
Since the model returns two [outputs](https://github.com/huggingface/optimum/blob/e0f58121140ce4baa01919ad70a6c13e936f7605/optimum/exporters/onnx/model_configs.py#L851-L855) (`token_embeddings` and `sentence_embedding`), I think it will be a good idea to try get the `sentence_embedding` if it is present and the user does not specify a pooling method. Otherwise, if pooling is set, we compute it the normal way. We should also allow the user to normalize their sentence embeddings with the `normalize` option.

Is this something you'd like to add to the PR?

src/pipelines.js
```diff
@@ -3282,4 +3282,4 @@ async function loadItems(mapping, model, pretrainedOptions) {
     }
 
     return result;
-}
\ No newline at end of file
+}
```
xenova, line 3285:
```suggestion
}

```

src/pipelines.js
```diff
@@ -3282,4 +3282,5 @@ async function loadItems(mapping, model, pretrainedOptions) {
...
+
```
xenova, line 3286:
```suggestion
}
```

xenova, line 3286:
```suggestion
}
```

xenova, line 3286:
```suggestion
}
```

xenova:
^^ Will merge for now, but could be a good follow-up PR. Let me know if you're interested!

Issue #803, opened 2024-06-10
Decoding Tokens added by the user for Whisper models 


aravindMahadevan:
### Feature request

Support decoding user defined added tokens that get added to end of the tokenizer's vocabulary for Whisper based models. This requires modifying the if statement in [_decode_asr](https://github.com/xenova/transformers.js/blob/da2688626d7812ad1ea47fd304c2072cc685051b/src/tokenizers.js#L3700) to make this work. 

### Motivation

Motivation for this proposal is to have feature parity with the tokenizers.decode which is able to decode user added tokens. 

### Your contribution

To support this feature, we just need to modify the if statement in [_decode_asr](https://github.com/xenova/transformers.js/blob/da2688626d7812ad1ea47fd304c2072cc685051b/src/tokenizers.js#L3700) from `token >= timestamp_begin` to `token >= timestamp_begin && token <= timestamp_end` where `timestamp_end = this.model.convert_tokens_to_ids(["<|30.00|>"])[0]`.

### Why this should work: 
When a user adds a new token to the tokenizer, it gets placed at the end of the tokenizer's vocabulary.  The last 1500 vocab tokens in whisper-tiny, whisper-tiny.en, whisper-small.en, whisper-small, whisper-base, whisper-base.en, whisper-large, and whisper-large-v2 correspond to timestamp tokens from `"<|0.00|>"` to `"<|30.00|>"`. By bounding the if statement condition from `token >= timestamp_begin` to `token >= timestamp_begin && token <= timstamp_end`, we will ensure that added user tokens will be decoded as regular tokens as the condition will evaluate to False and we will go to the [else block](https://github.com/xenova/transformers.js/blob/da2688626d7812ad1ea47fd304c2072cc685051b/src/tokenizers.js#L3757)

xenova:
Good spot! Feel free to submit a PR for this. Thanks! 🤗

aravindMahadevan:
@xenova submitted the PR with a fix!

PR #804, opened 2024-06-11
Fixes Issue #803

aravindMahadevan:
Modify the _decode_asr method to support decoding user defined token in Whisper based models. Addresses issue in #803


src/tokenizers.js
```diff
@@ -3610,6 +3610,7 @@ export class WhisperTokenizer extends PreTrainedTokenizer {
         let chunk = new_chunk();
         let time_offset = 0.0;
         const timestamp_begin = this.model.convert_tokens_to_ids(["<|notimestamps|>"])[0] + 1;
+        const timestamp_end = this.model.convert_tokens_to_ids(["<|30.00|>"]);
```
xenova, line 3613:
```suggestion
        const timestamp_end = this.model.convert_tokens_to_ids(["<|30.00|>"])[0];
```
to get the token id of a single element

xenova (changes requested):
Looks good! Just one thing:

src/tokenizers.js
aravindMahadevan: Commited suggestion, thanks for catching!

aravindMahadevan:
@xenova committed the fix, let me know if there is anything else needed!

avinashr175 (approved):
lgtm!

xenova:
Thanks! Will merge after tests pass. By the way, do you have an example of a whisper model which such tokens? Might be good to add a test.

aravindMahadevan:
Hi @xenova, there are two tests failing due to a few test models not having the <30.00> token. Here's one such example:  https://huggingface.co/Xenova/whisper-small/resolve/output_attentions/tokenizer.json. 

Do you have any suggestions on how to overcome this issue? This token does exist in all of the base whisper models.

xenova:
We could probably use the time precision (0.02) to calculate the offset: 30/0.02 + 1 = 1501 tokens (50364 -> 51864). Another fix is to simply update the tokenizer.json.

aravindMahadevan:
Hi @xenova , that was a good suggestion and I have updated the logic to the following: `const timestamp_end = timestamp_begin + total_timestamp_tokens`
`const total_timestamp_tokens = (30.00 - 0.00) / 0.02`

This logic should work on both English only Whisper and multilingual Whisper variants. The beginning timestamp of 0.00 is equal to token id 50363 and 50364 in the English only Whisper variants and Multilingual Whisper variants respectively. Similarly, the final timestamp token id of 30.00 is equal to 51863 and 51864 in the English only Whisper variants and Multilingual Whisper variants respectively as well. In both cases, the final timestamp token id occurs at an offset of 1500 which is what total_timestamp_tokens evaluates to. 

I do not have a model with added tokens that I can share publicly. I did find this model https://huggingface.co/oza75/whisper-bambara-asr-001 which has added tokens after the final timestamp but it's a special token.

aravindMahadevan:
Hey @xenova, are there any more changes that should be made to this PR?

radai-jamescheng:
@xenova Hi! I believe all test cases are passing, is there anything else needed before we can merge this change? Would you be able to take one last look? Thanks!

xenova (approved):
Apologies for the delay... This PR got lost in my notifs...

Thanks again! 🙏

src/tokenizers.js
```diff
@@ -3610,6 +3610,11 @@ export class WhisperTokenizer extends PreTrainedTokenizer {
         let chunk = new_chunk();
         let time_offset = 0.0;
         const timestamp_begin = this.model.convert_tokens_to_ids(["<|notimestamps|>"])[0] + 1;
+	// Whisper timestamp tokens start from 0.00 and go to timestamp 30.00 in 0.02 increments. 
+	// We can calculate the last time stamp token as timestamp_begin plus the number of tokens 
+	// tokens from 0.00 to 30.00 which is 1500. 
+	const total_timestamp_tokens = (30.00 - 0.00) / 0.02;
+	const timestamp_end = timestamp_begin + total_timestamp_tokens;
```
xenova, line 3617:
```suggestion
        const timestamp_begin = this.timestamp_begin;
        // Whisper timestamp tokens start from 0.00 and go to timestamp 30.00 in 0.02 increments. 
        // We can calculate the last time stamp token as timestamp_begin plus the number of tokens 
        // tokens from 0.00 to 30.00 which is 1500. 
        const total_timestamp_tokens = 1500; // (30.00 - 0.00) / 0.02
        const timestamp_end = timestamp_begin + total_timestamp_tokens;
```

PR #808, opened 2024-06-11
Support transformers original behavior to load processors

hans00:
As title.

I'm trying implement custom processor which has tokenizer.
The original behavior will help to load it.

### DIFF
--- a/src/processors.js
+++ b/src/processors.js
@@ -2022,6 +2022,195 @@ export class ClapFeatureExtractor extends FeatureExtractor {
 
 export class SpeechT5FeatureExtractor extends FeatureExtractor { }
 
+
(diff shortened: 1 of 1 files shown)


xenova:
Thanks for this! This is exactly what I was planning to do for v3. Can you give the model id of the processor/tokenizer you are loading so I can add a unit test?

hans00:
> Thanks for this! This is exactly what I was planning to do for v3. Can you give the model id of the processor/tokenizer you are loading so I can add a unit test?

The processor still in WIP.
https://github.com/mybigday/Bert-VITS2/tree/convert-hf/hf

this model uses multiple tokenizer.

PR #812, opened 2024-06-17
Implement numerically stable log_softmax()

taha-yassine:
A numerically stable implementation of `log_softmax()`. It's similar to the current implementation of `softmax()` and is what's used in PyTorch and SciPy too.

```javascript
log_softmax([1000,1])

// Current implementation
// > Array [ 0, -Infinity ]

// Proposed implementation
// > Array [ 0, -999 ]
```

### DIFF
--- a/src/utils/maths.js
+++ b/src/utils/maths.js
@@ -162,8 +162,10 @@ export function log_softmax(arr) {
     const maxVal = max(arr)[0];
 
     // Compute the sum of the exponentials
-    // @ts-ignore
-    const sumExps = arr.reduce((acc, val) => acc + Math.exp(val - maxVal), 0);
+    let sumExps = 0;
+    for(let i = 0; i < arr.length; ++i) {
+        sumExps += Math.exp(arr[i] - maxVal);
+    }
 
     // Compute the log of the sum
     const logSum = Math.log(sumExps);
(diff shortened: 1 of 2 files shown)


xenova:
Thanks so much! 🤗 Can you add a unit test for this to [tests/maths.test.js](https://github.com/xenova/transformers.js/blob/30720089c6215f71b519c17369404cdfd14c32e0/tests/maths.test.js#L30-L41)? One normal case and one case where the unstable version will break? 🔥

taha-yassine:
Done :)

src/utils/maths.js
```diff
@@ -158,11 +158,18 @@ export function softmax(arr) {
...
  */
 export function log_softmax(arr) {
-    // Compute the softmax values
-    const softmaxArr = softmax(arr);
+    // Compute the maximum value in the array
+    const maxVal = max(arr)[0];
+
+    // Compute the sum of the exponentials
+    // @ts-ignore
+    const sumExps = arr.reduce((acc, val) => acc + Math.exp(val - maxVal), 0);
```
xenova, line 166:
```suggestion
    // Compute the sum of the exponentials
    let sumExps = 0;
    for(let i = 0; i < arr.length; ++i) {
        sumExps += Math.exp(arr[i] - maxVal);
    }
```

xenova (approved):
As seen in https://github.com/xenova/transformers.js/pull/667, a standard for loop is faster for computing sums, so I've updated that. I tested both and indeed it is faster:
image
(code block 1 = reduce; code block 2 = normal for loop)

xenova:
(ignore failing tests; related to something else)

PR #864, opened 2024-07-25
v3: fix bundler config for latest ORT

fs-eire:
This PR fixes bundler config for latest ORT.

After a few months' work, onnxruntime-web has finally stabilized its library export. There is not going to be big changes to this part in near future.

Discussion:
- I disabled default wasm path override to CDN. Latest onnxruntime-web (ort.webgpu.bundle.min.js) allows bundler to pick the .wasm files by default, and this is verified locally for webpack and Vite(need a workaround for 'development' mode).
  - as mentioned above, the vite.config.js needs update. I only did that for segment-anything-client in this PR, but it may be necessary to update the config for all configuration in the code base.
  - transformerjs may offer users a choice of either using CDN or using self-hosted artifacts (by setting `env.wasm.wasmPaths = undefined`)


webpack.config.js
```diff
@@ -31,29 +31,25 @@ function buildConfig({
...
-            // we can access the files with /dist/...
-            [`dist/transformers${name}`]: './src/transformers.js',
-            [`dist/transformers${name}.min`]: './src/transformers.js',
+            [`transformers${name}`]: './src/transformers.js',
+            [`transformers${name}.min`]: './src/transformers.js',
         },
         output: {
             filename: `[name]${suffix}`,
-            path: __dirname,
+            path: path.join(__dirname, 'dist'),
```
fs-eire, line 39: output path is changed so that webpack can correctly put .wasm files into output folder.

webpack.config.js
```diff
@@ -31,29 +31,25 @@ function buildConfig({
...
             library: {
                 type,
             },
+            assetModuleFilename: '[name][ext]',
```
fs-eire, line 43: this config is required for the .wasm files (which are considered as asset dependencies) to preserve their names. Otherwise, they will be renamed to [hash][ext].

webpack.config.js
```diff
@@ -31,29 +31,25 @@ function buildConfig({
...
+            chunkFormat: 'module',
+        },
+        module: {
+            parser: {
+                javascript: {
+                    importMeta: false
+                }
+            }
         },
```
fs-eire, line 52: This is required to preserve `import.meta.url` in the code. Otherwise webpack will replace the expression to a static string starting with "file://", which breaks worker/wasm loading.

examples/segment-anything-client/vite.config.js
```diff
@@ -1,6 +1,17 @@
...
-  build: {
-    target: 'esnext'
+export default defineConfig(env => {
+  const config = {
+    build: {
+      target: 'esnext'
+    }
+  };
+
+  if (env.mode === 'development') {
```
fs-eire, line 9: Vite works perfect in production mode (`vite build`)

xenova: thanks!

webpack.config.js
xenova: thanks!

xenova:
Thanks! Merged into `@huggingface/transformers` branch 👍

Issue #883, opened 2024-08-07
Failed to run Xenova/wav2vec2-bert-CV16-en on transformer.js-v3 webgpu


axinging:
### System Info

xenova/transformers.js#v3

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension
- [ ] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

I tried to reproduce the problem depicted by https://github.com/xenova/transformers.js/pull/545#issuecomment-2003556475, @young-developer.  

At first, I run the model on transformers@2.17.2, it runs well (not webgpu):
```
import { pipeline } from 'https://cdn.jsdelivr.net/npm/@xenova/transformers@2.17.2';
// Create an Automatic Speech Recognition pipeline
const transcriber = await pipeline('automatic-speech-recognition', 'Xenova/wav2vec2-bert-CV16-en');
// Transcribe audio
const url = 'https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/jfk.wav';
const output = await transcriber(url);
// { text: 'and so my fellow americans ask not what your country can do for you ask what you can do for your country' }
console.log(output);
```

Then I turn this to webgpu with xenova/transformers.js#v3, this runs into error:  Can't create a session. ERROR_CODE: 1, ERROR_MESSAGE: Deserialize tensor wav2vec2_bert.encoder.layers.15.conv_module.depthwise_layer_norm.bias failed.Failed to load external data file "./model.onnx_data", error: Module.MountedFiles is not available. (Same error when turn device to 'wasm')

```
import {pipeline} from '@xenova/transformers';

// Create an Automatic Speech Recognition pipeline
const transcriber = await pipeline(
    'automatic-speech-recognition', 'Xenova/wav2vec2-bert-CV16-en', {
      device: 'webgpu',
      dtype: 'fp32',  // or 'fp16'
    });

// Transcribe audio
const url =
    'https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/jfk.wav';
const output = await transcriber(url);
// { text: 'and so my fellow americans ask not what your country can do for you
// ask what you can do for your country' }
console.log(output);
```

I am not sure about the latest status of this problem, hoping @young-developer could help to comment.

### Reproduction

See above description.

young-developer:
[...]

PR #884, opened 2024-08-08
 Add missing options in pipeline API

gyagp:
This is to add missing options in pipeline API to align with from_pretrained API.
Bug #883 can be fixed with this. That model has a separate data file, so use_external_data_format has to be supported and designated to true.


src/pipelines.js
```diff
@@ -1,15 +1,15 @@
 /**
  * @file Pipelines provide a high-level, easy to use, API for running machine learning models.
- * 
+ *
```
gyagp, line 3: My editor trims the tailing spaces. Please let me know if you don't like this change. For review, you may set the option to "Hide whitespace".

gyagp:
@xenova Please review this change, thanks!

xenova:
Thanks! To keep the diff small (and consistent across other files), can you add back the whitespace for those lines (I am adding prettier which should fix this later)?

gyagp:
> Thanks! To keep the diff small (and consistent across other files), can you add back the whitespace for those lines (I am adding prettier which should fix this later)?

Updated. 
Sorry that my local changes are a bit messy, so I had to do a force push. I think it's because I resolved some merge conflicts using GitHub directly.

PR #890, opened 2024-08-14
Initial WebNN Support

ibelem:
This is the initial PR to support [WebNN API](https://webnn.dev).

### User Code

```

    let dataType = 'fp16';
    let provider = 'webnn';
    let deviceType = 'gpu';

    let options = {
      dtype: dataType,
      device: provider,
      session_options: {
        executionProviders: [
          {
            name: provider,
            deviceType: deviceType
          },
        ],
        freeDimensionOverrides: {
          batch_size: 1,
          num_channels: 3,
          height: 224,
          width: 224
        }
      },
    };

    // Create an image classification pipeline
    const classifier = await pipeline('image-classification', 'xenova/resnet-50', options);
    const url = 'https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/tiger.jpg';
    const output = await classifier(url);
    // [
    //   { "label": "tiger, Panthera tigris", "score": 0.9231255054473877 },
    //   { "label": "tiger cat", "score": 0.07358699291944504 },
    //   { "label": "jaguar, panther, Panthera onca, Felis onca", "score": 0.00047683052252978086 },
    //   { "label": "leopard, Panthera pardus", "score": 0.00017269655654672533 },
    //   { "label": "lynx, catamount", "score": 0.00015724201512057334 }
    // ]
```

### WebNN Installation Guide

WebNN is being rapidly implemented in browsers. Before it formally ships, WebNN requires a compatible browser to run, and Windows* 11 v21H2 (DML 1.6.0) or higher.

- Download the latest [Google Chrome Canary](https://google.com/chrome/canary) or [Microsoft Edge Canary](https://www.microsoft.com/edge/download/insider) browser
- To enable WebNN, in your browser address bar, enter `about://flags`, and then press Enter. An Experiments page opens
- In the Search flags box, enter `webnn`. `Enables WebNN API` appears
- In the drop-down menu, select `Enabled`
- Relaunch your browser

### Known Issue

From Transformers.js perspective, it's better to improve the following float16 check code (current via WebGPU and WebGPU shader-f16) since fp16 support is not GPU device only.

- https://github.com/xenova/transformers.js/blob/v3/src/models.js#L178
- https://github.com/xenova/transformers.js/blob/v3/src/utils/dtypes.js#L12

@honry @huningxin @xenova PTAL.

### DIFF
--- a/src/backends/onnx.js
+++ b/src/backends/onnx.js
@@ -21,7 +21,7 @@ import { env, apis } from '../env.js';
 // NOTE: Import order matters here. We need to import `onnxruntime-node` before `onnxruntime-web`.
 // In either case, we select the default export if it exists, otherwise we use the named export.
 import * as ONNX_NODE from 'onnxruntime-node';
-import * as ONNX_WEB from 'onnxruntime-web/webgpu';
+import * as ONNX_WEB from 'onnxruntime-web/all';
 
 export { Tensor } from 'onnxruntime-common';
 
@@ -39,6 +39,9 @@ if (apis.IS_NODE_ENV) {
     ONNX = ONNX_WEB;
     if (apis.IS_WEBGPU_AVAILABLE) {
         supportedExecutionProviders.push('webgpu');
+    }
+    if(apis.IS_WEBNN_AVAILABLE) {
+        supportedExecutionProviders.push('webnn');
     }
     supportedExecutionProviders.push('wasm');
     defaultExecutionProviders = ['wasm'];
--- a/src/env.js
+++ b/src/env.js
@@ -33,6 +33,7 @@ const IS_BROWSER_ENV = typeof self !== 'undefined';
 const IS_WEBWORKER_ENV = IS_BROWSER_ENV && self.constructor.name === 'DedicatedWorkerGlobalScope';
 const IS_WEB_CACHE_AVAILABLE = IS_BROWSER_ENV && 'caches' in self;
 const IS_WEBGPU_AVAILABLE = typeof navigator !== 'undefined' && 'gpu' in navigator;
+const IS_WEBNN_AVAILABLE = typeof navigator !== 'undefined' && 'ml' in navigator;
 
 const IS_PROCESS_AVAILABLE = typeof process !== 'undefined';
 const IS_NODE_ENV = IS_PROCESS_AVAILABLE && process?.release?.name === 'node';
@@ -54,6 +55,9 @@ export const apis = Object.freeze({
 
     /** Whether the WebGPU API is available */
     IS_WEBGPU_AVAILABLE,
+
+    /** Whether the WebNN API is available */
+    IS_WEBNN_AVAILABLE,
 
     /** Whether the Node.js process API is available */
     IS_PROCESS_AVAILABLE,
--- a/src/utils/devices.js
+++ b/src/utils/devices.js
@@ -4,6 +4,7 @@ export const DEVICE_TYPES = Object.freeze({
     gpu: 'gpu',
     wasm: 'wasm',
     webgpu: 'webgpu',
+    webnn: 'webnn',
 });
 
 /**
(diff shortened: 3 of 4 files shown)


xenova:
Thanks for the PR! Since our goal is to align with the python library as closely as possible, we may need to separate out `device` and `provider` into separate parameters (don't worry, we'll make sure the correct format is passed to the ORT session).

Here are some proposals for the API:
```js
const classifier = await pipeline('image-classification', 'xenova/resnet-50', {
  dtype: 'fp32',
  device: 'gpu', // or 'npu'
  provider: 'webnn',
});
```

```js
const classifier = await pipeline('image-classification', 'xenova/resnet-50', {
  dtype: 'fp32',
  device: 'webnn-gpu', // e.g., or 'webnn-npu'
});
```

---

What do you think?

ibelem:
Thank you @xenova for starting the review. Yes, the current `options` is obviously redundant, I just wanted to get your feedback. :)

```
// Create a feature-extraction pipeline
const extractor = await pipeline('feature-extraction', 'Xenova/all-MiniLM-L6-v2', {
  dtype: 'fp32',
  device: 'webgpu', // <- Run on WebGPU
});
```

I think the `device` value can't be `gpu` or `npu` since it should be `webgpu` or `wasm`? To make the API simpler, I think the second proposal is better which doesn't require extra `provider` key.

xenova:
> BTW, WebNN needs freeDimensionOverrides for inputs with symbolic dimensions, please also make sure it can be passed to ORT web easily.

Yes this should be handled by specifying these values in the model's config.json under something like `{"transformers.js_config": { "session_options": ... }}`.

kungfooman:
I couldn't test this yet, just wondering: if we pick WebNN as provider, do we have a different set of supported ops again based on the browsers WebNN implementation?

I see three scenarios:

1) A model that used to work, suddenly doesn't work anymore (op not supported in browsers WebNN implementation)
2) A model that didn't work, suddenly works (browser supports more ops than e.g. ONNX WebGPU)
3) Completely the same

ibelem:
Thanks @xenova for the device detection and selection improvement between ff1853ce626a908e7c0e57d42c31c4a511dcc0ec [Initial WebNN Support](https://github.com/xenova/transformers.js/commit/ff1853ce626a908e7c0e57d42c31c4a511dcc0ec) and 5b2cac21143978e2d18cc6203edb2c4ded49918b [Improve WebNN selection](5b2cac21143978e2d18cc6203edb2c4ded49918b), now WebNN can be run successfully by using code below:

```
    import { pipeline } from './dists/upstream_v3_0819/transformers.js';
    let options = {
      dtype: 'fp16',
      device: 'webnn-cpu', // 'webnn-gpu' and upcoming 'webnn-npu'
      session_options: {
        freeDimensionOverrides: {
          batch_size: 1,
          num_channels: 3,
          height: 224,
          width: 224
        }
      },
    };

    // Create an image classification pipeline
    const classifier = await pipeline('image-classification', 'xenova/resnet-50', options);
    const url = 'https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/tiger.jpg';
    const output = await classifier(url);
    console.log(output);
```

@kungfooman Thanks for the great question! Currently WebNN is still adding support for new ops and new data types. We are keep updating the [Implementation Status of WebNN Operations](https://webmachinelearning.github.io/webnn-status/) based on the WebNN Spec and the browser implementation. 

Our team is also implementing and maintaining the WebNN EP of ONNX Runtime Web, if the op not supported in browsers WebNN implementation, there is a fallback mechanism that fallback these WebNN not supported ops to CPU EP (Wasm) of ONNX Runtime Web. WebNN is also adding WebGPU interops capabilities, interacting with WebGPU to support custom ops or cases that may not be supported by WebNN. Usually the more WebNN ops supported, the fewer subgraphs are split. The less partitions the better performance for WebNN.

WebNN will enable users to leverage multiple AI accelerators, including CPUs, GPUs, and NPUs, opening up new possibilities such as power-efficient AI inferencing on NPUs.

PR #945, opened 2024-09-26
use cuda by default and bump `onnxruntime-node` version

laptou:
- use cuda by default and bump `onnxruntime-node` version


laptou:
i'm having trouble running the tests on my system, so i'm not sure whether this breaks anything, but this patch is enough to speed up the `@xenova/transformers` package like 10x by making it use my GPU

PR #946, opened 2024-09-26
Update requirements.txt for the optimum script

tarekziade:
Numpy needs to stay under 2.0 (that was out recently)


xenova:
Hi @tarekziade! 👋 Can you confirm this change is still necessary with the latest (main) set of requirements? https://github.com/huggingface/transformers.js/blob/main/scripts/requirements.txt

PR #966, opened 2024-10-09
Add function to map tensor data array.

BritishWerewolf:
I was looking for a way to run a binary map over my tensors.

With this I can do something like this:
```js
const tensor = new Tensor('float32', [2, 4, 6, 8, 10, 12], [2, 3]);
const mapped = tensor.map(x => x <= 5 ? 0 : 1);
```

### DIFF
--- a/src/utils/tensor.js
+++ b/src/utils/tensor.js
@@ -215,8 +215,9 @@ export class Tensor {
      * @returns {Tensor} Returns `this`.
      */
     map_(callback) {
-        for (let i = 0; i < this.data.length; ++i) {
-            this.data[i] = callback(this.data[i], i, this.data);
+        const this_data = this.data;
+        for (let i = 0; i < this_data.length; ++i) {
+            this_data[i] = callback(this_data[i], i, this_data);
         }
         return this;
     }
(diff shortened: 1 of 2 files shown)


src/utils/tensor.js
```diff
@@ -198,6 +198,29 @@ export class Tensor {
...
+    /**
+     * Apply a callback function to each element of the tensor in place.
+     * @param {Function} callback - The function to apply to each element. It should take three arguments:
+     *                              the current element, its index, and the tensor's data array.
+     * @returns {Tensor} Returns `this`.
+     */
+    map_(callback) {
+        for (let i = 0; i < this.data.length; ++i) {
+            this.data[i] = callback(this.data[i], i, this.data);
+        }
```
xenova, line 220:
```suggestion
        const this_data = this.data;
        for (let i = 0; i < this_data.length; ++i) {
            this_data[i] = callback(this_data[i], i, this_data);
        }
```

xenova (changes requested):
Great idea! I've changed the base to the `v3` branch since that's what will be merged + released soon.

Minor nit: Dereferencing `this.data` before the for loop is slightly more efficient.

src/utils/tensor.js
BritishWerewolf: Thanks for the optimisation!  
JavaScript's silent pointers; I never think about this :smile:

PR #972, opened 2024-10-11
Add option to pad an image to a square

BritishWerewolf:
This PR is stacked on [PR 971](https://github.com/xenova/transformers.js/pull/971).  
On my fork, this was branched off the other PR, but I couldn't find a way to do this with PRs - sorry 😅

---

It's common to want to create a square image, currently we need to do some manual calculations.

With this PR, we can do something like this, with no arguments to just fill the blank area.
```js
const image = await RawImage.fromURL("https://picsum.photos/600/400");
const resized = await image.resize(300, null); // 300x200
const padded = await resized.padToSquare();    // 300x300
```

Or we can provide a value, and make the image pad to that size.
```js
const image = await RawImage.fromURL("https://picsum.photos/600/400");
const resized = await image.resize(300, null); // 300x200
const padded = await resized.padToSquare(400); // 400x400
```

Providing a value that is smaller than either the width or height will just return the image unchanged.
```js
const image = await RawImage.fromURL("https://picsum.photos/600/400");
const resized = await image.resize(300, null); // 300x200
const padded = await resized.padToSquare(100); // 300x200
```

### DIFF
--- a/src/utils/core.js
+++ b/src/utils/core.js
@@ -1,10 +1,10 @@
 
 /**
  * @file Core utility functions/classes for Transformers.js.
- * 
+ *
  * These are only used internally, meaning an end-user shouldn't
  * need to access anything here.
- * 
+ *
  * @module utils/core
  */
 
@@ -46,7 +46,7 @@ export function escapeRegExp(string) {
  * Check if a value is a typed array.
  * @param {*} val The value to check.
  * @returns {boolean} True if the value is a `TypedArray`, false otherwise.
- * 
+ *
  * Adapted from https://stackoverflow.com/a/71091338/13989043
  */
 export function isTypedArray(val) {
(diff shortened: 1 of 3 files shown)


xenova:
> On my fork, this was branched off the other PR, but I couldn't find a way to do this with PRs - sorry 😅

No worries - we will merge that one first 👍 


As for this PR, could you also provide a reference to the function that is in another image processing library (e.g., PIL) which is similar? It's useful to align with existing libraries to make sure we don't stray too far away from python standards. Thanks!

BritishWerewolf:
Actually, I am going to close this PR.  
I originally suggested it because I found myself needing it - despite it not being present elsewhere! - however I have not found myself using it since opening.  
Thanks.

PR #947, opened 2024-09-26
Support to load ONNX APIs based on JS runtime

kallebysantos:
Hey there 🫡
I've been working withing supabase team on https://github.com/supabase/edge-runtime/pull/368. 
Since `edge-runtime` is based on short-live sandboxed workers, the use of `transformers.js` was very expensive due the model caching and session state. At this point the original proposal was to replace all the JS libraries like `transformers.js` to a embedded runtime backend by rust. 

<details>
<summary>The Pipeline proposal</summary>

image

</details>

After some problems with `enc/dec` models I'd discuss with Supabase team, where @gregnr did the great suggestion of provide a more performative environment while keep using a well tested and maintained library like yours.

So we decided to move on it and export our custom `ONNX runtime APIs` to the `edge-runtime` environment, from the `transformers.js` side the only changed needed is to priority load the `ONNX` variable from `globalThis`. We'd follow the same API of `onnxruntime-custom` package so everything should work as expected.

<details>
<summary>The custom binding proposal</summary>

image

</details>

---

If you would like to test our runtime pls check my [tranformers.js integration branch](https://github.com/kallebysantos/edge-runtime/tree/transformers-js-ort-backend), at [this file](https://github.com/kallebysantos/edge-runtime/blob/transformers-js-ort-backend/examples/ort-rust-backend/index.ts). 
I also recommend to read the [DEVELOPERS.md](https://github.com/supabase/edge-runtime/blob/main/DEVELOPERS.md) section to get more information about how to run the environment.
But basically just need to:
```bash
$ ./scripts/run_dind.sh

$ curl --request POST \
  --url http://localhost:9998/ort-rust-backend \
  --header 'Content-Type: application/json' \
  --data '{
	"input": [
		"one day I will see the world",
		"I will never forgive u"
	]
}'
```

PR #976, opened 2024-10-16
Improvements on Exposed ORT support

kallebysantos:
Hey there 🙏,
Coming from #947, I did test the new `Exposed ORT` feature in `alpha-v20` and `alpha-21` releases. I'd notice that to make it work we must specify `'auto'` as inference `device`:

<details>
<summary>Current behaviour:</summary>

Without specify a device:
```js

let pipe = await pipeline('sentiment-analysis');
// The exposed runtime is loaded 
// but when trying to inference:
let out = await pipe('I love transformers!');
//Error: Unsupported device: "wasm". Should be one of: 
```

Explicit using `auto` as device:
```js

let pipe = await pipeline('sentiment-analysis', null, { device: 'auto' });

let out = await pipe('I love transformers!');
// [{'label': 'POSITIVE', 'score': 0.999817686}]
```
---

</details>

To solve this annoying config, the PR improves the environment support for `Exposed ORT` by implicitly selecting `'auto'` as the default fallback instead of `'wasm'`.
```js

let pipe = await pipeline('sentiment-analysis');

let out = await pipe('I love transformers!');
// [{'label': 'POSITIVE', 'score': 0.999817686}]
```

### DIFF
--- a/src/backends/onnx.js
+++ b/src/backends/onnx.js
@@ -59,11 +59,11 @@ const supportedDevices = [];
 /** @type {ONNXExecutionProviders[]} */
 let defaultDevices;
 let ONNX;
-const ORT_SYMBOL = Symbol.for('onnxruntime');
 
-if (ORT_SYMBOL in globalThis) {
-  // If the JS runtime exposes their own ONNX runtime, use it
-  ONNX = globalThis[ORT_SYMBOL];
+if (apis.IS_EXPOSED_RUNTIME_ENV) {
+    // If the JS runtime exposes their own ONNX runtime, use it
+    ONNX = globalThis[apis.EXPOSED_RUNTIME_SYMBOL];
+    defaultDevices = ['auto'];
 
 } else if (apis.IS_NODE_ENV) {
     ONNX = ONNX_NODE.default ?? ONNX_NODE;
--- a/src/env.js
+++ b/src/env.js
@@ -35,6 +35,9 @@ const IS_WEB_CACHE_AVAILABLE = IS_BROWSER_ENV && 'caches' in self;
 const IS_WEBGPU_AVAILABLE = typeof navigator !== 'undefined' && 'gpu' in navigator;
 const IS_WEBNN_AVAILABLE = typeof navigator !== 'undefined' && 'ml' in navigator;
 
+const EXPOSED_RUNTIME_SYMBOL = Symbol.for('onnxruntime');
+const IS_EXPOSED_RUNTIME_ENV = EXPOSED_RUNTIME_SYMBOL in globalThis;
+
 const IS_PROCESS_AVAILABLE = typeof process !== 'undefined';
 const IS_NODE_ENV = IS_PROCESS_AVAILABLE && process?.release?.name === 'node';
 const IS_FS_AVAILABLE = !isEmpty(fs);
@@ -58,6 +61,12 @@ export const apis = Object.freeze({
 
     /** Whether the WebNN API is available */
     IS_WEBNN_AVAILABLE,
+
+    /** Symbol from JS environment that exposes their own ONNX runtime */
+    EXPOSED_RUNTIME_SYMBOL,
+
+    /** Whether we are running in a JS environment that exposes their own ONNX runtime */
+    IS_EXPOSED_RUNTIME_ENV,
 
     /** Whether the Node.js process API is available */
     IS_PROCESS_AVAILABLE,
--- a/src/models.js
+++ b/src/models.js
@@ -166,7 +166,10 @@ async function getSession(pretrained_model_name_or_path, fileName, options) {
 
     // If the device is not specified, we use the default (supported) execution providers.
     const selectedDevice = /** @type {import("./utils/devices.js").DeviceType} */(
-        device ?? (apis.IS_NODE_ENV ? 'cpu' : 'wasm')
+        device ?? (
+          apis.IS_EXPOSED_RUNTIME_ENV ? 'auto' : (
+            apis.IS_NODE_ENV ? 'cpu' : 'wasm'
+        ))
     );
     const executionProviders = deviceToExecutionProviders(selectedDevice);
 


xenova:
Hi again 👋 Apologies for the late reply.

For this PR, could you explain the meaning/origin of the `"auto"` device? Is this an execution provider defined by the custom ORT runner?

kallebysantos:
> Hi again 👋 Apologies for the late reply.

That's ok, saw that you had a loot of work to do!
Btw congratulations `v3` is finally launched 🎉

> For this PR, could you explain the meaning/origin of the `"auto"` device? Is this an execution provider defined by the custom ORT runner?

Currently the runner will not look on it to execute, it does based on the [environment](https://github.com/supabase/edge-runtime/blob/388d2ea69729777f31426bb16d9766d67d902cee/crates/sb_ai/onnxruntime/session.rs#L85). So the `device` property should do nothing in this case. 

---

But the problem is that `transformers.js` tries to explicit use some `device`, like `"wasm"`, and  that is not possible to execute it without specify `"auto"`. 

Example, if we don't specify a device, or anything instead of `"auto"` it will throw:

```javascript
const pipe = await pipeline('feature-extraction', 'supabase/gte-small')
// Error: Unsupported device "wasm", Should be one of:    .

const pipe = await pipeline('feature-extraction', 'supabase/gte-small', { device: 'cpu' })
// Error: Unsupported device "cpu", Should be one of:    .

const pipe = await pipeline('feature-extraction', 'supabase/gte-small', { device: 'webgpu' })
// Error: Unsupported device "webgpu", Should be one of:    .
```

But if I do `"auto"` it works:
```javascript
const pipe = await pipeline('feature-extraction', 'supabase/gte-small',  { device: 'auto' })

// [.... embeddings ]
```

<details>
<summary>Custom ORT exposed:</summary>

```javascript
console.log(globalThis[Symbol.for("onnxruntime"])
/*
{
  Tensor: [class Tensor],
  env: {},
  InferenceSession: { create: [AsyncFunction: fromBuffer] }
}
*/
```

</details>


So the purpose of this PR is to automatically set `"auto"` when its running from some Custom ORT. To avoid this annoying config.
Also I did add some global flags to detect if is or not running from Custom ORT.

xenova:
Thanks for the additional context! 👍 Just on that note, what if an exposed ORT library does allow for different devices to be specified, and the user could choose the version they would like?

For Supabase's runner, specifically, I imagine it is running on CPU, right? So "cpu" could be mapped to the default device on their side perhaps?

kallebysantos:
> Thanks for the additional context! 👍 Just on that note, what if an exposed ORT library does allow for different devices to be specified, and the user could choose the version they would like?
> 
> For Supabase's runner, specifically, I imagine it is running on CPU, right? So "cpu" could be mapped to the default device on their side perhaps?

Sure, it makes sense. But this way they still need to manually specify a device?  
> -*I need to study more about how `transformers.js` handle the available devices.*

At this moment, could you give me any suggestion? about how can we achieve: the following, without it throws an error? 
```javascript
const pipe = await pipeline('feature-extraction', 'supabase/gte-small')
```
Do I need to export some property that says  *"Hey `trasnformers.js` I'm using `"cpu"` as default so pls map to it"*  ?

xenova:
Note that there is no "auto" execution provider in onnxruntime-web/onnxruntime-node (and this is a layer added on top by Transformers.js).

That said, I think I see what the problem is: for the custom runtime, we don't specify `supportedDevices` or `defaultDevices`, which will be sent to the `executionProviders` option in `createInferenceSession`

https://github.com/huggingface/transformers.js/blob/2c92943c36e17294fec92ddaefbd4350069b4335/src/backends/onnx.js#L91-L92

device -> execution provider mapping:

https://github.com/huggingface/transformers.js/blob/2c92943c36e17294fec92ddaefbd4350069b4335/src/backends/onnx.js#L117-L136

---

So, I think the solution here would be to set them both to `undefined` by default (instead of `[]` as it is currently), since then the custom ORT environment will use its defaults: https://onnxruntime.ai/docs/api/js/interfaces/InferenceSession.SessionOptions.html#executionProviders

kallebysantos:
Hey Joshua, thanks for your suggestion

> So, I think the solution here would be to set them both to undefined by default (instead of [] as it is currently), since then the custom ORT environment will use its defaults..

Yes it worked, https://github.com/huggingface/transformers.js/pull/976/commits/42ae1fecc482daf8a4ca88949126a03de4cd4907

xenova (approved):
Thanks for iterating! ✅ Looks good!

src/env.js
```diff
@@ -59,6 +62,12 @@ export const apis = Object.freeze({
     /** Whether the WebNN API is available */
     IS_WEBNN_AVAILABLE,
 
+    /** Symbol from JS environment that exposes their own ONNX runtime */
+    EXPOSED_RUNTIME_SYMBOL,
+
```
xenova, line 67:
```suggestion
```

Maybe one question - is there a reason this should be exposed via `apis`? It isn't used elsewhere

xenova: nvm - it is!

src/backends/onnx.js
```diff
@@ -59,11 +59,10 @@ const supportedDevices = [];
...
 let defaultDevices;
 let ONNX;
-const ORT_SYMBOL = Symbol.for('onnxruntime');
 
-if (ORT_SYMBOL in globalThis) {
-  // If the JS runtime exposes their own ONNX runtime, use it
-  ONNX = globalThis[ORT_SYMBOL];
+if (apis.IS_EXPOSED_RUNTIME_ENV) {
+    // If the JS runtime exposes their own ONNX runtime, use it
+    ONNX = globalThis[apis.EXPOSED_RUNTIME_SYMBOL];
```
xenova, line 65: Could we maybe add some checks here to ensure that the runtime implements the necessary functions - throwing an error otherwise? May not be necessary, but just in case `onnxruntime` maps to something else in `globalThis` (and not a valid runtime)

kallebysantos: Sure, it makes sense.

kallebysantos: Hey Joshua, I did add it on https://github.com/huggingface/transformers.js/pull/976/commits/f8da67af8c09333dbc59d3c5563a514b57268da6

src/backends/onnx.js
```diff
@@ -54,12 +54,25 @@ const supportedDevices = [];
...
+    }
+
+    if (!Object.hasOwn(onnxruntime, 'InferenceSession')) {
+        throw new Error(`Invalid "globalThis[${String(apis.EXPOSED_RUNTIME_SYMBOL)}]" definition. Missing required exported member "InferenceSession".`)
+    }
+    if(!Object.hasOwn(onnxruntime?.InferenceSession, 'create')) {
+        throw new Error(`Invalid "globalThis[${String(apis.EXPOSED_RUNTIME_SYMBOL)}].InferenceSession" definition. Missing required exported member "InferenceSession.create".`)
+    }
+
+    ONNX = onnxruntime;
```
xenova, line 75: Using an array map would be great here! (minimal code duplication).

xenova (changes requested):
Thanks again for the modifications! I was talking to a colleague about this PR and one thing which was brought up was the possibility of accidental namespace clashes (if `onnxruntime` is detected globally).

For that reason, I think this should be "opt-in" behaviour based on the presence or absence of an environment variable, and should not be the default for all users.

In the supabase runtime, are you able to set default environment variables. e.g., `HF_TRANSFORMERS_USE_EXPOSED_RUNTIME` (name subject to change, ofc).

We can then use this to detect when to use the exposed runtime or not.

kallebysantos:
Hi Joshua, I hope  that you had a happy new year!
> ... based on the presence or absence of an environment variable ... In the supabase runtime, are you able to set default environment variables. e.g., `HF_TRANSFORMERS_USE_EXPOSED_RUNTIME` (name subject to change, ofc).

I agree with this suggestion, the only thing to consider is that `process.env` is only available on `NodeJs` runtime, and since Supabase uses `Deno` I'll need to add an `IS_DENO_ENV` to get it working. 

---

Btw I think that `IS_DENO_ENV` should have already been added cause since `v3` Deno it's not working anymore.
I hope to be wrong, but from my tests I can't get `transformers.js v3` working on Deno without a custom execution provider like we did for supabase. 
I get errors like:
```
Error: Unsupported device: "wasm". Should be one of: .
---
Cannot read properties of undefined (reading 'wasm')
```

src/backends/onnx.js
kallebysantos: Sure! done in [db909e6](https://github.com/huggingface/transformers.js/pull/976/commits/db909e6642120c5699c495aa132d828658d5b4d5)

PR #1005, opened 2024-11-01
Remove duplicate `gemma` value from NO_PER_CHANNEL_REDUCE_RANGE_MODEL set

bekzod:
noticed `gemma` repeated twice in set


xenova:
Thanks!

PR #1014, opened 2024-11-05
fix(docs): fixed a broken link to quantization guide

ThomasWT:
quantization guide link was absolute and pointed to root `/guides/dtype` which would send the user all the way to the root of the url.

Clicking on the link from https://huggingface.co/docs/transformers.js/index took you to https://huggingface.co/guides/dtypes

Now takes you to the right page https://huggingface.co/docs/transformers.js/guides/dtypes


xenova:
Thanks!

PR #1015, opened 2024-11-06
[BUG] When padding an image, the dimensions get stretched

BritishWerewolf:
Fix that padding was stretching the original image, and the right side was overflowing the container.

## Example
This is the code I used to generate the images.  
I resize the images, so that they can be padded again to the same height as the original image.
```javascript
RawImage.fromBlob(file)
.then(image => image.resize((256 / image.height) * image.width, 256))
.then(image => image.pad([128, 128, 128, 128]))
.then(image => image.save('padded.png'));
```

| Original image | Old solution | New solution |
|----------------|--------------|--------------|
| A Moana Loungefly backpack on a white background | A Moana Loungefly backpack, where this is 128 pixels of padding along the top and left, however the image is cropped on the right and bottom | A Moana Loungefly backpack where there is an even 128 pixels of padding all around, and the image is still 256 pixels tall |

As you can see above, the old solution will add the 128 pixel padding to the top and left, however the image is resized and drawn with the same size as the canvas, hence it is cropped by 128 pixels on the right and bottom.

However, the new solution will correctly draw the image with the provided padding all around as we now draw the image with the same dimensions that it was passed in with.

## Explanation
The current implementation was drawing the image with a width of `newWidth`, where `newWidth` will be the entire width.  
This is incorrect, since the width should remain unchanged when drawing, otherwise the image will become stretched.

As a result, we are getting the correct left padding and top padding, but the right and bottom of the image are getting cropped off since the image is being stretched out.  

[From the docs](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage), the `sWidth` and `sHeight` should match `dWidth` and `dHeight`.

### DIFF
--- a/src/utils/image.js
+++ b/src/utils/image.js
@@ -1,10 +1,10 @@
 
 /**
- * @file Helper module for image processing. 
- * 
- * These functions and classes are only used internally, 
+ * @file Helper module for image processing.
+ *
+ * These functions and classes are only used internally,
  * meaning an end-user shouldn't need to access anything here.
- * 
+ *
  * @module utils/image
  */
 
@@ -91,7 +91,7 @@ export class RawImage {
         this.channels = channels;
     }
 
-    /** 
+    /**
      * Returns the size of the image (width, height).
      * @returns {[number, number]} The size of the image (width, height).
      */
@@ -101,9 +101,9 @@ export class RawImage {
 
     /**
      * Helper method for reading an image from a variety of input types.
-     * @param {RawImage|string|URL} input 
+     * @param {RawImage|string|URL} input
      * @returns The image object.
-     * 
+     *
      * **Example:** Read image from a URL.
      * ```javascript
      * let image = await RawImage.read('https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/football-match.jpg');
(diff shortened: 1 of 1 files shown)


xenova:
Hi there, thanks for the PR! 👋 Can you provide some example code/input to see the incorrect behavior observed, and then the proposed corrected behaviour?

BritishWerewolf:
@xenova, I have updated the original PR to demonstrate how the proposed changes fix the previous bug.  
I hope this is clear enough, although if you need further explanation please let me know.  
Thank you.

src/utils/image.js
```diff
@@ -408,13 +408,14 @@ export class RawImage {
             // Draw image to context, padding in the process
             ctx.drawImage(canvas,
                 0, 0, this.width, this.height,
-                left, top, newWidth, newHeight
+                left, top, this.width, this.height
```
BritishWerewolf, line 411: The first row is effectively saying "start a selection at `0,0` and `this.width, this.height`", which gives us a rectangle.

It then draws this selection on line 2, starting at a new `x, y` of `left, top`, and the new dimensions should match what the original image looks like otherwise it gets stretched.

---

You can see why the cropping happens, as shown in my first post, by just looking at width.  
We are stretching the image to `newWidth` and the new canvas is already at a width of `newWidth` hence when we add the padding of `left` to it, we now get a new image that is larger than a canvas - `left + newWidth + right` is greater than `newWidth`!

xenova: Amazing explanation - thanks!

xenova (approved):
Thanks again!

PR #1020, opened 2024-11-10
Add function to apply mask to RawImage.

BritishWerewolf:
I'm working on an app to remove the background from an image.  
This PR adds the ability to very easily apply the result of the model as a mask to a RawImage.  
The image alpha channel will be replaced with the values of the mask.

Here is an example of the sort of thing I am doing.
```javascript
const image = await RawImage.fromURL('https://picsum.photos/600/400');
const padded = image.clone().pad([0, 0, 100, 100]);
const model = await AutoModel.from_pretrained(modelName);
const processed = await processor(padded);
const output = model({ 'input': processed.pixel_values });

// Resize the mask to match the image since many models have pretrained sizes.
// `output` is a square so resize and centre crop to remove padding.
const imageSize = Math.max(image.width, image.height);
const mask = await RawImage.fromTensor(output)
.resize(imageSize, imageSize)
.then(async image => image.center_crop(image.width, image.height));

// Finally, apply the mask and save it.
image.applyMask(mask)
.then(image => image.save('masked_image.png'));
```

### DIFF
--- a/src/utils/image.js
+++ b/src/utils/image.js
@@ -1,10 +1,13 @@
+/**
+ * @typedef {import('./tensor.js').DataArray} DataArray
+ */
 
 /**
- * @file Helper module for image processing. 
- * 
- * These functions and classes are only used internally, 
+ * @file Helper module for image processing.
+ *
+ * These functions and classes are only used internally,
  * meaning an end-user shouldn't need to access anything here.
- * 
+ *
  * @module utils/image
  */
 
@@ -91,7 +94,7 @@ export class RawImage {
         this.channels = channels;
     }
 
-    /** 
+    /**
      * Returns the size of the image (width, height).
      * @returns {[number, number]} The size of the image (width, height).
      */
@@ -101,9 +104,9 @@ export class RawImage {
 
     /**
      * Helper method for reading an image from a variety of input types.
-     * @param {RawImage|string|URL} input 
+     * @param {RawImage|string|URL} input
      * @returns The image object.
-     * 
+     *
      * **Example:** Read image from a URL.
      * ```javascript
      * let image = await RawImage.read('https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/football-match.jpg');
@@ -181,7 +184,7 @@ export class RawImage {
 
     /**
      * Helper method to create a new Image from a tensor
-     * @param {Tensor} tensor 
+     * @param {Tensor} tensor
      */
     static fromTensor(tensor, channel_format = 'CHW') {
         if (tensor.dims.length !== 3) {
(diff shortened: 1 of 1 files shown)


src/utils/image.js
```diff
@@ -304,6 +307,89 @@ export class RawImage {
...
+     */
+    async applyMask(mask) {
+        if (mask.width !== this.width || mask.height !== this.height) {
+            throw new Error('Mask must be the same size as the image');
+        }
+
+        // We want the current image to have an alpha channel, but the mask will
+        // just be a single channel.
+        this.convert(4);
+        mask.convert(1);
```
BritishWerewolf, line 373: Should we throw, or just silently converted the mask to a single channel?

xenova: Probably best to throw an error, since the user may experience incorrect/unexpected results if we do the conversion for them :) (e.g., if they accidentally switch the inputs)

xenova:
Useful PR! 🔥 Could you reference the similar function / usage / inspiration? e.g., how PIL does mask application?

src/utils/image.js
BritishWerewolf: That was my thought as I was reading through this PR again!  
Will make the change now.

BritishWerewolf:
> Useful PR! 🔥 Could you reference the similar function / usage / inspiration? e.g., how PIL does mask application?

@xenova, I am not completely familiar with the Python Imaging Library, however it looks like a similar thing is achieved with the following snippet - assuming I understand the docs correctly 😅

```python
from PIL import Image

background = Image.open('background.png')
foreground = Image.open('foreground.png')
mask = Image.open('mask.png').convert('L')

result = Image.composite(background, foreground, mask)
result.save('masked_image.png')
```

The mask is `convert`ed to greyscale ([docs here](https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes)).

The second image would map to what `RawImage` would represent, and the first image is used to fill in the gaps that are created in accordance with mask ([docs here](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.composite)).  
My implementation instead uses a single image, and will instead use transparent pixels instead of filling them with another image.

This was a great resource for learning how that library and function works:  
https://note.nkmk.me/en/python-pillow-composite/

xenova:
Thanks! I followed that resource you provided and found https://note.nkmk.me/en/python-pillow-putalpha/ - perhaps it's more applicable? If so, I say we rename `applyMask` to `putalpha` :) WDYT?

src/utils/image.js
```diff
@@ -304,6 +307,95 @@ export class RawImage {
...
+    async applyMask(mask) {
+        if (mask.width !== this.width || mask.height !== this.height) {
+            throw new Error('Mask must be the same size as the image');
+        }
+
+        // We want the current image to have an alpha channel, but the mask will
+        // just be a single channel.
+        if (this.channels !== 4) {
+            throw new Error('Image must have 4 channels');
+        }
```
xenova, line 376: Would be useful to support both 3- and 4-channel images as the input image.

src/utils/image.js
```diff
@@ -304,6 +307,95 @@ export class RawImage {
...
+        for (let channel = 0; channel < this.channels; channel++) {
+            const channelIndex = baseIndex + channel;
+            if (channelIndex >= this.data.length) {
+                throw new Error('Index out of bounds');
+            }
+            this.data[channelIndex] = value[channel];
+        }
+
+        return this;
+    }
```
xenova, line 357:
```suggestion
```

Since these functions aren't used in this PR, I suggest we remove them and (if needed) add them to a separate PR. Is there a specific reason for their inclusion?

BritishWerewolf: Good catch!  
Originally I was using both of these functions during the loop, but I removed them as the overhead was not worth it.

BritishWerewolf:
> Thanks! I followed that resource you provided and found https://note.nkmk.me/en/python-pillow-putalpha/ - perhaps it's more applicable? If so, I say we rename `applyMask` to `putalpha` :) WDYT?

Yes - that's a much more apt name!

src/utils/image.js
BritishWerewolf: Would you consider something like this?  
Automatically convert to 4 channels if 3 are provided, however anything else will throw an error.
```suggestion
        if (this.channels === 3) {
            this.convert(4);
        } else if (![3, 4].includes(this.channels)) {
            throw new Error('Image must have 3 or 4 channels');
        }
```

xenova: To prevent unnecessary conversions, I think we can set the step size accordingly (3 for RGB, 4 for RGBA) and write the image data accordingly.

BritishWerewolf: Hey, sorry I’ll get to this soon.
I’m currently poorly and bedridden at the minute, but will work on a fix ASAP.

BritishWerewolf: Hey, I’m still not 100% but I’ve been thinking about this.  
How would a 3 channel RGB image work. For setting the alpha? Wouldn’t we always want that fourth alpha channel?

xenova (approved):
Updated it to support 3-channel and 4-channel images. Will merge now, and will add unit tests in follow-up PR. (Couldn't do now since I can't merge main into the PR; permission error)

PR #1027, opened 2024-11-14
Avoid resizing images when they already have the desired size

nemphys:
This turns the resize method of RawImage to a no-op when the image already has the right size, saving some (or a lot of) CPU cycles.


xenova (approved):
Thanks! LGTM 🚀

PR #1043, opened 2024-11-20
Run prettier command manually once and automatically on CI

pdufour:
I noticed that there is a prettier config, but it doesn't look like the command has been ran in a while, so there are a lot of fixes applied when I run it. Therefore, I did the following:
1) manually ran it once
2) and also added a CI check step that will automatically check PR's with `format:check`

You may get some merge conflicts on PR's when / if this is merged, so for that I recommend the following steps:

**Rebase steps**
- On your **branch**
- `git checkout main -- .prettierrc` (I just made a minor change to max line length to align with existing code)
- `npm run format`
- Then merge in main, that way both files will be formatted and you shouldn't have issues

**Test Plan**
- I made a test commit that broke linting and see that the PR failed - https://github.com/pdufour/transformers.js/actions/runs/11924988068/job/33236295119


pdufour:
Hi, thanks for the quick response! Done - https://github.com/huggingface/transformers.js/pull/1044.

xenova:
Amazing! Thanks so much! 🤗

PR #1057, opened 2024-11-27
Fixes Janus model for WebGPU EP

prathikr:
Fixes the following error reported in the Janus model for WebGPU EP:

image

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -44,10 +44,10 @@ import {
 } from './configs.js';
 
 import {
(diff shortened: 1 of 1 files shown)


xenova:
Hi there 👋 This was intentionally added to avoid needing to recompute image embeddings during generation. The model works on CPU/WASM, so I believe this is an error with the WebGPU EP itself. cc @guschmue

PR #1063, opened 2024-12-02
Fix optional chaining for batch size calculation in PreTrainedModel

emojiiii:
image

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -1757,7 +1757,7 @@ export class PreTrainedModel extends Callable {
             const dtype = session?.config?.kv_cache_dtype ?? 'float32';
             const empty = (dtype === 'float16') ? new Uint16Array() : [];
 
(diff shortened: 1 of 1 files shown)


emojiiii:
This pull request includes a minor change to the `PreTrainedModel` class in the `src/models.js` file. The change improves the robustness of the code by using optional chaining to handle undefined values more gracefully.

* [`src/models.js`](diffhunk://#diff-a19725de73a858e12dc293aef7c29e8d9d107497643f2c65175ab0064d804665L1760-R1760): Modified the `batch_size` calculation to use optional chaining for accessing `dims` property, ensuring the code handles cases where `dims` might be undefined.

xenova:
Thanks for the PR! Can you provide the sample code which produces the error?

emojiiii:
> Thanks for the PR! Can you provide the sample code which produces the error?

I found this when running the [example code ](https://github.com/huggingface/transformers.js/blob/main/examples/text-to-speech-client/src/worker.js#L87)

xenova (approved):
Thanks! I can confirm this fixes the issue.

Issue #394, opened 2023-11-16
[Feature request] streamer callback for text-generation task


seonglae:
**Streamer**
https://huggingface.co/docs/transformers/generation_strategies#streaming

**Reason for request**
Currently, iterating `max_new_tokens: 1` takes much longer time than single generation. Text generation takes time even for light model. Token streaming is key feature for user experience. In my case. Task-specific text generation could be a key feature of AI app development using transformers.js with low cost.

**Additional context**
I'm not sure the `TextStreamer` class need to be compatibility with python transformers. I wrote an use case proposal with `TextStreamer extends TransformStream`. AsyncIterable, AsynsGeneator and Stream API might be usable.  

**Suggesting streaming code**
```ts
let aggregatedResponse = '';
const streamer = TextStreamer()
const pipe: QuestionAnsweringPipeline = await pipeline(
  'text-generation',
  model,
  { quantized: true }
)
pipe(prompt, { streamer: streamer })
res = new Response(streamer.readable)
```
This is vercel's approach 
https://github.com/vercel/ai/blob/main/packages/core/streams/ai-stream.ts
https://github.com/vercel-labs/ai-chatbot/blob/main/app/api/chat/route.ts

xenova:
Hi there 👋 I definitely think the addition of an equivalent `TextStreamer` class to the library will be great! If someone in the community would like to contribute this, it should be as simple as rewriting [this file](https://github.com/huggingface/transformers/blob/fd65aa981868cd1066738658ebd3efdc9c43df55/src/transformers/generation/streamers.py#L38) in JavaScript.

The current approach to text streaming (which was actually added before the python library added `TextStreamer`) is to add a `callback_function` to the generate/pipeline function. For example:
```js
const pipe = await pipeline(
  'text-generation',
  model,
  { quantized: true }
)
pipe(prompt, { callback_function: beams => { console.log(beams) }})
```


Here's an example of streaming + decoding:

https://github.com/xenova/transformers.js/blob/4e4148cb5ce7f4a9265f58b4eeb660c64bed0386/examples/demo-site/src/worker.js#L189-L202

wujohns:
@xenova How to define the callback_function to make the text-generation stop at special words(like the openai api's "stop" param)
I also find the code from transformer.js you shows, but I am confused about how to do this next

02shanks:
@xenova is this issue still open for contribution?

seonglae:
[...]

PR #1066, opened 2024-12-02
Add an example and type enhancement for TextStreamer

seonglae:
Added a GenerationConfig option for TextStreamer and included an example in the documentation as part of #394. I used the example you provided directly and tried to adhere to the repository's style as much as possible, but please let me know if anything deviates from the standard.


docs/source/pipelines.md
```diff
@@ -148,6 +148,76 @@ Cheddar is my go-to for any occasion or mood;
...
+const messages = [
+  { role: "system", content: "You are a helpful assistant." },
+  { role: "user", content:  "Write a quick sort algorithm." },
+];
+
+// Create text streamer
+const streamer = new TextStreamer(generator.tokenizer, {
+  skip_prompt: true,
+  callback_function: (text) => process.stdout.write(text), // Stream to command line
+})
```
xenova, line 175: Unfortunately, `process.stdout` isn't available in browser environments, so I think we should replace this with another function. I know that `console.log` outputs everything on a new line, which doesn't look great.

In fact, how it is defined here is the default when running in Node.js, so perhaps we can do this:


```suggestion
const streamer = new TextStreamer(generator.tokenizer, {
  skip_prompt: true,
  // Optionally, do something with the text (e.g., write to a textbox)
  // callback_function: (text) => { /* Do something with text */ },
})
```

docs/source/pipelines.md
```diff
@@ -148,6 +148,76 @@ Cheddar is my go-to for any occasion or mood;
 It adds depth and richness without being overpowering its taste buds alone
 ```
 
+### Streaming
+
+Some pipelines such as `text-generation` or whisper `speech-to-text` models support streaming output. For example, when using a chat model like `Qwen2.5-Coder-0.5B-Instruct`, you can specify a callback function that will be called with each generated token text.
```
xenova, line 153:
```suggestion
Some pipelines such as `text-generation` or `automatic-speech-recognition` support streaming output. This is achieved using the `TextStreamer` class. For example, when using a chat model like `Qwen2.5-Coder-0.5B-Instruct`, you can specify a callback function that will be called with each generated token text (if unset, new tokens will be printed to the console).
```

docs/source/pipelines.md
```diff
@@ -148,6 +148,76 @@ Cheddar is my go-to for any occasion or mood;
...
+
+- **Pivot Selection**: The pivot is chosen as the middle element of the array.
+
+- **Partitioning**: The array is partitioned into three parts: elements less than the pivot (`left`), elements equal to the pivot (`middle`), and elements greater than the pivot (`right`). These partitions are then recursively sorted.
+
+- **Recursive Sorting**: The subarrays are sorted recursively using `quick_sort`.
+
+This approach ensures that each recursive call reduces the problem size by half until it reaches a base case.
+```
+</pre>
```
xenova, line 216: I would recommend putting this in a `<details>` tag to avoid cluttering the main content.

For example:
```
<details>

<summary>Open here</summary>

Content goes here

</details>
```

Would look like:


<details>

<summary>Open here</summary>

Content goes here

</details>

src/generation/streamers.js
```diff
@@ -34,7 +34,12 @@ const stdout_write = apis.IS_PROCESS_AVAILABLE
 export class TextStreamer extends BaseStreamer {
     /**
      * 
-     * @param {import('../tokenizers.js').PreTrainedTokenizer} tokenizer 
+     * @param {import('../tokenizers.js').PreTrainedTokenizer} tokenizer
+     * @param {Object} options
+     * @param {boolean} [options.skip_prompt=false] Whether to skip the prompt tokens
+     * @param {function(string): void} [options.callback_function=null] Function to call when a piece of text is ready to display
+     * @param {function(bigint[]): void} [options.token_callback_function=null] Function to call when a new token is generated
+     * @param {Object} [options.decode_kwargs={}] Additional keyword arguments to pass to the tokenizer's decode method
```
xenova, line 42: Very useful! Thanks!

xenova (changes requested):
Thanks! Just some suggestions.

docs/source/pipelines.md
seonglae: I've made the fix. Could you check?

xenova (approved):
LGTM ✅ Will run tests + build docs & merge soon.

seonglae:
If you don’t mind, I can contribute a `chat-streaming` example to the `/example` folder. Would it be okay to use Vite with Tailwind styling like others (or possibly Shadcn and Next.js), or is there a standard dependency stack for this repository examples?

xenova:
That would be much appreciated! However, could you instead contribute it to https://github.com/huggingface/transformers.js-examples? We're in the process of moving all examples to that repository.

In terms of stack, feel free to use what you like, but it should contribute to the example you are trying to showcase. For example, for a simple streaming example, I would recommend keeping it as simple as possible (just to highlight the important part; i.e., streaming). That said, it could be useful to see how to pass this information back from a web worker too, so feel free to contribute two examples if you'd like the second one to be more "fully-featured". In that case, feel free to use what you feel most comfortable with.

One other thing to note is that Next.js introduces another level of complexity since it can be run as a non-static site. Anyway, things to consider - I look forward to the contribution! Thanks!

Issue #1078, opened 2024-12-07
An error on musicgen-web example


porry420:
### System Info

OS: Windows 10
node version 20.16.0
npm version10.8,1

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension
- [ ] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

I tired to run this example.
And I have got an error like this.

 This error came from the "onResolve" callback registered here:

    node_modules/esbuild/lib/main.js:1150:20:
      1150 │       let promise = setup({

### Reproduction

clone full repository.
go to sub direcoty
npm install
npm run dev

SeanGallen:
@porry420, I was able to reproduce the bug.  I have a PR with a potential solution here, if you would like to try. I was able to get it working with those changes.

porry420:
I am going to change this Musicgen-web so that it can run without the internet. I have followed the custom usage instructions, but it is not working correctly. Can you give me more details?

SeanGallen:
Here are the steps I used to get it working with it connected to the internet:

In the package.json change the `"@xenova/transformers": "github:xenova/transformers.js#v3"` to `"@huggingface/transformers": "^3.1.2",`.
In the `examples/musicgen-web/src/App.jsx`, change `"@xenova/transformers` to `@huggingface/transformers`

Just to be safe, remove the `"@xenova/transformers` from your node_modules. `npm uninstall @xenova/transformers`.
Run `npm install` in the root of the `musicgen-web`, to re-install all the packages.

This should get it working locally for you.

PR #1080, opened 2024-12-07
Fix musicgen build issue

SeanGallen:
After looking into [this bug with Musicgen-Web](https://github.com/huggingface/transformers.js/issues/1078), I was able to reproduce it.  I notice the Transformesr.js dependence package was pointed to its previous location.  I updated it to its new location `"huggingface/transformers": "^3.1.2",` and retested. It worked.  I also remove the previous `xenova/transformers` library from my node_modules and any reference to it in the app.

I notice other examples also use the `huggingface/transformers` so I figured it was fine to make a PR and see if it will help others facing this issue.

Screenshot 2024-12-07 at 3 37 38 PM

### DIFF
--- a/examples/musicgen-web/package-lock.json
+++ b/examples/musicgen-web/package-lock.json
@@ -8,7 +8,7 @@
       "name": "musicgen-web",
       "version": "0.0.0",
       "dependencies": {
-        "@huggingface/transformers": "^3.1.1",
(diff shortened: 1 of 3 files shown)


SeanGallen:
Hi @xenova, could you please review these changes?  If you're happy with them, I noticed a few other places with similar issues. I'd be happy to apply the same corrections to those as well, if you'd like."

PR #1081, opened 2024-12-07
Check TypeScript errors during build

ocavue:
`npm run typegen` (and `npm run build`) now will check TypeScript errors.

I migrated `jsconfig.json` to `tsconfig.json` because `tsc --build` can only recognize `tsconfig.json`. `tsc --build` is faster than `tsc` because `tsc --build` has cache, so it's better to use `tsc --build`. 

---

Since it reveals all typing errors now, I need to address all of them to make the build pass.

- Most of them are just missing types in configs, like the one below: 

  ```ts
  this.config.image_token_index;
              ~~~~~~~~~~~~~~~~~
              ^ Property 'image_token_index' does not exist on type 'PretrainedConfig'.
  ```

  For those errors, I simply add `@ts-expect-error` comment to let TypeScript ignore it. I use `@ts-expect-error` instead of `@ts-ignore` because we can remove those comments once we add missing types in the future.

- I fixed some other types issues that I can fix. 
- One thing I noticed is that [MgpstrProcessor.batch_decode](https://github.com/huggingface/transformers.js/blob/9914e7a4365523d6e6d3a807dd2a38d6c8d95ee7/src/models/mgp_str/processing_mgp_str.js#L122) has different shape compared to the method in the parent class. This might not be a good idea.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "scripts": {
     "format": "prettier --write .",
     "format:check": "prettier --check .",
-    "typegen": "tsc ./src/transformers.js --allowJs --declaration --emitDeclarationOnly --declarationMap --outDir types",
+    "typegen": "tsc --build",
     "dev": "webpack serve --no-client-overlay",
     "build": "webpack && npm run typegen",
     "test": "node --experimental-vm-modules node_modules/jest/bin/jest.js --verbose",
--- a/src/base/image_processors_utils.js
+++ b/src/base/image_processors_utils.js
@@ -985,6 +985,7 @@ export class ImageProcessor extends Callable {
         // NOTE: All pixel-level manipulation (i.e., modifying `pixelData`)
         // occurs with data in the hwc format (height, width, channels), 
         // to emulate the behavior of the original Python code (w/ numpy).
+        /** @type {Float32Array} */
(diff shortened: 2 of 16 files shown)


xenova:
Thanks for this! I'll do a full review soon, but just to address your third point: https://github.com/huggingface/transformers/blob/c8c8dffbe45ebef0a8dba4a51024e5e5e498596b/src/transformers/models/mgp_str/processing_mgp_str.py#L102 has a similar issue where `MgpstrProcessor` has a non-standard `batch_decode` function. This is because the model returns three tensors that all need to be used in the decode method. Not too sure how to make it fit the parent class type though.

ocavue:
I would consider this as an issue in the original Python transformers lib. At least its [docstring](https://github.com/huggingface/transformers/blob/c8c8dffbe45ebef0a8dba4a51024e5e5e498596b/src/transformers/models/mgp_str/processing_mgp_str.py#L104-L105) is wrong. 

I don't know any context about `MgpstrProcessor`, but maybe renaming this method to something else (e.g. just `decode()`) could improve it.

xenova:
Apologies for the delay - reviewing now 👍

xenova (approved):
Great PR! Thanks again 😇

PR #1083, opened 2024-12-08
Refactor per-model unit testing

xenova:
- Move per-model unit tests to separate folders
- Fix issue when overriding model file name


xenova:
Will do follow-up PR for auto-detection of tokenizers and models.

Issue #302, opened 2023-09-14
[Feature request] Whisper Language Detection


FelippeChemello:
Name of the feature
Language Detection with the Whisper Model in Transformers.js

Reason for request
The original Whisper model includes a dedicated function for language detection. It would be awesome to have a similar capability within Transformers.js. In my current application, Whisper's models are running efficiently on the browser. However, for language detection, I find myself repeatedly making requests to my backend service.

Additional context
Is it possible to implement language detection as an event returned from the pipeline?

xenova:
Hi there! 👋 Definitely a possibility I'd say! I assume the original (openai) library analyses attention scores across the different language tokens and picks the most likely one. Do you perhaps have example code for how to achieve this with the python `transformers` library?

FelippeChemello:
Hi,
I found this code in a [thread discussion](https://discuss.huggingface.co/t/language-detection-with-whisper/26003) on Hugging Face. It's not part of the default implementation of the `transformers` library, but perhaps it could be added to the library to make it easier to obtain this information.

```python
def detect_language(model: WhisperForConditionalGeneration, tokenizer: WhisperTokenizer, input_features,
                    possible_languages: Optional[Collection[str]] = None) -> List[Dict[str, float]]:
    # hacky, but all language tokens and only language tokens are 6 characters long
    language_tokens = [t for t in tokenizer.additional_special_tokens if len(t) == 6]
    if possible_languages is not None:
        language_tokens = [t for t in language_tokens if t[2:-2] in possible_languages]
        if len(language_tokens) < len(possible_languages):
            raise RuntimeError(f'Some languages in {possible_languages} did not have associated language tokens')

    language_token_ids = tokenizer.convert_tokens_to_ids(language_tokens)

    # 50258 is the token for transcribing
    logits = model(input_features,
                   decoder_input_ids = torch.tensor([[50258] for _ in range(input_features.shape[0])])).logits
    mask = torch.ones(logits.shape[-1], dtype=torch.bool)
    mask[language_token_ids] = False
    logits[:, :, mask] = -float('inf')

    output_probs = logits.softmax(dim=-1).cpu()
    return [
        {
            lang: output_probs[input_idx, 0, token_id].item()
            for token_id, lang in zip(language_token_ids, language_tokens)
        }
        for input_idx in range(logits.shape[0])
    ]
```

xenova:
[...]

PR #1097, opened 2024-12-13
Add Whisper language detection

ae9is:
See #302 

Adds support for automatically detecting language to Whisper tasks.

The existing HuggingFace and Whisper implementations in Python were used as reference:
[Hugging Face Transformers](https://github.com/huggingface/transformers/blob/98e8062df3cf82b367e8a243963e4afb0d9d3407/src/transformers/models/whisper/generation_whisper.py#L1425)
[Original Whisper](https://github.com/openai/whisper/blob/90db0de1896c23cbfaf0c58bc2d30665f709f170/whisper/decoding.py#L19)

Also updates the existing Whisper test suites, including adding a string similarity check for actual model output (as opposed to just output length). Please note that the "new" development dependency for these tests, "fastest-levenshtein" is already used by "webpack-cli".

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -90,6 +90,7 @@ import {
     TopKLogitsWarper,
     TopPLogitsWarper,
     ClassifierFreeGuidanceLogitsProcessor,
+    OnlyGoodWordsLogitsProcessor,
 } from './generation/logits_process.js';
 
 import {
@@ -112,7 +113,7 @@ import {
 import { RawImage } from './utils/image.js';
 
 import { dynamic_time_warping, max, medianFilter } from './utils/maths.js';
-import { EosTokenCriteria, MaxLengthCriteria, StoppingCriteriaList } from './generation/stopping_criteria.js';
+import { AlwaysStopCriteria, EosTokenCriteria, MaxLengthCriteria, StoppingCriteriaList } from './generation/stopping_criteria.js';
 import { LogitsSampler } from './generation/logits_sampler.js';
 import { apis } from './env.js';
 
@@ -1210,6 +1211,10 @@ export class PreTrainedModel extends Callable {
 
         if (generation_config.bad_words_ids !== null) {
             processors.push(new NoBadWordsLogitsProcessor(generation_config.bad_words_ids, generation_config.eos_token_id));
+        }
+
+        if (generation_config.good_words_ids !== null) {
+            processors.push(new OnlyGoodWordsLogitsProcessor(generation_config.good_words_ids, generation_config.eos_token_id));
         }
 
         if (generation_config.min_length !== null && generation_config.eos_token_id !== null && generation_config.min_length > 0) {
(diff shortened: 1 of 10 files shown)


xenova:
Thanks for the PR! This will certainly be a useful feature. Regarding the implementation, I think it can be greatly simplified as follows:
- Instead of using `.generate`, perform a single forward pass of the inputs
- Then, consider all logits which correspond to the language token ids
- Choose the language with the highest score

Currently, the implementation seems to perform a full generation step.

ZhangPeng4242:
hey there, please approve this feature, it is a quite useful feature :)

src/models.js
```diff
@@ -3119,10 +3124,48 @@ export class WhisperForConditionalGeneration extends WhisperPreTrainedModel {
...
+            ...options,
+            generation_config: {
+                ...generation_config,
+                good_words_ids,
+                num_beams: 1,
+                do_sample: false,
+            },
+            stopping_criteria,
+            decoder_input_ids,
+        });
```
xenova, line 3158: We should be able to replace this with a single forward pass (by called `this.forward(...)` instead of using a generation step.

src/models.js
```diff
@@ -3119,10 +3124,48 @@ export class WhisperForConditionalGeneration extends WhisperPreTrainedModel {
...
 
     /**
+     * Detects language by running input through the model and checking for language tokens in the output.
      * 
-     * @param {WhisperGenerationConfig} generation_config 
+     * @param {import('./models/whisper/generation_whisper.js').WhisperGenerationFunctionParameters} options
+     * @returns {Promise<number[]>} A list of language token IDs detected.
+     */
+    async _detect_language(options) {
+        const inputs = options.inputs
```
edbrdi, line 3133: When testing this PR "inputs" was in my case not present, instead I had "input_features".

I noticed the type returns: "(Tensor of varying shape depending on the modality, optional): The sequence used as a prompt for the generation or as model inputs to the encoder. If null the method initializes it with bos_token_id and a batch size of 1. For decoder-only models inputs should be in the format of input_ids. For encoder-decoder models inputs can represent any of input_ids, input_values, input_features, or pixel_values."

By the way thanks for adding language detection, hope it will be merged soon :)

edbrdi: I think it depends on the model used, you probably can reproduce it with https://huggingface.co/onnx-community/whisper-large-v3-turbo - I was able to fix it with `const inputs = options.inputs ?? options.input_features;` in _detect_language on my side.

vklyukin:
Hello! I wanted to check in on this PR - would it be possible to get this approved and merged soon as it would significantly improve user experience with automatic language detection?

ZhangPeng4242:
Any updates on this merge? thank you.

ae9is:
In case it helps anyone else out there: https://github.com/ae9is/transformers.js/tree/develop-3.7.x

Replayed my commits on main, plus some minor fixes and changes for my use case.

ae9is:
For anyone still following this issue, a month ago I updated my fork including Whisper language autodetection for version 4 here: https://github.com/ae9is/transformers.js/tree/develop-4.0.0. See particularly [#3e2adc](https://github.com/ae9is/transformers.js/commit/3e2adc52dc700d7afb820c7dcde29031fe78a37e)

If the maintainers are still interested, I can update this PR. Perhaps best to wait until after #1594 

I've also noticed that since my changes another PR has popped up for v4 (haven't reviewed this at all): #1541

ae9is:
Going to close this one out of respect for everyone's time.

Offer to update for v4 still stands if the maintainers are interested, just reach out!

PR #1098, opened 2024-12-13
fix: update WebGPU embedding benchmark to prefer `GPUAdapter.info` over `requestAdapterInfo`

ErichDonGubler:
Migrates to use the new spec.-compliant way of getting adapter info in WebGPU (see also <https://github.com/microsoft/onnxruntime/pull/21065>).

This has been removed by both Chrome and Firefox, so there might be enough reason to simply stop using `requestAdapterInfo`, but I wasn't sure.

### DIFF
--- a/examples/webgpu-embedding-benchmark/main.js
+++ b/examples/webgpu-embedding-benchmark/main.js
@@ -107,7 +107,7 @@ let gpuHasFp16 = false;
 try {
   // Shouldn't fail since the WebGPU model has loaded successfully
   const adapter = await navigator.gpu.requestAdapter();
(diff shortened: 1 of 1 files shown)


ErichDonGubler:
This might not be ready to merge yet, if the main library doesn't yet expose the `info` member. I wasn't sure about that, either. 😅

xenova:
Thanks for the heads up! I'm busy migrating all our examples to http://github.com/huggingface/transformers.js-examples, and will update them to use `adapter.info` when necessary!

xenova:
(closed recently by https://github.com/huggingface/transformers.js/pull/1382)

PR #1103, opened 2024-12-19
[WIP] Add Gaussian Blur to RawImage

BritishWerewolf:
A very slow, first draft of a Gaussian Blur for RawImage.  
The reason for this, is for models like U2Net where the input image size is 320x320, meaning that the resulting output needs to be resized, and can have sharp edges; blurring before resizing can smooth this out.

This is the code that I used to generate the images.
```js
const kernels = [5, 9, 17];
const sigmas = [2, 3, 5];
for (let i = 0; i < kernels.length; i++) {
    const kernel = kernels[i];
    const sigma = sigmas[i];
    image
        .gaussianBlur(kernel, sigma)
        then(image => image.save(`blurred-(${kernel}-${sigma}).png`));
}
```
Sigma should roughly be a third of kernel size (from what I found researching).  
Kernel must be odd, since we want an even square around the pixel.  
For example, if we had a 3x3 kernel grid, then when we place this on our pixel there is an even space all around that pixel (in this case 1 pixel).

Here is a demo of what this implementation currently looks like.

| Image | Kernel | Sigma |       Duration       |
|-------|--------|-------|----------------------|
| The original image. |  N/A   |  N/A  |         N/A          |
| The image slightly more blurred than the original. |    5   |   2   |            9 seconds |
| The slightly more blurred image, slightly more blurred again. |    9   |   3   |           31 seconds |
| The final image that is blurred even more! |   17   |   5   | 1 minute, 42 seconds |

As you can see, this is _very_ slow, but I will continue to look into how to improve this significantly!

### DIFF
--- a/src/utils/image.js
+++ b/src/utils/image.js
@@ -665,30 +665,19 @@ export class RawImage {
 
         const kernel = [];
         const center = Math.floor(kernelSize / 2);
+        const sigma2 = sigma * sigma;
         let sum = 0;
 
         for (let i = 0; i < kernelSize; i++) {
-            kernel[i] = [];
-            for (let j = 0; j < kernelSize; j++) {
-                const x = i - center;
-                const y = j - center;
-
-                // Square the numbers.
-                const x2 = x * x;
-                const y2 = y * y;
-                const sigma2 = sigma * sigma;
-
-                const value = Math.exp(-(x2 + y2) / (2 * sigma2));
-                kernel[i][j] = value;
-                sum += value;
-            }
+            const x = i - center;
+            const value = Math.exp(-(x * x) / (2 * sigma2));
+            kernel[i] = value;
+            sum += value;
         }
 
-        // Normalise the kernel.
+        // Normalize the kernel
         for (let i = 0; i < kernelSize; i++) {
-            for (let j = 0; j < kernelSize; j++) {
-                kernel[i][j] /= sum;
-            }
+            kernel[i] /= sum;
         }
 
         return kernel;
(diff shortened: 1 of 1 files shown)


src/utils/image.js
```diff
@@ -651,6 +651,100 @@ export class RawImage {
...
+     * Generates a 1D Gaussian kernel.
+     * @param {number} kernelSize - Kernel size (must be odd).
+     * @param {number} sigma - Standard deviation of the Gaussian.
+     * @returns {Array<number>} - The 1D Gaussian kernel.
+     */
+    generateGaussianKernel(kernelSize, sigma) {
+        // Kernel must be odd because each pixel must sit evenly in the middle.
+        if (kernelSize % 2 === 0) {
+            throw new Error('Kernel size must be odd.');
+        }
```
BritishWerewolf, line 664: The reason for this, is because a kernel is essentially a square that is applied to every pixel in an image.

Consider a grid of 5x5, it would look like this, with our pixel (`P`) in the middle.
```
. . . . .
. . . . .
. . P . .
. . . . .
. . . . .
```
Here, it is clear to see that there is an even spacing around the pixel.

Now, consider what would happen if we used an even kernel size of `4`.
```
. . . .
. . . .
. . P .
. . . .
```
The pixel does not sit neatly in the middle of the kernel and this is a problem.

src/utils/image.js
```diff
@@ -651,6 +651,126 @@ export class RawImage {
...
+        const channels = this.channels;
+
+        // Rather than checking an entire grid of elements, we can instead do
+        // two separate passes with a 1d array (rather than 2d).
+        // Consider a 3x3 kernel, instead of calculating each pixel 9 times, we
+        // can instead calculate two sets of 3 values and then combine them.
+        const horizontalPass = new Float32Array(this.data.length);
+        const verticalPass = new Uint8ClampedArray(this.data.length);
+
+        const numChunks = navigator.hardwareConcurrency || 4;
```
BritishWerewolf, line 707: This was originally a hard coded value, passed into the function, however this approach makes more sense.  
Looking at the table in the opening post, it shows that (at least for my machine) there is no benefit to increase the chunks above 4.

xenova:
Similarly to https://github.com/huggingface/transformers.js/pull/1101#issuecomment-4079370914, any requirement for Gaussian blurring should be performed inside the optimized model, allowing transformers.js to utilize ONNX ops for the operation.

PR #1116, opened 2024-12-25
Remove unused imports from tokenizers.js

pratapvardhan:
Use of these imports was removed in 77ebe0d, this PR removes the unused imports. cc @xenova


xenova (approved):
Thanks!

PR #1128, opened 2024-12-31
Update React guide for v3

sroussey:
Revise the React guide to incorporate changes related to the new streaming capabilities.

### DIFF
--- a/docs/source/tutorials/react.md
+++ b/docs/source/tutorials/react.md
@@ -423,17 +423,14 @@ self.addEventListener('message', async (event) => {
       self.postMessage(x);
   });
 
-  let allText = "";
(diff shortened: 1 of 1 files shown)


xenova (commented):
Thanks for this! I missed this when updating the docs for the Transformers.js v3 release. One thing which would probably be better is for the callback function to only post back the new text (and then the new text should be appended to the existing text, on the react side of things).

sroussey:
True, particularly for an example. 

Old habit of production code where fixing bugs with the smallest number of changes.

xenova (approved):

xenova:
Thanks again @sroussey!

PR #1129, opened 2024-12-31
Update demo-site dependencies and installation instructions

sroussey:
Upgrade the demo-site to use the latest version of Vite and replace the transformer library with Hugging Face's version, updating installation instructions accordingly. Uses streamer instead of bare callback.

### DIFF
--- a/examples/demo-site/package-lock.json
+++ b/examples/demo-site/package-lock.json
@@ -1,0 +1,2050 @@
+{
+  "name": "demo-site",
+  "version": "0.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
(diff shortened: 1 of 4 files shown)


sroussey:
The text generation example still does not work since it is creating a url where the file name starts with model, but in the case of `Xenova/distilgpt2` the file starts with decoder_model

sroussey:
If your pipeline tests were not ignored, I think you would have noticed that the summarize test would fail.

pselvana:
@xenova - Any chance this can be prioritized? The existing @xenova/transformers packages are deprecated and anyone using the demo-site as a starter will find themselves unsupported for newer models.

PR #1133, opened 2025-01-02
Add the ONNX TensorRT execution provider and respective device.

nemphys:
ONNX has a TensorRT EP that is available for NodeJS along with the CUDA EP (under linux x64) when installing onnxruntime-node.
This commit makes it possible to use it by specifying 'tensorrt' as the device.

### DIFF
--- a/src/backends/onnx.js
+++ b/src/backends/onnx.js
@@ -38,6 +38,7 @@ const DEVICE_TO_EXECUTION_PROVIDER_MAPPING = Object.freeze({
     webgpu: 'webgpu', // WebGPU
     cuda: 'cuda', // CUDA
     dml: 'dml', // DirectML
+    tensorrt: 'tensorrt', // TensorRT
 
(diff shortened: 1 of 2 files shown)


xenova:
Hi there! Thanks for the PR. Is there a way to detect when this execution provider is available on a device? Is it only available on linux x64? It doesn't seem to be listed in the [README](https://www.npmjs.com/package/onnxruntime-node):
image

nemphys:
Hello, as mentioned above, the tensorrt EP is available only under linux and seems to be included with the cuda EP in the default package distribution.
Indeed, they do not mention it anywhere, but it is there from the first onnxruntime distribution that introduced cuda support.

xenova:
Maybe @guschmue can share some more information on why it's not listed 👀 

Also, @nemphys if you could provide a link to onnxruntime-node example usage/docs which shows this being used, that would be helpful 👍

PR #1136, opened 2025-01-04
[FEAT] Add dilate and erode operations to Tensor

BritishWerewolf:
This PR is inspired by the functionality of [OpenCV](https://github.com/egonSchiele/OpenCV/blob/master/modules/imgproc/src/morph.cpp#L981).

With this, we can create a kernel that is operated over each element of the `data` array and performs some calculation.  
For this PR, we only have `dilate` and `erode`.

I decided to set these functions to `async`, despite that making the chaining a bit clunkier, since these functions are often used together.

---

# Explanation
### `kernel`
This is an array of 0s and 1s that will be applied to each element of the `data` array, 1 by 1.

There are multiple shapes supported in this PR, they are: `RECT`, `CROSS`, `ELLIPSE`.  
Consider a 5x5 kernel for each shape:
* `RECT`:
    ```
    1 1 1 1 1
    1 1 1 1 1
    1 1 1 1 1
    1 1 1 1 1
    1 1 1 1 1
    ```
* `CROSS`:
    ```
    0 0 1 0 0
    0 0 1 0 0
    1 1 1 1 1
    0 0 1 0 0 
    0 0 1 0 0 
    ```
* `ELLIPSE`:
    ```
    0 0 1 0 0
    0 1 1 1 0
    1 1 1 1 1
    0 1 1 1 0 
    0 0 1 0 0 
    ```
The element we are checking will be in the centre, and then it will apply this kernel to all those values around it.  
Any value of 1 will be included and 0 will be discarded.

### `dilate`
Once we have a list of values for that element, we will apply a `Math.max` to return the biggest value of these.

Consider this Tensor.
```
 1  2  3  4  5
 6  7  8  9 10
11 12 13 14 15
16 17 18 19 20
21 22 23 24 25
```
If we were to apply `CROSS` with a kernel size of 3x3, then the first iteration would return the values of `[1, 2, 6]`. This is because the top and left side of the kernel are off the edge.

When the loop gets to the value of `13`, the extracted values would be `[8, 12, 13, 14, 18]`.

Finally, the value in each instance will be replaced with the maximum value.  
Again, in the case of the first element, that would be `6` and for the centre element, that would be replaced with `18`.

### `erode`
This works in the exact same way as `dilate`, however it will apply the `Math.min` to the elements instead.

### DIFF
--- a/src/utils/tensor.js
+++ b/src/utils/tensor.js
@@ -39,8 +39,13 @@ const DataTypeMap = Object.freeze({
 });
 
 /**
- * @typedef {keyof typeof DataTypeMap} DataType
- * @typedef {import('./maths.js').AnyTypedArray | any[]} DataArray
+ * @typedef {keyof typeof DataTypeMap} DataType  A Tensor data type, for example `uint8`.
+ * @typedef {import('./maths.js').AnyTypedArray | any[]} DataArray  A typed array or an array of values.
+ * @typedef {{x: number, y: number}} Point  An object representing coordinates.
+ * @typedef {'RECT' | 'CROSS' | 'ELLIPSE'} Shape  A shape for morphological operations.
+ * @typedef {{width: number, height: number}} Size  An object representing the size of an object.
+ *
+ * @typedef {number | [number, number] | Size} KernelSize  A kernel size for morphological operations.
  */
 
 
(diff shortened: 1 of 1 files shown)


src/utils/tensor.js
```diff
@@ -789,6 +794,189 @@ export class Tensor {
...
+                        if (neighborRowIndex >= 0 && neighborRowIndex < rows && neighborColIndex >= 0 && neighborColIndex < cols) {
+                            const neighborIndex = (batchIndex * rows * cols) + neighborRowIndex * cols + neighborColIndex;
+                            // Only include values where the kernel has a value
+                            // of 1.
+                            // Rather than multiply against this value, we use
+                            // the if check to reduce the size of the array.
+                            const kernelValue = kernel[kernelRowOffset + paddingSize.height][kernelColOffset + paddingSize.width];
+                            if (kernelValue === 1) {
+                                kernelValues.push(data[neighborIndex] * kernelValue);
+                            }
```
BritishWerewolf, line 909: Originally I was not using the if statement, and utilising the fact that a kernel contains either `0` or `1`, however that meant that `kernelValues` array would be bigger than it needs to be.

Also, `erode` wouldn't work properly because `0` is always the minimum value.

src/utils/tensor.js
```diff
@@ -1542,3 +1730,141 @@ export function quantize_embeddings(tensor, precision) {
...
+    } else {
+        throw new Error("Invalid kernel size.");
+    }
+
+    if (kernel.width % 2 === 0 || kernel.height % 2 === 0) {
+        throw new Error("Kernel size must be odd");
+    }
+
+    return kernel;
+}
```
BritishWerewolf, line 1789: The purpose of this function is to help users with passing in kernel values.

To get a 3x3 kernel to the `erode` function we can do this:
* `tensor.erode(3)`
* `tensor.erode([3, 3])`
* `tensor.erode({ width: 3, height: 3})`

Most likely we want a symmetrical kernel, and it's easier to just pass in a single number.

xenova:
Similarly to https://github.com/huggingface/transformers.js/pull/1101#issuecomment-4079370914, any requirement for dilation or erosion should be performed inside the optimized model instead of on `Tensor`, allowing transformers.js to utilize ONNX ops for the operation.

PR #1139, opened 2025-01-09
Add option to skip special tokens in TextStreamer

sroussey:
Introduce a new option to the TextStreamer class that allows skipping special tokens during decoding, aligning its functionality with WhisperTextStreamer.

### DIFF
--- a/src/generation/streamers.js
+++ b/src/generation/streamers.js
@@ -55,7 +55,7 @@ export class TextStreamer extends BaseStreamer {
         this.skip_prompt = skip_prompt;
         this.callback_function = callback_function ?? stdout_write;
         this.token_callback_function = token_callback_function;
-        this.decode_kwargs = { ...decode_kwargs, ...kwargs , skip_special_tokens };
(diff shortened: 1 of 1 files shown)


src/generation/streamers.js
```diff
@@ -53,7 +55,7 @@ export class TextStreamer extends BaseStreamer {
         this.skip_prompt = skip_prompt;
         this.callback_function = callback_function ?? stdout_write;
         this.token_callback_function = token_callback_function;
-        this.decode_kwargs = { ...decode_kwargs, ...kwargs };
+        this.decode_kwargs = { ...decode_kwargs, ...kwargs , skip_special_tokens };
```
xenova, line 58:
```suggestion
        this.decode_kwargs = { skip_special_tokens, ...decode_kwargs, ...kwargs };
```

xenova (approved):
Thanks! Technically this functionality was already present since `skip_special_tokens` would be passed inside `kwargs`, but it is much better to be documented. Much appreciated!

PR #1141, opened 2025-01-12
add test for input_boxes

sabman:
The model should in theory support `input_boxes` as per this issue: https://huggingface.co/Xenova/slimsam-77-uniform/discussions/2#65c022977c6ad1994b893a8d but there's not test or example to verify this. I also couldn't get the model to accept the `input_boxes` on the main branch. Getting error:

```
Error: An error occurred during model execution: "Missing the following inputs: input_points, input_labels.
```

### DIFF
--- a/tests/models/sam/test_modeling_sam.js
+++ b/tests/models/sam/test_modeling_sam.js
@@ -41,6 +41,28 @@ export default () => {
       MAX_TEST_EXECUTION_TIME,
     );
 
+    it(
+      "w/ input_boxes",
+      async () => {
+        const raw_image = await load_cached_image("corgi");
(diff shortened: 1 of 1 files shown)


mhassanch:
I had a look at optimum package and it seems that **input_boxes** is missing as input in SamOnnxConfig(https://github.com/huggingface/optimum/blob/adcae3841ac638aa911e6a185d3f1617e58495f8/optimum/exporters/onnx/model_configs.py#L2408). 

Also there is no test available with **input_boxes** as a input in transformer.js (https://github.com/huggingface/transformers.js/blob/1130961ac26766ba4a0f5af2640925f4f39b9974/tests/models/sam/test_modeling_sam.js)

PR #1145, opened 2025-01-14
Typescript

tsekiguchi:
Hi @xenova and team,

First off, I want to say how much I’ve been enjoying Transformers.js! I’ve been using it for a couple of months now, and it’s truly an incredible library. Thank you for all the work you’ve put into it!

That said, I’ve encountered some challenges related to types, auto-complete, and error handling while working with the library. To help address these and improve the overall developer experience, I’d love to embark on a bigger project: rewriting Transformers.js in TypeScript.

The benefits of this rewrite would be substantial—it would make the library even more robust while greatly enhancing DX. It also presents a great opportunity to revisit some core code, tackle any outstanding TODOs, and adopt modern JavaScript patterns, such as using typed objects instead of maps/dicts where appropriate.

To get things rolling, I’ve started with a few files in the utils directory, but I’d like to submit my work on RawImage `image.ts` first to ensure the approach aligns with everyone’s expectations before proceeding further.

Additionally, I’ve implemented consistent error handling using try/catch. Errors will still throw as before but will now pass through console.error to log them first. This addresses a gap where some existing error statements weren’t getting logged, leading to crashes without feedback.

Importantly, the goal of this TypeScript rewrite is **not** to break anything. We’re maintaining the existing methods and inputs unless absolutely necessary, so nothing currently in production should be affected.

Thanks, and looking forward to making this library even better!

### DIFF
--- a/src/utils/image.ts
+++ b/src/utils/image.ts
@@ -1,0 +1,1254 @@
+/**
+ * @file Helper module for image processing.
+ *
+ * These functions and classes are only used internally,
+ * meaning an end-user shouldn't need to access anything here.
+ *
+ * @module utils/image
+ */
+
+import { getFile } from "./hub.js";
+import { env, apis } from "../env.js";
+import { Tensor } from "./tensor.js";
+
+// Will be empty (or not used) if running in browser or web-worker
+import sharp from "sharp";
+
+type CreateCanvasFunction = (
+  width?: number,
+  height?: number,
+) => OffscreenCanvas;
+type LoadImageFunction = (input: sharp.Sharp) => Promise<RawImage>;
+type LoadBrowserImageFunction = (
+  input: ImageBitmapSource,
+) => Promise<ImageBitmap>;
+
+let createCanvasFunction: CreateCanvasFunction | undefined;
+
+const IS_BROWSER_OR_WEBWORKER = apis.IS_BROWSER_ENV || apis.IS_WEBWORKER_ENV;
+
+// Function to Get Browser Image Loader
+function getBrowserImageLoader(): {
+  loadBrowserImageFunction: LoadBrowserImageFunction;
+  createCanvasFunction: CreateCanvasFunction;
+} {
+  try {
+    if (!apis.IS_BROWSER_ENV && !apis.IS_WEBWORKER_ENV) {
+      throw new Error(
+        "This function can only be used in a browser or web worker environment.",
+      );
+    }
+
+    const createCanvasFunction: CreateCanvasFunction = (
+      width?: number,
+      height?: number,
+    ) => {
+      if (!self.OffscreenCanvas) {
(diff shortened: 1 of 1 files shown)


BritishWerewolf:
As stated above, there is no reason to rewrite this in TypeScript, because this library already has type safety through JSDoc.  
[JSDoc is supported Out the Box with VS Code](https://code.visualstudio.com/docs/languages/javascript#_jsdoc-support) - and I imagine that other IDEs will have support too.

tsekiguchi:
Okayyyyy. I hear you. 😮‍💨

The goal here isn’t to debate JSDoc versus TypeScript. At the end of the day, we can all agree that types are really helpful for improving the **developer experience** – ensuring that developers who are implementing this library know what information to input, and what to expect when it comes out. For many of us, TypeScript is simply one tool that we use to accomplish this. If specific decisions have already been made regarding TS, then let’s visibly document them so that others can understand the reasoning and we can avoid revisiting.

Don't get me wrong - there's a ton of great JSDoc annotations already, and I appreciate that! But there’s some specific modules that I found difficult to reckon with. For example, when trying to call on processor, there is no indication of text, image, or config, which was problematic when I was trying to get it to work with a CLIP model that only wanted an image input. 
Screenshot 2025-01-14 at 12 45 07

For comparison, looking at CLIPProcessor in the python transformers library, the `__call__` method is allowed to specify inputs, which is incredibly helpful. It also has plenty of information about the method
Screenshot 2025-01-14 at 12 50 35

I believe this stems from using the Callable structure. Is there a way to have a JS equivalent in terms of typing and information on the callable? Maybe there's additional methods that could be added/exposed that provide an alternate way in, like `processText`, `processImage`, and `processTextAndImage`? Would love to hear if maybe this has already been explored, or if it's a problem that could be worth solving.

Ultimately, I genuinely just want to help make this library a little more approachable and thought I saw a way I could help out. I understand it’s not how you would approach it. But honestly, for as kind as xenova has been in all his communications, it’s a bummer to be on the receiving end of comments like yours. There’s zero reason why we can’t talk about a problem and be nice to each other.

BritishWerewolf:
To clarify, I never meant to offend with my comments. I just didn't want you to spend anymore time translating anything else into TypeScript when there is already a type system in place.

Without digging into the code myself, I'm not 100% sure why the IDE doesn't work properly for Processors - I am sure there are comments.  
Perhaps that is something you can look into and you can create a PR for that?

The work you did here is fantastic, and your comments and thorough and in-depth; I am confident in saying that your pull requests will be greatly appreciated again in future.  
Please don't let this dissuade you from working on the project in future!

tsekiguchi:
Thank you! I really do appreciate that. To be clear it wasn't your comment that I was referencing.

I'll do some more digging and see if there isn't a solution for JSDoc, and will submit a PR if I find anything :)

PR #1161, opened 2025-01-20
fix package export

fs-eire:
### Summary

This PR includes a few changes that helps to fix the package export. Now transformers.js has the following exports:

| usage | file | description |
|------|------|-------------|
|   CDN <br/> `import {...} from "https://cdn.jsdelivr.net/npm/@huggingface/transformers@{version}"`  |    `dist/transformers[.min].js`  | This is a JavaScript file that bundled with all its JS dependencies. |
|   Node.js/Deno (ESM/CJS) <br/> `import {...} from "@huggingface/transformers"` <br/> `const {...} = require("@huggingface/transformers")` |    `dist/transformers.node[.min].mjs` <br/> `dist/transformers.node[.min].cjs` |  Treat `onnxruntime-common` and `onnxruntime-node` as external; Ignore `onnxruntime-web`     |
| Web (ESM) <br/> `import {...} from "@huggingface/transformers"` | `dist/transformers.web[.min].js` | Treat `onnxruntime-common` and `onnxruntime-web` as external; Ignore `onnxruntime-node` |

### Problem

There are complains about different kinds of failures when users try to import transformers.js and/or onnxruntime-web in their web applications.

It took me quite a while to fix most of the problems in onnxruntime-web. The latest dev version of onnxruntime-web should have worked for all known major use cases.

When things come to transformers.js, the situation become more complicated. After some investigation, I found that some import/export problems that resolved in onnxruntime-web occurred in transformers.js again. This is because when building transformers.js, webpack generated some code that will not work for using as input for bundler again.

### Solution

There are 2 major changes in this PR:
- remove the override of `env.wasm.wasmPaths`
- make `onnxruntime-web` as external dependency of transformers.js

#### Why treat `onnxruntime-web` as external dependency

ONNX Runtime Web has a few workarounds for webpack to ensure out-of-the-box user experience. Those workarounds requires the code as-is when webpack starts to process it. The workarounds cannot keep in the bundled JS file.

To make sure of backward compatibility, this change still keeps the file `dist/transformers[.min].js` unchanged. Instead, a new file `dist/transformers.web[.min].js` is added and replaced in the "exports" property of the package.json.

### Tests

The following tests are passed when applying this PR:

- [x] CDN
      https://github.com/huggingface/transformers.js-examples/tree/migrate-examples/segment-anything-webgpu
- [x] Next.js (Server, onnxruntime-node)
      https://github.com/huggingface/transformers.js-examples/tree/migrate-examples/next-server
      * some issue with the "standalone" mode but not related to package export.
      * need workaround for Nextjs. See [1] below
- [x] Next.js (Client, onnxruntime-web)
      * need workaround for Nextjs. See [1] below
- [x] Vite
      https://github.com/huggingface/transformers.js-examples/tree/migrate-examples/remove-background-webgpu
      * need workaround for Vite. See [2] below
- [x] Vite + worker
      https://github.com/huggingface/transformers.js-examples/tree/migrate-examples/phi-3.5-webgpu
      * need workaround for Vite. See [2] below
- [x] Service worker
      https://github.com/microsoft/onnxruntime/pull/20991

** [1] - For Next.js based app, a [workaround](https://github.com/huggingface/transformers.js-examples/blob/a48196ab77182b4c0e221eb0eb4fd4f19e545111/next-server/next.config.mjs#L3-L4) is needed if any code statically imports `@huggingface/transformers`. This is because Next.js tries to bundle the code for server even if those code starts with `"use client";`. This happens to both dev mode and prod mode, and to both default and turbopack bundler.
** [2] - For Vite 5.x based app, a [workaround](https://github.com/microsoft/onnxruntime/blob/41a49617523b2d5cd40552fc2fde507c111c9a66/js/web/test/e2e/exports/testcases/vite-default/vite.config.js#L6-L11) is needed according to https://github.com/vitejs/vite/issues/8427. This is a known issue affecting all Vite app depending on WebAssembly based packages, and may be fixed in Vite 6.x

### DIFF
--- a/src/backends/onnx.js
+++ b/src/backends/onnx.js
@@ -57,8 +57,8 @@ let ONNX;
 const ORT_SYMBOL = Symbol.for('onnxruntime');
 
 if (ORT_SYMBOL in globalThis) {
-  // If the JS runtime exposes their own ONNX runtime, use it
-  ONNX = globalThis[ORT_SYMBOL];
+    // If the JS runtime exposes their own ONNX runtime, use it
+    ONNX = globalThis[ORT_SYMBOL];
 
 } else if (apis.IS_NODE_ENV) {
     ONNX = ONNX_NODE.default ?? ONNX_NODE;
@@ -175,11 +175,20 @@ const ONNX_ENV = ONNX?.env;
 if (ONNX_ENV?.wasm) {
     // Initialize wasm backend with suitable default settings.
 
-    // (Optional) Set path to wasm files. This is needed when running in a web worker.
-    // https://onnxruntime.ai/docs/api/js/interfaces/Env.WebAssemblyFlags.html#wasmPaths
-    // We use remote wasm files by default to make it easier for newer users.
-    // In practice, users should probably self-host the necessary .wasm files.
-    ONNX_ENV.wasm.wasmPaths = `https://cdn.jsdelivr.net/npm/@huggingface/transformers@${env.version}/dist/`;
+    // (Optional) Set path to wasm files. This will override the default path search behavior of onnxruntime-web.
+    // Currently, this behavior is disabled. To enable it, uncomment the following code.
+
+    // // Override the wasm search path if:
+    // // 1. We are not in a service worker.
+    // // 2. ONNX Runtime Web version is defined.
+    // // 3. The wasmPaths are not already set.
+    // //
+    // // @ts-ignore Cannot find name 'ServiceWorkerGlobalScope'.ts(2304)
+    // if (!(typeof ServiceWorkerGlobalScope !== 'undefined' && self instanceof ServiceWorkerGlobalScope)
+    //     && typeof ONNX?.versions?.web === 'string'
+    //     && !ONNX_ENV.wasm.wasmPaths) {
+    //     ONNX_ENV.wasm.wasmPaths = `https://cdn.jsdelivr.net/npm/onnxruntime-web@${ONNX.versions.web}/dist/`;
+    // }
 
     // TODO: Add support for loading WASM files from cached buffer when we upgrade to onnxruntime-web@1.19.0
     // https://github.com/microsoft/onnxruntime/pull/21534
@@ -187,11 +196,6 @@ if (ONNX_ENV?.wasm) {
     // Users may wish to proxy the WASM backend to prevent the UI from freezing,
     // However, this is not necessary when using WebGPU, so we default to false.
     ONNX_ENV.wasm.proxy = false;
-
-    // https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated
-    if (typeof crossOriginIsolated === 'undefined' || !crossOriginIsolated) {
-        ONNX_ENV.wasm.numThreads = 1;
-    }
 }
 
 if (ONNX_ENV?.webgpu) {
--- a/src/backends/onnx.js
+++ b/src/backends/onnx.js
@@ -176,19 +176,18 @@ if (ONNX_ENV?.wasm) {
     // Initialize wasm backend with suitable default settings.
 
     // (Optional) Set path to wasm files. This will override the default path search behavior of onnxruntime-web.
-    // Currently, this behavior is disabled. To enable it, uncomment the following code.
 
-    // // Override the wasm search path if:
-    // // 1. We are not in a service worker.
-    // // 2. ONNX Runtime Web version is defined.
-    // // 3. The wasmPaths are not already set.
-    // //
-    // // @ts-ignore Cannot find name 'ServiceWorkerGlobalScope'.ts(2304)
-    // if (!(typeof ServiceWorkerGlobalScope !== 'undefined' && self instanceof ServiceWorkerGlobalScope)
-    //     && typeof ONNX?.versions?.web === 'string'
-    //     && !ONNX_ENV.wasm.wasmPaths) {
-    //     ONNX_ENV.wasm.wasmPaths = `https://cdn.jsdelivr.net/npm/onnxruntime-web@${ONNX.versions.web}/dist/`;
-    // }
+    // Override the wasm search path if:
+    // 1. We are not in a service worker.
+    // 2. ONNX Runtime Web version is defined.
+    // 3. The wasmPaths are not already set.
+    //
+    // @ts-ignore Cannot find name 'ServiceWorkerGlobalScope'.ts(2304)
+    if (!(typeof ServiceWorkerGlobalScope !== 'undefined' && self instanceof ServiceWorkerGlobalScope)
+        && typeof ONNX?.versions?.web === 'string'
+        && !ONNX_ENV.wasm.wasmPaths) {
+        ONNX_ENV.wasm.wasmPaths = `https://cdn.jsdelivr.net/npm/onnxruntime-web@${ONNX.versions.web}/dist/`;
+    }
 
     // TODO: Add support for loading WASM files from cached buffer when we upgrade to onnxruntime-web@1.19.0
     // https://github.com/microsoft/onnxruntime/pull/21534
--- a/src/backends/onnx.js
+++ b/src/backends/onnx.js
@@ -176,17 +176,13 @@ if (ONNX_ENV?.wasm) {
     // Initialize wasm backend with suitable default settings.
 
     // (Optional) Set path to wasm files. This will override the default path search behavior of onnxruntime-web.
-
-    // Override the wasm search path if:
-    // 1. We are not in a service worker.
-    // 2. ONNX Runtime Web version is defined.
-    // 3. The wasmPaths are not already set.
-    //
-    // @ts-ignore Cannot find name 'ServiceWorkerGlobalScope'.ts(2304)
-    if (!(typeof ServiceWorkerGlobalScope !== 'undefined' && self instanceof ServiceWorkerGlobalScope)
-        && typeof ONNX?.versions?.web === 'string'
-        && !ONNX_ENV.wasm.wasmPaths) {
-        ONNX_ENV.wasm.wasmPaths = `https://cdn.jsdelivr.net/npm/onnxruntime-web@${ONNX.versions.web}/dist/`;
+    // By default, we only do this if we are not in a service worker and the wasmPaths are not already set.
+    if (
+        // @ts-ignore Cannot find name 'ServiceWorkerGlobalScope'.ts(2304)
+        !(typeof ServiceWorkerGlobalScope !== 'undefined' && self instanceof ServiceWorkerGlobalScope)
+        && !ONNX_ENV.wasm.wasmPaths
+    ) {
+        ONNX_ENV.wasm.wasmPaths = `https://cdn.jsdelivr.net/npm/@huggingface/transformers@${env.version}/dist/`;
     }
 
     // TODO: Add support for loading WASM files from cached buffer when we upgrade to onnxruntime-web@1.19.0
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -24,35 +24,12 @@ class PostBuildPlugin {
         const file = path.join(dist, ORT_BUNDLE_FILE);
         if (fs.existsSync(file)) fs.unlinkSync(file);
       }
-      
+
       // 2. Copy unbundled JSEP file
       {
         const src = path.join(__dirname, 'node_modules/onnxruntime-web/dist', ORT_JSEP_FILE);
         const dest = path.join(dist, ORT_JSEP_FILE);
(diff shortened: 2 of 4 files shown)


examples/extension/src/background.js
```diff
@@ -7,7 +7,7 @@ env.allowLocalModels = false;
 
 // Due to a bug in onnxruntime-web, we must disable multithreading for now.
 // See https://github.com/microsoft/onnxruntime/issues/14445 for more information.
-env.backends.onnx.wasm.numThreads = 1;
+//env.backends.onnx.wasm.numThreads = 1;
```
fs-eire, line 10: onnxruntime-web should have always deal with numThreads correctly. (ie. the default value should not fail its function) 

explicitly setting numThreads to 1 is no longer necessary.

Need help verification whether this is true.

src/backends/onnx.js
```diff
@@ -189,9 +189,9 @@ if (ONNX_ENV?.wasm) {
     ONNX_ENV.wasm.proxy = false;
 
     // https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated
-    if (typeof crossOriginIsolated === 'undefined' || !crossOriginIsolated) {
-        ONNX_ENV.wasm.numThreads = 1;
-    }
+    // if (typeof crossOriginIsolated === 'undefined' || !crossOriginIsolated) {
+    //     ONNX_ENV.wasm.numThreads = 1;
+    // }
```
fs-eire, line 194: mentioned above.

webpack.config.js
```diff
@@ -42,7 +42,7 @@ function buildConfig({
         type,
       },
       assetModuleFilename: "[name][ext]",
-      chunkFormat: "module",
+      chunkFormat: false,
```
fs-eire, line 45: This is why the final bundle includes `new URL("./", import.meta.url)`.

Setting it to false to disable the chunks feature of webpack (which is not used anyway)

webpack.config.js
```diff
@@ -78,13 +78,13 @@ function buildConfig({
...
-        },
-      },
-    };
+    // config.module = {
+    //   parser: {
+    //     javascript: {
+    //       importMeta: false,
+    //     },
+    //   },
+    // };
```
fs-eire, line 87: ### Background:

webpack rewrite `import.meta.url` into a local file path (eg. "file://my/source/code/transformers/dist/transformers.min.js"), which is an unexpected (and incorrect) behavior.

There is a historical reason for this: when the old days browsers does not support `import.meta`, Webpack uses this behavior to allow writing code that uses `import.meta.url` in the following specific format:
`new URL('my-file.js', import.meta.url)`

By rewriting `import.meta.url` into a "file://" URL string, the whole URL can be located into a local file path and webpack can load and bundle the file my-file.js. This was useful in the old days.

Unfortunately, this behavior is conflict with the latest standard and because a lot of users already depends on this, it's hard to change.

### The old solution

The old solution is to add the config as commented above. It disabled the wrong behavior of webpack so that the generated transformer.min.js works.

However, if users of transformer.js uses webpack but they did not include the same config above, webpack will still rewriting `import.meta.url` into a "file://" URL.

This is frustrating so we need to find a new way to make it work with users when using default config.

### The new solution

in short, it is not something very elegant:

```
url = import.meta.url.startsWith('file://') ?
               new URL('ort.bundle.min.mjs', import.meta.url)
             : import.meta.url;
```

If the bundler kicks in, then the variable will be a processed URL that generated by bundler; otherwise it will be the unmodified `import.meta.url`.

This is already in onnxruntime-web. So here we should just remove the config.

src/backends/onnx.js
```diff
@@ -179,7 +179,7 @@ if (ONNX_ENV?.wasm) {
     // https://onnxruntime.ai/docs/api/js/interfaces/Env.WebAssemblyFlags.html#wasmPaths
     // We use remote wasm files by default to make it easier for newer users.
     // In practice, users should probably self-host the necessary .wasm files.
-    ONNX_ENV.wasm.wasmPaths = `https://cdn.jsdelivr.net/npm/@huggingface/transformers@${env.version}/dist/`;
+    // ONNX_ENV.wasm.wasmPaths = `https://cdn.jsdelivr.net/npm/@huggingface/transformers@${env.version}/dist/`;
```
fs-eire, line 182: Background is explained in https://github.com/microsoft/onnxruntime/pull/23257#issuecomment-2593753716.

there are actually 2 options:

1. keep this line, so that anyone using transformer.js always load the .mjs and .wasm from CDN.
   - pros:
     - consistent behavior and more robust to bundlers
   - cons:
     - users are not loading resources from their deployment (sometimes they expect it)
     - not working for service worker, but is possible to workaround
2. remove this line, so that the .mjs and .wasm is loaded following default behavior.
   - pros:
     - leverage onnxruntime-web's default behavior for best-effort optimizing for works with bundler in a way that as it's designed
       Should work with nextjs(webpack) and vite(rollup)
   - cons:
     - so far I cannot make it work with all bundlers in default config. For example, in Vite (rollup) I need to add the following line:
       ```js
       import wasmFilePath from "../../node_modules/onnxruntime-web/dist/ort-wasm-simd-threaded.jsep.wasm";
       // need a change in onnxruntime-web to use something like this:
       // import wasmFilePath from "onnxruntime-web/assets/.wasm";

       env.backands.onnx.wasm.wasmPaths = { wasm: wasmFilePath };
       ```

fs-eire: It's hard to say which one is better. But if you want to use (1), it's probably better to point to the CDN to package onnxruntime-web. This helps to remove the ort-wasm-simd-threaded.jsep.mjs file from the NPM package of transformerjs.

fs-eire:
@xenova what do you think of the proposed changes?

src/backends/onnx.js
fs-eire: ### Latest Update:

treating onnxruntime-web as external dependency will make the bundler behavior consistent so my suggestion is to not to set `wasmPaths` here.

fs-eire:
> @xenova what do you think of the proposed changes?

PR is ready for review.

xenova:
Thanks @fs-eire! I will review this fully soon!

src/backends/onnx.js
```diff
@@ -175,23 +175,27 @@ const ONNX_ENV = ONNX?.env;
...
     // https://github.com/microsoft/onnxruntime/pull/21534
 
     // Users may wish to proxy the WASM backend to prevent the UI from freezing,
     // However, this is not necessary when using WebGPU, so we default to false.
     ONNX_ENV.wasm.proxy = false;
-
-    // https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated
-    if (typeof crossOriginIsolated === 'undefined' || !crossOriginIsolated) {
-        ONNX_ENV.wasm.numThreads = 1;
-    }
```
sroussey, line 194: why is this included?

fs-eire: there was a bug that in a non-cross origin isolated environment, onnxruntime-web will use multiple threads and then fail the initialization. This code snippet was a workaround for that bug.

Now in latest onnxruntime-web, this check is included so no necessary to do it again in transformerjs

sroussey: Ah nice. I’ve run into this error so was curious.

src/backends/onnx.js
```diff
@@ -175,23 +175,27 @@ const ONNX_ENV = ONNX?.env;
...
+    // // 1. We are not in a service worker.
+    // // 2. ONNX Runtime Web version is defined.
+    // // 3. The wasmPaths are not already set.
+    // //
+    // // @ts-ignore Cannot find name 'ServiceWorkerGlobalScope'.ts(2304)
+    // if (!(typeof ServiceWorkerGlobalScope !== 'undefined' && self instanceof ServiceWorkerGlobalScope)
+    //     && typeof ONNX?.versions?.web === 'string'
+    //     && !ONNX_ENV.wasm.wasmPaths) {
+    //     ONNX_ENV.wasm.wasmPaths = `https://cdn.jsdelivr.net/npm/onnxruntime-web@${ONNX.versions.web}/dist/`;
+    // }
```
xenova, line 191: We've run into several issues in the past where the wasm files weren't able to be loaded locally, so we default to the CDN. Maybe in future we can switch to this, but I think we can keep it for now.

Another reason is that the raw .wasm file is ~21MB, but the CDN (gzipped) version is ~4MB. It might be a good idea to find a way to compress the raw file too. cc @fs-eire @guschmue

src/backends/onnx.js
xenova: Good to know! Thanks @fs-eire!

webpack.config.js
xenova: Good find! Thanks!

xenova (commented):
Thanks for this PR! I'm doing some more testing now.

src/backends/onnx.js
fs-eire: No problem. I will make a change to keep them.

> We've run into several issues in the past where the wasm files weren't able to be loaded locally

This is indeed the case back to a few versions ago. I made it much better now but I still cannot say it's working for 100% cases. 

> the CDN (gzipped) version is ~4MB.

I think it is already the default setting for all modern server to enable gzip. A dev server usually does not enable this feature. However I didn't check the file size.

xenova:
Just FYI, I'll be looking to integrate this update into the next minor release (v3.4).

endomorphosis:
> Just FYI, I'll be looking to integrate this update into the next minor release (v3.4).

hopefully it fixes a most obnoxious bug that I've been having when trying to run webgpu in a ServiceWorker context inside of a chrome extension.

```
Error in generate function: Error: no available backend found. ERR: [webgpu] TypeError: import() is disallowed on ServiceWorkerGlobalScope by the HTML specification. See https://github.com/w3c/ServiceWorker/issues/1356.
```

tsekiguchi (commented):
This change was instrumental in getting ONNX to work in a WebWorker in production when using Vite / SvelteKit. Before, I kept getting the error `No backend available [webgpu]`. Thank you for this.

gumelarme:
Hi, I am working on an extension, and after applying this PR I'm having build errors:

```
Failed to resolve entry for package "@huggingface/transformers". The package may have incorrect main/module/exports specified in its package.json.
```

I'm using WXT framework with Vite 5.4.14 
Is this a vite 5 error thing? because the repro repo is using version 6
Is there a work around for this?

Thanks

xenova:
Seems related to this workaround: https://github.com/microsoft/onnxruntime/blob/99c51a326e0ff54a56e7b194204d459932084408/js/web/script/build.ts#L120-L132

fs-eire:
It's expected because the override to `env.wasm.wasmPaths` is not always applied. For example, for service worker usage, it is not applied (you cannot fetch CDN in this context). Also, it should still work if user manually set `env.wasm.wasmPaths = undefined;`, where they should be able to use the wasm files from their dist folder.

The only one file that is not used in Vite (the current stable version) is the "ort.bundle.min<renamed>.mjs". However this is unfortunately because of a workaround that aiming to fix webpack usages. There is no way to use different files for different bundlers.

xenova:
@fs-eire https://github.com/xenova/onnxruntime/commit/9c50aa2c63bad4cb73ad77ff1c43e0c43da0907f works for me, but I'm not sure how this may affect other bundlers. (just a simple rename to prevent vite from modifying this line).

fs-eire:
> @fs-eire [xenova/onnxruntime@9c50aa2](https://github.com/xenova/onnxruntime/commit/9c50aa2c63bad4cb73ad77ff1c43e0c43da0907f) works for me, but I'm not sure how this may affect other bundlers. (just a simple rename to prevent vite from modifying this line).

Thank you for sharing this trick! Please give me some time to test if this works in other use cases.

xenova:
> > @fs-eire [xenova/onnxruntime@9c50aa2](https://github.com/xenova/onnxruntime/commit/9c50aa2c63bad4cb73ad77ff1c43e0c43da0907f) works for me, but I'm not sure how this may affect other bundlers. (just a simple rename to prevent vite from modifying this line).
> 
> Thank you for sharing this trick! Please give me some time to test if this works in other use cases.

There may be vite comments/directives (like `/* @vite-ignore */` which could be used (I'm not 100% sure about the usage here though), in case a better solution exists too!

xenova:
@fs-eire just FYI, this is the last blocker before the v3.4 release (hoping to get done in the next few days), so if we can sort this out, that'd be great!

xenova:
Re-opening since I think it was automatically closed

fs-eire:
Published onnxruntime-web@1.22.0-dev.20250306-aafa8d170a

Issue #1179, opened 2025-02-01
Can't use large models with pipeline()


sroussey:
### System Info

Example:

```js
        const p = pipeline('text-generation', 'Xenova/Phi-3-mini-4k-instruct', {
            device: 'webgpu',
            dtype: 'q4',
        });
```

I see this error in the console:

```
Uncaught (in promise) Error: Can't create a session. ERROR_CODE: 1, ERROR_MESSAGE: Deserialize tensor model.layers.5.mlp.gate_proj.MatMul.weight_Q4 failed.Failed to load external data file ""model_q4.onnx_data"", error: Module.MountedFiles is not available.
```

Seeing that `onnx_data` is the issue, I figured I needed to pass `use_external_data_format` along, but it does not work.

I have tried :
```js
        const p = pipeline('text-generation', 'Xenova/Phi-3-mini-4k-instruct', {
            device: 'webgpu',
            dtype: 'q4',
            use_external_data_format: true,
        });
```
and 
```js
        const p = pipeline('text-generation', 'Xenova/Phi-3-mini-4k-instruct', {
            device: 'webgpu',
            dtype: 'q4',
            session_options: {use_external_data_format: true},
        });
```
But neither of these will load the model correctly.

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension
- [ ] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

see above

### Reproduction

see above

sroussey:
I altered the source to pass `use_external_data_format` through, and I see that the extra data file loads. However, it seems more difficult to use this way... like how do you use a chat template?

xenova:
Thanks for https://github.com/huggingface/transformers.js/pull/1180! I don't know why this wasn't added before 👀 

>  However, it seems more difficult to use this way... like how do you use a chat template?

You can just pass in the messages object (and it will do the templating for you)

If you want access to the tokenizer to do things yourself, you can access `pipeline.tokenizer.apply_chat_template`

sroussey:
BTW: if you run the code I did above without the fix, it loads most of the files, but not all, and so fails.

Problem number 2 is that if you then fix the transformers code, it will still fail since not all the files were downloaded correctly, but it thinks it was.

I need to open DevTools, go to Application tab, then to Cache Storage, and then delete transformers-cache by right clicking and choosing delete. Only then will the code work. 

[...]

PR #1180, opened 2025-02-01
Add option to use external data format in pipeline initialization

sroussey:
This is for large models. See issue #1179

### DIFF
--- a/src/pipelines.js
+++ b/src/pipelines.js
@@ -3301,6 +3301,7 @@ export async function pipeline(
         dtype = null,
         model_file_name = null,
         session_options = {},
+        use_external_data_format = false,
     } = {}
 ) {
     // Helper method to construct pipeline
@@ -3331,6 +3332,7 @@ export async function pipeline(
         dtype,
         model_file_name,
         session_options,
+        use_external_data_format,
     }
 
     const classes = new Map([


benc-uk:
Wait what? So the option `use_external_data_format` wasn't being used at all, when creating a pipeline !?
🤯🤯🤯

sroussey:
> Wait what? So the option `use_external_data_format` wasn't being used at all, when creating a pipeline !? 🤯🤯🤯

I don't have the code in front of me, but I think you can use session_options as a workaround:

pipeline("...", { session_options: use_external_data_format

benc-uk:
> session_options

Setting it under `session_options` didn't make a difference.
Even if it did the levels of undocumented configuration this project has is unfathomable

sroussey:
I could publish my version with the fix if you are in a bind. Was thinking of starting to do that again to get some things I want where I have made changes. 

As for as docs... the API stuff looks like it comes from source. Which is likely why the docs callback params are non-existent.

I'm trying to learn the code a bit by translating into TypeScript (transformers.ts!). But I'm trying not to change the code while I do it (or even the JSDocs) as I don't want to change too many lines of code. I really should keep some notes as I go though.

I think I can make the bundle a bit smaller and get better types. 

I just do when I'm frustrated at something else I'm working on.

xenova:
Thanks for the discussion :) I've answered a similar question at https://github.com/huggingface/transformers.js/issues/1179#issuecomment-2691792832, so I'll close this one and we can continue discussion there.

> Even if it did the levels of undocumented configuration this project has is unfathomable

Feel free to open PRs! 🤗

PR #1184, opened 2025-02-07
minor fix: correct demo site link from xenova to hf

viksit:
minor fix: correct demo site link from xenova to hf


xenova (approved):
Thanks!

PR #1189, opened 2025-02-11
Added target to playground link

damianr13:
Currently if I click the `locally` link in the course page, it breaks the iframe: 

image

This PR adds `target="_blank"` to that link and fixes this issue.


xenova:
Hi there 👋 Thanks for the PR! We have moved the examples source code to https://github.com/huggingface/transformers.js-examples; could you re-open this there?

PR #1190, opened 2025-02-12
Added request_options parameter to various classes and functions for canceling requests

emojiiii:
Added request_options parameter to various classes and functions for canceling requests

### DIFF
--- a/src/configs.js
+++ b/src/configs.js
@@ -367,6 +367,7 @@ export class PretrainedConfig {
         cache_dir = null,
         local_files_only = false,
         revision = 'main',
+        request_options = {},
     } = {}) {
         if (config && !(config instanceof PretrainedConfig)) {
             config = new PretrainedConfig(config);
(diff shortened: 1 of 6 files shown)


xenova:
This can be very useful! Thanks! Regarding the API, since this is pretty non-standard for the python library, perhaps a better approach would be to allow the user to pass their own `fetch` function instead of only allowing them to pass options? This could be useful for loading models from custom filesystems too.

emojiiii:
> This can be very useful! Thanks! Regarding the API, since this is pretty non-standard for the python library, perhaps a better approach would be to allow the user to pass their own `fetch` function instead of only allowing them to pass options? This could be useful for loading models from custom filesystems too.
> 
> In fact, I think this would be better at the `env` level. something like `env.fetch = ...`

You're right, I need some time to revisit it.

emojiiii:
I added a custom fetch method, but after some thought, I realized that I still need to pass request_options as I do currently. Otherwise, every time an AbortController is used, I would need to reset env.customFetch, and there would be no way to cancel only specific fetches.

PR #1193, opened 2025-02-14
feat: Add AbortSignal support across library components

sroussey:
- Introduce `abort_signal` parameter to multiple methods and constructors
- Update file retrieval and loading mechanisms to support request cancellation
- Add AbortSignal handling in pipelines, models, tokenizers, and utility functions

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,0 +1,1515 @@
+{
+  "lockfileVersion": 1,
+  "workspaces": {
+    "": {
+      "name": "@huggingface/transformers",
+      "dependencies": {
+        "@huggingface/jinja": "^0.3.3",
+        "onnxruntime-node": "1.20.1",
+        "onnxruntime-web": "1.21.0-dev.20250206-d981b153d3",
+        "sharp": "^0.33.5",
+      },
+      "devDependencies": {
+        "@types/jest": "^29.5.14",
+        "@types/node": "^22.10.1",
+        "@webgpu/types": "^0.1.51",
+        "catharsis": "github:xenova/catharsis",
+        "jest": "^30.0.0-alpha.6",
+        "jest-environment-node": "^30.0.0-alpha.6",
+        "jsdoc-to-markdown": "^9.1.1",
+        "prettier": "3.4.2",
+        "typescript": "^5.7.2",
+        "wavefile": "11.0.0",
+        "webpack": "^5.97.1",
+        "webpack-cli": "^5.1.4",
+        "webpack-dev-server": "^5.1.0",
+      },
+    },
+  },
+  "trustedDependencies": [
+    "protobufjs",
+    "onnxruntime-node",
+    "onnxruntime-web",
+    "sharp",
+    "catharsis",
+  ],
+  "packages": {
+    "@ampproject/remapping": ["@ampproject/remapping@2.3.0", "", { "dependencies": { "@jridgewell/gen-mapping": "^0.3.5", "@jridgewell/trace-mapping": "^0.3.24" } }, "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw=="],
+
+    "@babel/code-frame": ["@babel/code-frame@7.26.2", "", { "dependencies": { "@babel/helper-validator-identifier": "^7.25.9", "js-tokens": "^4.0.0", "picocolors": "^1.0.0" } }, "sha512-RJlIHRueQgwWitWgF8OdFYGZX328Ax5BCemNGlqHfplnRT9ESi8JkFlvaVYbS+UubVY6dpv87Fs2u5M29iNFVQ=="],
+
+    "@babel/compat-data": ["@babel/compat-data@7.26.2", "", {}, "sha512-Z0WgzSEa+aUcdiJuCIqgujCshpMWgUpgOxXotrYPSA53hA3qopNaqcJpyr0hVb1FeWdnqFA35/fUtXgBK8srQg=="],
+
+    "@babel/core": ["@babel/core@7.26.0", "", { "dependencies": { "@ampproject/remapping": "^2.2.0", "@babel/code-frame": "^7.26.0", "@babel/generator": "^7.26.0", "@babel/helper-compilation-targets": "^7.25.9", "@babel/helper-module-transforms": "^7.26.0", "@babel/helpers": "^7.26.0", "@babel/parser": "^7.26.0", "@babel/template": "^7.25.9", "@babel/traverse": "^7.25.9", "@babel/types": "^7.26.0", "convert-source-map": "^2.0.0", "debug": "^4.1.0", "gensync": "^1.0.0-beta.2", "json5": "^2.2.3", "semver": "^6.3.1" } }, "sha512-i1SLeK+DzNnQ3LL/CswPCa/E5u4lh1k6IAEphON8F+cXt0t9euTshDru0q7/IqMa1PMPz5RnHuHscF8/ZJsStg=="],
+
+    "@babel/generator": ["@babel/generator@7.26.2", "", { "dependencies": { "@babel/parser": "^7.26.2", "@babel/types": "^7.26.0", "@jridgewell/gen-mapping": "^0.3.5", "@jridgewell/trace-mapping": "^0.3.25", "jsesc": "^3.0.2" } }, "sha512-zevQbhbau95nkoxSq3f/DC/SC+EEOUZd3DYqfSkMhY2/wfSeaHV1Ew4vk8e+x8lja31IbyuUa2uQ3JONqKbysw=="],
+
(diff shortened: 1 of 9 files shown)


sroussey (commented):
@xenova -- thoughts?

emojiiii:
Looks like the same as #1190

sroussey:
> Looks like the same as #1190

I will have a look

sroussey:
@xenova a few thoughts on what i did:

I added abort_signal to PretrainedOptions. Most places just pass the options around, but many places rebuild them, so you will see more changes than expected because of this.

I also added it to the ModelTokenizerProcessorConstructorArgs constructor and there are edge cases like _call_text_to_spectrogram. Not sure I like this though. :/

This PR so far only handles downloading models, etc.

Do you know where I should look as far as ONNX/ORT to cancel generation?

sroussey:
> > Looks like the same as #1190
> 
> I will have a look

I can see reasons to have a custom fetch() but I think those are orthogonal to the ability to abort.

But perhaps we could have both.

I want to be able to abort generation as well as downloads, in which case fetch is not enough.

sroussey:
Also, I made `abort_signal` required in various places... that is to help me find the places I have missed. I will make it optional when done.

joecal:
> > > Looks like the same as #1190
> > 
> > 
> > I will have a look
> 
> I can see reasons to have a custom fetch() but I think those are orthogonal to the ability to abort.
> 
> But perhaps we could have both.
> 
> I want to be able to abort generation as well as downloads, in which case fetch is not enough.

I’ve been trying to figure out a way to stop text generation and so far I’ve come up with the following:

Something like this `abortOnSignal` helper function to wrap async operations to abort:

```javascript

/**
 * A helper function to wrap asynchronous operations that can be stopped/aborted by an AbortSignal.
 * @param {Promise<unknown>} promise Any promise/async operation to be aborted.
 * @param {AbortSignal | null | undefined} signal The abort signal from an AbortController that can cancel the wrapped promise/async operation.
 * @returns {Promise<unknown>} Either the param promise or a rejected promise with AbortError if aborted.
 */
export function abortOnSignal(promise, signal) {
    if (!signal) {
        return promise;
    }
    return new Promise((resolve, reject) => {
        const abortHandler = () => {
            reject(new DOMException('Aborted', 'AbortError'));
        };
        signal.addEventListener('abort', abortHandler, { once: true });
        promise.then(resolve, reject)
            .finally(() => {
                signal.removeEventListener('abort', abortHandler);
            });
    });
}
```

Then add an abort `signal` param to the `sessionRun` function in `models.js` and wrap the `session.run` invocation with the `abortOnSignal` function with the `signal` param passed into it:

```javascript

/**
 * Executes an InferenceSession using the specified inputs.
 * NOTE: `inputs` must contain at least the input names of the model.
 *  - If additional inputs are passed, they will be ignored.
 *  - If inputs are missing, an error will be thrown.
 * 
 * @param {Object} session The InferenceSession object to run.
 * @param {Object} inputs An object that maps input names to input tensors.
 * @param {AbortSignal | null | undefined} signal Optional abort signal from an AbortController to cancel the InferenceSession.
 * @returns {Promise<Object>} A Promise that resolves to an object that maps output names to output tensors.
 * @private
 */
async function sessionRun(session, inputs, signal = null) {
    const checkedInputs = validateInputs(session, inputs);
    try {
        // pass the original ort tensor
        const ortFeed = Object.fromEntries(Object.entries(checkedInputs).map(([k, v]) => [k, v.ort_tensor]));
        let output = await abortOnSignal(session.run(ortFeed), signal);
        output = replaceTensors(output);
        return output;
    } catch (e) {
        if (e.name === 'AbortError') {
            console.log("Generation aborted by user.");
        }
        // Error messages can be long (nested) and uninformative. For this reason,
        // we apply minor formatting to show the most important information
        const formatted = Object.fromEntries(Object.entries(checkedInputs)
            .map(([k, { type, dims, data }]) => [k, {
                // Extract these properties from the underlying ORT tensor
                type, dims, data,
            }]));

        // This usually occurs when the inputs are of the wrong type.
        console.error(`An error occurred during model execution: "${e}".`);
        console.error('Inputs given to model:', formatted);
        throw e;
    }
}
```

To use it you pass the signal into the generator as an argument:

```javascript
const prompt = “How many letter ‘r’s are in the word ‘strawberry’?” 
const controller = new AbortController();
const generator = await pipeline('text-generation', MODEL, {
    dtype: 'q4f16',
    device: 'webgpu'
});
const generationArgs = { max_new_tokens: maxTokens, streamer: streamer, signal: controller.signal };
await generator(prompt, generationArgs);
```

Here’s a little demo of it working (Ignore the text output nonsense. I think my laptop processor is messed up somehow):


https://github.com/user-attachments/assets/30595d06-28f8-40b9-8f52-fde8f5a1d52c



It works at stopping the generation but the way I have it now feels kinda messy passing the `signal` around so much. 

Take `TextGenerationPipeline` for example. The signal gets passed in through its `_call` args into `this.model.generate` then into `this.forward` then into `this._forward` which is `decoderForward` then into `decoderForward`s invocation of `sessionRun` then finally into the `abortOnSignal` wrapped around `session.run`. 

I'm sure there's a cleaner way to do it. Would love to get your guys take on this approach.

sroussey:
I haven't tried it myself yet, but it is definitely undocumented. I don't see signal or anything in the docs. 

https://onnxruntime.ai/docs/api/js/interfaces/InferenceSession.SessionOptions.html#freeDimensionOverrides

sroussey:
Oh, I see now... you are simulating an abort. The text generation keeps going though.

If you are using progress callbacks, you can just ignore them after the user hits stop to the same effect.

joecal:
> Oh, I see now... you are simulating an abort. The text generation keeps going though.
> 
> If you are using progress callbacks, you can just ignore them (after the user hits stop) to the same effect.

Oh, the text generation actually does stop completely when the abort signal is triggered. I did some testing with Chrome dev tools and the CPU usage drops to 0% when aborted, showing the generation process fully terminates.  

The way it works is:
1. When the abort signal triggers, the `abortOnSignal` wrapper rejects the ONNX session.run() Promise with an AbortError
2. This error bubbles up through the generation loop in `generate()`, stopping the entire process 

But yeah, I'm not super happy with having to pass the signal through so many layers. I was trying to follow the pattern used by the Fetch API where you need to get the abort signal all the way down to the actual async operation being aborted. Would be cool if in the future the `InferenceSession` could support the abort signal internally.

But maybe there's a cleaner way to structure this. Like maybe we could store the signal at the pipeline level instead of threading it through all those args. Let me know if you have any other ideas for improving it.

emojiiii:
It seems like you're looking for https://github.com/huggingface/transformers.js/blob/main/src/models.js#L1573,  
you can refer to https://github.com/huggingface/transformers.js-examples/blob/main/deepseek-r1-webgpu/src/worker.js#L114.

joecal:
> It seems like you're looking for https://github.com/huggingface/transformers.js/blob/main/src/models.js#L1573, you can refer to https://github.com/huggingface/transformers.js-examples/blob/main/deepseek-r1-webgpu/src/worker.js#L114.

I've been using the `EosTokenCriteria` and found it works great at stopping text generation at known specific token outputs. However, when I encountered the funny-looking text output issue in the video I posted above and then saw these abort signal PR's, it got me thinking: it would be cool if we could use an abort signal to stop any long-running async task in the entire pipeline, whether the task is downloading with fetch, token generation, or any other time-consuming asynchronous task. What do you guys think?

sroussey:
I updated the API to make sure abort_signal is never a naked parameter. Instead it is always part of an options object so others can be added later. Also set default value to null instead of undefined just so it doesn't stick out in the codebase as different.

sroussey:
TODO: at least for `pipeline()`, I will create an `InterruptableStoppingCriteria` that will activate when the signal activates. I'll add it if no other stopping criteria, and convert any other to a list version and add this one to it if there is an abort_signal passed. Thanks for the headsup @emojiiii and @joecal

sroussey:
I have a version of transformers.js with abort signal handling at @sroussey/transformers

xenova:
Hi 👋 I just saw your comment in https://github.com/huggingface/transformers.js/pull/1465#issuecomment-3583945702, and looking at this PR in a bit more detail, what do you think about a possibly simpler implementation which allows the user to specify their own fetch function? It's something we support in the `@huggingface/hub` library, and could make sense here.

For example, a user could do something like this:
```js

let globalAbortController = new AbortController();
function myFetch(input, init = {}) {
  globalAbortController.abort(); // Abort any previous request that used this controller
  globalAbortController = new AbortController(); // Create a new controller for the next use

  const mergedInit = {
    ...init,
    signal: globalAbortController.signal,
  };

  return fetch(input, mergedInit);
}

env.fetch = myFetch;
```

and I think this will have a similar effect as your proposed PR. WDYT @sroussey ? :)

sroussey:
You need more than fetch. You need fetch, ReadableStream, and cache after a quick glance. I think with those three you can hack together a global abort. Though if you wanted a UI to stop some models from downloading and not others, it would be very difficult to map these on the user side for what call resulted in what fetch, readablestream, or cache.

PR #1196, opened 2025-02-16
Add support for SmolVLM2

xenova:
See https://github.com/huggingface/transformers/pull/36126.

TODO:
- Improve video processing (e.g., from `<video>` element or stream).


xenova:
The functionality in video.js is a good start, but fully integrating this into the processor would need a bit more thinking (since video decoding is currently only available in the main thread with access to `document). Will merge for now, and will work on improvements in future. If you have any thoughts, feel free to let me know!

PR #1197, opened 2025-02-18
[WIP] Translate JS to TS

sroussey:


kungfooman:
image

Did it ever cross your mind that you just make beautiful JS unreadable with no gains and forcing everyone to install endless amounts of node_modules for... a PITA? Please learn modern JS, ESM and importmaps.

BritishWerewolf:
This has been discussed in #1145 where the consensus was the same: TypeScript doesn't provide any additional benefits than what the library already does.  
Types exist already.

Issue #960, opened 2024-10-04
[V3 proposal] Improved defaults for quantization and device selection


xenova:
### Feature request

Currently, Transformers.js V3 defaults to use CPU (WASM) instead of GPU (WebGPU) due to lack of support and instability across browsers (specifically Firefox and Safari, and Chrome in Ubuntu). However, this provides a poor user experience since is performance left on the table. As browser support for WebGPU increases (currently [~70%](https://caniuse.com/webgpu)), this will become more important since users may experience poor performance when better settings are available.

A better proposal should be to use `device: "auto"` instead of `device: null` by default, which should select (1) quantization and (2) device) based on the following:
1. Browser support (e.g., whether WebGPU is enabled)
1. Device capabilities (OS, mobile vs. desktop, fp16 support)
1. Model architecture/type (BERT models are more likely to succeed than encoder-decoder models) - some models have ops which are not supported in WebGPU.

### Motivation

Improve user experience and performance with better defaults

### Your contribution

Will work with @FL33TW00D on this

FL33TW00D:
![auto_device](https://github.com/user-attachments/assets/189df3db-a1ac-459b-96b2-672912e309bb)

xenova:
Current logic for session selection: https://github.com/xenova/transformers.js/blob/6505abb164a3eea1dd5e80e56a72f7d805715f0a/src/models.js#L148-L262

FL33TW00D:
Some thoughts:
1. We could create a `class Device` and keep the enum `DEVICE_TYPES` [here](https://github.com/xenova/transformers.js/blob/6505abb164a3eea1dd5e80e56a72f7d805715f0a/src/utils/devices.js#L5)
2. Next we convert [this](https://github.com/xenova/transformers.js/blob/6505abb164a3eea1dd5e80e56a72f7d805715f0a/src/backends/onnx.js#L38) to a class `ORTBackend` or something. We then encapsulate all of the logic in the struct to struct conversion from a `Transformers.js` device to an `ORTBackend`. This separation of concerns should work out quite nicely.
3. Users can then only worry about providing `Device.CPU`, `Device.GPU` and `Device.NPU` and we handle the rest!

webmaxru:
Congrats with v3 release! I have a couple of questions about WebNN

[...]

PR #1199, opened 2025-02-19
refactor: deprecate onnx.env.webgpu.powerPreference

Hengry:
Onnxruntime-web deprecated `powerPreference` in WebGPUFlag.

Instead, they suggested
> Create your own GPUAdapter, use it to create a GPUDevice instance and set [device](https://onnxruntime.ai/docs/api/js/interfaces/Env.WebGpuFlags.html#device) property if you want to use a specific power preference.

Reference:
https://onnxruntime.ai/docs/api/js/interfaces/Env.WebGpuFlags.html#powerPreference:~:text=Create%20your%20own%20GPUAdapter%2C%20use%20it%20to%20create%20a%20GPUDevice%20instance%20and%20set%20device%20property%20if%20you%20want%20to%20use%20a%20specific%20power%20preference.

### DIFF
--- a/src/backends/onnx.js
+++ b/src/backends/onnx.js
@@ -152,7 +152,12 @@ export async function createInferenceSession(buffer, session_options, session_co
         // so we wait for it to resolve before creating this new session.
         await wasmInitPromise;
     }
-
+    if (ONNX_ENV?.webgpu) {
+        const adapter = await navigator.gpu.requestAdapter({ powerPreference: 'high-performance' })
+        if (adapter) {
+            ONNX_ENV.webgpu.device = await adapter.requestDevice();
+        }
+    }
     const sessionPromise = InferenceSession.create(buffer, session_options);
     wasmInitPromise ??= sessionPromise;
     const session = await sessionPromise;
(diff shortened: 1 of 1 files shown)


fs-eire:
Creating a custom device is supported in onnxruntime-web, but there are also several things to be careful with:

- onnxruntime-web calls [`adapter.requestDevice(params)`](https://developer.mozilla.org/en-US/docs/Web/API/GPUAdapter/requestDevice) with specific `requiredFeatures` and `requiredLimits` in `params`. Creating the device using different params will not be guaranteed to work.
- there are slightly different behavior between different versions for the implementation of dealing with `env.webgpu.device` and `env.webgpu.adapter`. the stable implementation is planned for onnxruntime-web 1.22.0 release.

And regarding the "powerPreference" option, it actually makes little difference:
- This option is currently not working in Chrome on Windows: https://issues.chromium.org/issues/369219127
- From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/GPU/requestAdapter):
  > Note: On Chrome running on dual-GPU macOS devices, if requestAdapter() is called without a powerPreference option, the high-performance discrete GPU is returned when the user's device is on AC power. Otherwise, the low-power integrated GPU is returned.

  So, the difference between default and `high-performance` only happens on Macbook with Intel CPU when it's on battery. (seems that is the only macOS device that both has multiple GPU and can run on battery - please correct me if I was wrong)

I am also trying to figure out if setting `powerPreference` to "low-power" is possibly a real requirement for onnxruntime-web. If not, probably the easiest way is to always set it to "high-performance" in onnxruntime-web.

xenova:
This can be linked to #960, which we can now work on thanks to onnxruntime-web exposing the device (as you point out in this PR).

PR #1208, opened 2025-02-25
Use Float16Array instead of Uint16Array for kvcache when available

xenova:
Due to limited availability of [Float16Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Float16Array) across browsers, we defaulted to Uint16Array to store data for fp16 tensors. However, when it is available (e.g., Firefox) it previously threw an error:
image

Reported on discord: https://discord.com/channels/1089876418936180786/1342161413254287452/1342191027594854451

cc @tarekziade @daavoo

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -108,6 +108,7 @@ import {
     stack,
     std_mean,
     Tensor,
+    DataTypeMap,
 } from './utils/tensor.js';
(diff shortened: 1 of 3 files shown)


xenova:
Linked: https://github.com/microsoft/onnxruntime/issues/23817

PR #1213, opened 2025-02-28
fix extension example with 3.4.0

fs-eire:
This fixes the extension example.

The code does not work yet, because it depends on transformerjs v3.4.0, which is not released yet. It needs #1161 to work.

### DIFF
--- a/examples/extension/package.json
+++ b/examples/extension/package.json
@@ -15,6 +15,6 @@
     "webpack": "^5.79.0"
   },
   "dependencies": {
-    "@xenova/transformers": "^2.0.0"
+    "@huggingface/transformers": "^3.4.0"
   }
 }
--- a/examples/extension/src/background.js
+++ b/examples/extension/src/background.js
@@ -1,6 +1,6 @@
 // background.js - Handles requests from the UI, runs the model, then sends back a response
 
-import { pipeline, env } from '@xenova/transformers';
+import { pipeline, env } from '@huggingface/transformers';
 
 // Skip initial check for local models, since we are not loading any local models.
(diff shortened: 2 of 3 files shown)


xenova:
While we're migrating all examples to http://github.com/huggingface/transformers.js-examples, we might as well merge this here now that Transformers.js v3.4.0 is out!

Thanks @fs-eire.

mattmacf98:
@xenova this works great when importing @hugginface/transformers in a background script! In my use case I wanted to run the script in the popup.js (which doesn't respect the` import-scripts` parameter for chunk loading) it will still try to load from the CDN.

below is how I modified the popup.js to break with @huggingface/transformers 3.4.0 (but it works with @xenova/transformers ^2.0.0)
```

class PipelineSingleton {
    static task = 'text-classification';
    static model = 'Xenova/distilbert-base-uncased-finetuned-sst-2-english';
    static instance = null;

    static async getInstance(progress_callback = null) {
        this.instance ??= pipeline(this.task, this.model, { progress_callback });

        return this.instance;
    }
}

const inputElement = document.getElementById('text');
const outputElement = document.getElementById('output');

// Listen for changes made to the textbox.
inputElement.addEventListener('input', async (event) => {

    // Bundle the input data into a message.
    const message = {
        action: 'classify',
        text: event.target.value,
    }

    let model = await PipelineSingleton.getInstance();
    let result = await model(message.text);
    outputElement.innerText = JSON.stringify(result, null, 2);
});
```

if the recommendation is just to use background scripts for working with the huggingface library and utilize the chrome messaging API to take acton from the popup, that is totally valid and I can just update my project.

PR #1216, opened 2025-03-03
Add background removal pipeline

xenova:
This PR makes it far easier to use background-removal models.

```js
import { pipeline } from '@huggingface/transformers';

const segmenter = await pipeline('background-removal', 'onnx-community/BEN2-ONNX');
const url = 'https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/woman-with-afro_medium.jpg';
const output = await segmenter(url);
output[0].save('mask.png');
```

| Input| Output |
|--------|--------|
| input | output |

Tested with:
- ✅ https://huggingface.co/Xenova/modnet
- ✅ https://huggingface.co/briaai/RMBG-1.4
- ✅ https://huggingface.co/briaai/RMBG-2.0
- ✅ https://huggingface.co/onnx-community/mediapipe_selfie_segmentation
- ✅ https://huggingface.co/onnx-community/mediapipe_selfie_segmentation_landscape
- ✅ https://huggingface.co/onnx-community/modnet-webnn
- ✅ https://huggingface.co/onnx-community/BiRefNet-ONNX
- ✅ https://huggingface.co/onnx-community/BiRefNet_lite-ONNX
- ✅ https://huggingface.co/onnx-community/BiRefNet-portrait-ONNX
- ✅ https://huggingface.co/onnx-community/BiRefNet-COD-ONNX
- ✅ https://huggingface.co/onnx-community/BiRefNet-DIS5K-TR_TEs-ONNX
- ✅ https://huggingface.co/onnx-community/BiRefNet-DIS5K-ONNX
- ✅ https://huggingface.co/onnx-community/BiRefNet-HRSOD_DHU-ONNX
- ✅ https://huggingface.co/onnx-community/BiRefNet_512x512-ONNX
- ✅ https://huggingface.co/onnx-community/ormbg-ONNX
- ✅ https://huggingface.co/onnx-community/BEN2-ONNX
- ✅ https://huggingface.co/onnx-community/MVANet-ONNX
- ✅ https://huggingface.co/onnx-community/ISNet-ONNX

TODO:
- [ ] Update model cards

PR #1222, opened 2025-03-06
[FEAT] Add U-2-Net

BritishWerewolf:
# Prerequisites
* #1101
* #1103
* #1136

These PRs are required because the processors lean heavily on the new features.

---

Add support for the U-2-Net architecture which is a mask generation model; useful for background removal.  
Both U-2-Net and ISNet can be used with these processors.

Below is an example of what the model will achieve when using the [BritishWerewolf/U-2-Netp](https://huggingface.co/BritishWerewolf/U-2-Netp) model.

| Model | Original | Masked |
|-------|----------|--------|
| U-2-Netp | A photo of a Loungefly backpack with a Moana design on it. The background is a single shade of white. | The same photo of the Loungefly backpack, however this time the white background has been removed and become transparent. The masking isn't perfect, and there is still a white halo around the image. |
| IS-Net | A photo of a Loungefly backpack with a Moana design on it. The background is a single shade of white. | The same photo of the Loungefly backpack, however this time the white background has been removed and become transparent. The masking isn't perfect, though better than U-2-Netp, and there is still a white halo around the image. |

If you would like to run this code, you can do with the [BritishWerewolf/U-2-Net collection of models](https://huggingface.co/collections/BritishWerewolf/u-2-net-67c64b670602b2949573234e).
```js
// Create the processor.
const processor = await AutoProcessor.from_pretrained(modelName);

// Create the model.
const model = await AutoModel.from_pretrained(modelName, {
	dtype: 'fp32',
});

// Process the image.
const image = await RawImage.read('https://huggingface.co/ybelkada/segment-anything/resolve/main/assets/car.png');
const processed_image = await processor(image);

// Run the model.
const output = await model({
	input: processed_image.pixel_values,
});

// Retrieve the mask, and scale it up to the size of the original image.
const mask = await RawImage.fromTensor(output.mask)
	.resize(image.width, image.height);

// Apply the mask to the image and save it.
const maskedImage = await image.putAlpha(mask);
maskedImage.save('masked_image.png');
```


xenova:
Hi @BritishWerewolf 👋 Thanks for the PR!

I would imagine that the additional pre-processing needed introduces quite a large latency to model inference, especially because the necessary operations are implemented in JS as opposed to WASM/WebGPU. Is that correct?

Do you see any benefits of a model like this over a model listed in the recent background-removal pipeline update? https://github.com/huggingface/transformers.js/pull/1216

BritishWerewolf:
Hey @xenova, I started work on this before the pipeline, but kept this PR as draft because I hoped to fix it down the line.
My current laptop just became too cumbersome to work with, because each iteration I was making just started to take a few minutes and it started adding up.

If you’re happy to leave this open as draft, then I will have a look into it when I buy a new laptop; but if you want to clean up the PRs and close this I completely understand (and this may be the better approach, I can reopen in future _if needed_).

Thanks

PR #1231, opened 2025-03-13
ONNX Runtime improvements (experimental native webgpu; fix iOS)

fs-eire:
This change allows using WebGPU in transformers.js with ORT Node.js binding.

Still doing testing (while the tests need this change)

Closes https://github.com/huggingface/transformers.js/issues/1242

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@huggingface/jinja": "^0.3.4",
-        "onnxruntime-node": "1.21.0",
+        "onnxruntime-node": "1.22.0-dev.20250418-c19a49615b",
         "onnxruntime-web": "1.22.0-dev.20250409-89f8206ba4",
         "sharp": "^0.34.1"
       },
@@ -1009,6 +1009,7 @@
       "version": "8.0.2",
       "resolved": "https://registry.npmjs.org/@isaacs/cliui/-/cliui-8.0.2.tgz",
       "integrity": "sha512-O8jcjabXaleOG9DQ0+ARXWZBTfnP4WNAqzuiJK7ll44AmxGKv/J2M4TPjxjY3znBCfvBXFzucm1twdyFybFqEA==",
+      "dev": true,
       "dependencies": {
         "string-width": "^5.1.2",
         "string-width-cjs": "npm:string-width@^4.2.0",
@@ -1025,6 +1026,7 @@
       "version": "6.0.1",
       "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-6.0.1.tgz",
       "integrity": "sha512-n5M855fKb2SsfMIiFFoVrABHJC8QtHwVx+mHWP3QcEqBHYienj5dHSgjbxtC0WEZXYt4wcD6zrQElDPhFuZgfA==",
+      "dev": true,
       "engines": {
         "node": ">=12"
       },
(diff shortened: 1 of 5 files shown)


AdamStrojek:
Wouldn't it be better to do the same thing as it is done in Onnx Runtime Web?
```
    if (apis.IS_WEBGPU_AVAILABLE) {
    }
```
Electron applications can have WebGPU enabled when terminal Node not. Also `onnx-runtime-node` provides only backers for native modules, when `onnx-runtime-web` have bindings for WebGPU, so just adding supported devices will not work without switching runtime

fs-eire:
If I remember it correctly, IS_WEBGPU_AVAILABLE is checked against nagivator.gpu, which is only available in browser.

For electron, the rendering process is actually a "web" environment instead of "node"

AdamStrojek:
Yes, you are correct, `IS_WEBGPU_AVAILABLE` is just a simple check against `navigation.gpu`. In theory, it is possible to install a 3rd-party package for WebGPU support in Node, but it is a complicated topic. Still, my comment is valid; I copied my example from a few lines higher in the same source file.

I recently did tests. Unfortunately, transformers.js are not detecting Electron applications correctly and mark them as Node applications, so it provides only CPU. I had a lot of trouble getting it running in an Electron app. Mostly, it was picky about `path` and `fs` packages. If I changed the target platform to Node, it generated other problems. I'm preparing a new issue report for developers with my findings.

I already did tests with your branch, and this simple change didn’t enable WebGPU in Electron apps.

fs-eire:
Updated the version of `onnxruntime-node` to 1.22.0-dev.20250418-c19a49615b. This version supports WebGPU on Windows and macOS.

xenova:
Wow thanks @fs-eire! Very exciting!!!

fs-eire:
> Wow thanks @fs-eire! Very exciting!!! Does the browser package https://www.npmjs.com/package/onnxruntime-web/v/1.22.0-dev.20250418-c19a49615b release also add anything of significance?

No.

BTW for WebGPU EP support in onnxruntime-web : There are still some perf issue for using WebGPU EP in a WebAssembly build. If you want to do conformance test only for WebGPU EP (eg. check correctness but not latency), I can offer you a private build of onnxruntime-web with WebGPU EP.

xenova:
That would be great! Feel free to send via slack perhaps? Eventually, we can hook this into the Transformers.js CI to ensure correctness across all supported architectures.

xenova:
I've been testing the webgpu EP for some llama/qwen models, and running into a few correctness issues.

Here's some code to help test/debug:
```

  "text-generation",
  "onnx-community/ZR1-1.5B-ONNX",
  { dtype: "q4f16", device: "webgpu" }, // device="cpu" works fine
);

const messages = [
  { role: "user", content: "Write me a poem about Machine Learning." },
];

// Generate a response
    max_new_tokens: 512,
    do_sample: false,
    streamer: new TextStreamer(generator.tokenizer, { skip_prompt: true, skip_special_tokens: true}),
});
```

xenova:
I can confirm that q4 (instead of q4f16) works correctly, so it looks to be an issue with the f16 implementation.

guschmue:
looks like the same issue as deepseek when GQA uses FA2 with fp16. fp32 seems ok.
I'll put this high on my list to look at.

xenova:
Great, thanks @guschmue!

xenova:
I'm accumulating all these changes into https://github.com/huggingface/transformers.js/tree/ort-improvements to make development and testing a bit easier (many version bumps and ort-specific changes)

PR #1244, opened 2025-03-18
[FEAT] Allow RawImage to read from all types of supported sources.

BritishWerewolf:
A simple PR that allows the `read` method to convert from any supported type.

This will be useful for the pipeline API which uses the `RawImage.read` method call under the hood.  
Currently `read` just interprets URLs, but this will allow File Uploads for instance.

```js
// File upload
const file = event.target.files[0];
const image = await RawImage.read(file);

// Canvas
const canvas = document.querySelector('canvas');
const image = await RawImage.read(canvas);
```

### DIFF
--- a/src/pipelines.js
+++ b/src/pipelines.js
@@ -2301,7 +2301,7 @@ export class ImageSegmentationPipeline extends (/** @type {new (options: ImagePi
  * // ]
  * ```
  */
-export class BackgroundRemovalPipeline extends (/** @type {new (options: ImagePipelineConstructorArgs) => ImageSegmentationPipelineType} */ (ImageSegmentationPipeline)) {
+export class BackgroundRemovalPipeline extends (/** @type {new (options: ImagePipelineConstructorArgs) => BackgroundRemovalPipelineType} */ (/** @type {any} */(ImageSegmentationPipeline))) {
     /**
      * Create a new BackgroundRemovalPipeline.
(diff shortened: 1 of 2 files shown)


xenova (changes requested):
This is pretty useful, and is actually an update I wanted to make for a long time! Thanks! Can you also make sure the relevant pipelines also expose that blobs and canvas elements are now supported?

src/pipelines.js
```diff
@@ -2277,20 +2277,20 @@ export class ImageSegmentationPipeline extends (/** @type {new (options: ImagePi
 
 
 /**
- * @typedef {Object} BackgroundRemovalPipelineOptions Parameters specific to image segmentation pipelines.
- * 
+ * @typedef {ImageSegmentationPipelineOptions} BackgroundRemovalPipelineOptions Parameters specific to image segmentation pipelines.
```
BritishWerewolf, line 2280: I've also updated the `BackgroundRemovalPipelineOptions` to match `ImageSegmentationPipelineOptions` for now, seeing as it just extends that.  
This can always be changed in future if the options change.

xenova: I intentionally removed all these options since they aren't applicable to the background removal pipeline (for now, at least). So, I think it's okay to keep this empty for now.

xenova (approved):
Thanks!

Issue #1242, opened 2025-03-17
🐛 v3 crashes on iOS and macOS devices due to increasing memory usage


GaLambert:
### System Info

- transformers.js version: 3.2.2
- iOS: 18.3.2
- macOS: 15.3.1 
- Framework: Next.js (14.2.22) and Vite (6.0.5)

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension
- [ ] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

We've encountered issues with the v3 of transformers.js (specifically v3.2.2, potentially others) on iOS devices and macOS. The application crashes on iOS (both Safari and Chrome), and causes extremely high and growing memory usage (10+GB) on macOS, eventually also crashing.

Tested with the following examples:

- https://github.com/huggingface/transformers.js-examples/tree/main/next-client
- https://github.com/huggingface/transformers.js-examples/tree/main/video-object-detection

**Temporary solution**: As a workaround, downgrading to @xenova/transformers@2.15.1 resolves the iOS crashing issue. However, this is not an ideal long-term solution.

We would greatly appreciate any insights or assistance in resolving these issues  with the latest version. Thank you! 🙏

### Reproduction

1. Use transformers.js version 3.2.2
2. Run a model (e.g., YoloV9) on an iOS device (iPhone)
3. Observe that the application crashes
4. Run the same model on macOS
5. Check Activity Monitor and observe memory usage exceeding 10GB+ for Safari

xenova:
Thanks for the report. Could you try again with the latest version to see if this is still an issue? Currently v3.4.0

cc @fs-eire @guschmue too

Jerboas86:
> Thanks for the report. Could you try again with the latest version to see if this is still an issue? Currently v3.4.0

the bug is still there in v3.4.0

GaLambert:
After testing, I confirm that the bug persists in transformers.js v3.4.0.

fs-eire:
This may be caused by JSC not doing well with multi-threaded WASM. Could you help to check when you use `@xenova/transformers@2.15.1` and set numThreads what will the memory usage be like?

Jerboas86:
With @huggingface/transformers v3.4 on ipadOs 18.3.2, using:

`env.backends.onnx.wasm.numThreads = N;`

- N = 1 -> Crash / 3.5Go memory usage
[...]

PR #1250, opened 2025-03-21
customize the wasm paths

fs-eire:
This PR updates the logics to have a better customized `wasmPaths` setting. Specifically, the following changes were made:

- (Safari) Fallback to non ASYNCIFY build of webassembly, because of an OOM issue: #1242 
  - This fallback does not affect any existing feature, because WebGPU is not available on Safari yet anyway

- Because of the fallback feature, we need to access file ort-wasm-simd-threaded.mjs and ort-wasm-simd-threaded.wasm, which is not in package `@huggingface/transformers`. point to `onnxruntime-web`.

- only apply override to `wasmPaths` when `env.backends.onnx.versions.web` exists. The only situation that it does not exist is the nodejs build.

The feature need to be well tested.

### DIFF
--- a/src/backends/onnx.js
+++ b/src/backends/onnx.js
@@ -180,9 +180,16 @@ if (ONNX_ENV?.wasm) {
     if (
         // @ts-ignore Cannot find name 'ServiceWorkerGlobalScope'.ts(2304)
         !(typeof ServiceWorkerGlobalScope !== 'undefined' && self instanceof ServiceWorkerGlobalScope)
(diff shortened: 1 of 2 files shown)


xenova:
Thanks for this! I'll do some additional testing, and hopefully have this included with https://github.com/huggingface/transformers.js/pull/1231.

xenova:
Superseded by https://github.com/huggingface/transformers.js/pull/1231

Issue #1259, opened 2025-03-27
3.2.4 has wrong env check in transformers.web.js


sanixa:
## Background
I have developed a chrome extension which is followed by the [example](https://github.com/huggingface/transformers.js/tree/main/examples/extension). The example was used the package @xenova/transformers.

## Motivation
It seems that multithreads is work now. [Issue](https://github.com/huggingface/transformers.js/issues/928) [Issue2](https://github.com/huggingface/transformers.js/issues/882)

## Question
I change the package from **@xenova/transformers@2.17.2** to **@huggingface/transformers@3.4.1**. It shows a error **TypeError: sharp__WEBPACK_IMPORTED_MODULE_4__ is not a function** which have no been shown before. Anyone can help?

## Code (background.js)
``` 
// import { pipeline, env } from '@xenova/transformers';
// env.localModelPath = './';
// env.allowRemoteModels = false;
// env.backends.onnx.wasm.numThreads = 1;

import { env, pipeline } from '@huggingface/transformers';
env.localModelPath = './';

class ImagePipelineSingleton {
    static task = 'image-classification';
    static model = '/deepfake/';
    static instance = null;

    static async getInstance() {
        try {
            if (this.instance === null) {
                this.instance = await pipeline(this.task, this.model);
            }
        } catch (error) {
            console.error("Initialization error:", error);
        }
        return this.instance;
    }
}

...
try{
    let model = await ImagePipelineSingleton.getInstance();
    let classification = await model(url); 
}catch (error) {
    console.error("image processing error:", error); //error here
}
...
```

## Folder Structure
- deepfake
  - onnx
    - model_quantized.onnx

sanixa:
It seems that  transformers.web.js:30693 has wrong env check in my project.

```
_env_js__WEBPACK_IMPORTED_MODULE_2__.apis.IS_BROWSER_ENV false
_env_js__WEBPACK_IMPORTED_MODULE_2__.apis.IS_WEBWORKER_ENV false
```

maybe need to fix it

PR #1267, opened 2025-04-01
fix: hub.js Missing "else" leading to local model loading error

### LINKED ISSUE #1253: 3.4.0 version of transformer does not respect env.localPath
### System Info

The version 3.4.0 breaks the use local model. The version 3.3.3 works perfectly.

I have a path like 

import {env, pipeline } from '@huggingface/transformer";

env.localModelPath = '../model-dir';

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension

### Description

The version 3.4.0 breaks the use local model. The version 3.3.3 works perfectly.

I have a path like 

import {env, pipeline } from '@huggingface/transformer";

env.localModelPath = '../model-dir';

### Reproduction

The version 3.4.0 breaks the use local model. The version 3.3.3 works perfectly.

I have a path like 

import {env, pipeline } from '@huggingface/transformer";

env.localModelPath = '../model-dir';

env.allowRemoteModels = false;
   const model = await pipeline(
      'feature-extraction',
      'bge-base-en',
   );

johnhorsema:
Missing else statement causing inference error with local models. Fixes #1253 and #1259

### DIFF
--- a/src/utils/hub.js
+++ b/src/utils/hub.js
@@ -503,7 +503,7 @@ export async function getModelFile(path_or_repo_id, filename, fatal = true, opti
             }
         }
 
-        if (response === undefined || response.status === 404) {
+        else if (response === undefined || response.status === 404) {
             // File not found locally. This means either:
             // - The user has disabled local file access (`env.allowLocalModels=false`)
             // - the path is a valid HTTP url (`response === undefined`)


xenova:
Hi there :wave: This is actually intended, since that block is only there for error checking after the file could not be found locally.

https://github.com/huggingface/transformers.js/pull/1268 should fix the original issue.

PR #1271, opened 2025-04-03
Fix error when dtype in config is unset

hans00:


hans00:
Some model config have not set dtype.
If specific `dtype: auto` on init, will got error.

xenova (approved):
Thanks! LGTM

PR #1280, opened 2025-04-11
Return buffer instead of file_path if cache unavailable for model loading

### LINKED ISSUE #1249: customCache `match` method maybe get undefined input
### System Info

electron 32.2.7
windows 10 19045.4717

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension
- [x] Desktop app (e.g., Electron)

### Description

```ts
export class ProxyCache {

  constructor(

  ) {

  }
  async match(request: string): Promise<FileResponse | undefined> {
// request maybe undefined in model request
}}
env.customCache=new ProxyCache() 
```
in `3.3.3` worked in `3.4.0` will receive undefined
not all undefined, But it's an additional undefined call
### Reproduction

1. use customCache
2. request `Xenova/bge-base-zh-v1.5`
3. match method receive undefined and throw error

### LINKED ISSUE #1279: Missing check for caching availability when getting model files
When running in a Node.js environment, it looks like there's an edge case where the unavailability of a caching system causes an error when loading model files. [The problematic code](https://github.com/huggingface/transformers.js/blob/10c09fb561857abf817d651a02898f590f5b7954/src/utils/hub.js#L640):

```js
// https://github.com/huggingface/transformers.js/blob/10c09fb561857abf817d651a02898f590f5b7954/src/utils/hub.js#L650

const path = await cache.match(cacheKey);
if (path instanceof FileResponse) {
    return path.filePath;
}
```

Specifically, in a Node.js environment, `getModelFile()` attempts to download the respective model.onnx (or equivalent) file, store it in a cache and return the path to it. In cases where caching is disabled or unavailable and a remote model is to be loaded, `cache` stays undefined and the `cache.match()` call throws an error.

Simple repro (just setting `useFSCache` to be `false`):
```js
import * as transformers from '@huggingface/transformers';

transformers.env.useFSCache = false;

transformers.pipeline("sentiment-analysis").then(async pipe => {
    console.log( await pipe('Doesn\'t work when cache is unavailable!') )
})
```
raises
`TypeError: Cannot read properties of undefined (reading 'match')`.

File system availability for caching might be an okay assumption to make here, but I feel like it should still be checked for. The function does make allowances for it in most places, it's just this last piece that seems to miss it.

As a possible suggestion, the `response` variable already contains the downloaded model file, so in cases where caching is unavailable, it might make sense to just pass the buffer back. Seems like it's passed directly to `ONNX.InferenceSession.create()`, which accepts both buffers and paths. Happy to raise a PR if that's acceptable.

PrafulB:
This PR resolves #1249 and #1279 . When running in a Node.js environment, there's an edge case where the unavailability of a caching system causes an error when loading model files. [The problematic code is here.](https://github.com/huggingface/transformers.js/blob/10c09fb561857abf817d651a02898f590f5b7954/src/utils/hub.js#L650). Can be verified by simply setting both `env.useFSCache` and `env.useCustomCache` to false.

Specifically, in a Node.js environment, `getModelFile()` attempts to download the respective model.onnx (or equivalent) file, store it in a cache and return the path to it. In cases where caching is disabled or unavailable and a remote model is to be loaded, `cache` stays undefined and the `cache.match()` call throws an error. Can be verified by simply setting both `env.useFSCache` and `env.useCustomCache` to false.

This PR makes it so that the file_path is only requested if the environment is Node.js AND either of the two caching options are available. Otherwise, the buffer containing the model file is returned as is.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -248,7 +248,7 @@ async function getSession(pretrained_model_name_or_path, fileName, options) {
         );
     }
 
-    const return_path = apis.IS_NODE_ENV && (env.useFSCache || env.useCustomCache);
+    const return_path = apis.IS_NODE_ENV && env.useFSCache;
     const bufferOrPathPromise = getModelFile(pretrained_model_name_or_path, modelFileName, true, options, return_path);
 
     // handle onnx external data files
--- a/src/models.js
+++ b/src/models.js
@@ -116,7 +116,7 @@ import { RawImage } from './utils/image.js';
 import { dynamic_time_warping, max, medianFilter } from './utils/maths.js';
 import { EosTokenCriteria, MaxLengthCriteria, StoppingCriteriaList } from './generation/stopping_criteria.js';
 import { LogitsSampler } from './generation/logits_sampler.js';
-import { apis } from './env.js';
+import { apis, env } from './env.js';
 
 import { WhisperGenerationConfig } from './models/whisper/generation_whisper.js';
 import { whisper_language_to_code } from './models/whisper/common_whisper.js';
@@ -248,7 +248,8 @@ async function getSession(pretrained_model_name_or_path, fileName, options) {
         );
     }
 
-    const bufferOrPathPromise = getModelFile(pretrained_model_name_or_path, modelFileName, true, options, apis.IS_NODE_ENV);
+    const return_path = apis.IS_NODE_ENV && (env.useFSCache || env.useCustomCache);
+    const bufferOrPathPromise = getModelFile(pretrained_model_name_or_path, modelFileName, true, options, return_path);
 
     // handle onnx external data files
     const use_external_data_format = options.use_external_data_format ?? custom_config.use_external_data_format;
@@ -276,7 +277,7 @@ async function getSession(pretrained_model_name_or_path, fileName, options) {
             const path = `${baseName}_data${i === 0 ? '' : '_' + i}`;
             const fullPath = `${options.subfolder ?? ''}/${path}`;
             externalDataPromises.push(new Promise(async (resolve, reject) => {
-                const data = await getModelFile(pretrained_model_name_or_path, fullPath, true, options, apis.IS_NODE_ENV);
+                const data = await getModelFile(pretrained_model_name_or_path, fullPath, true, options, return_path);
                 resolve(data instanceof Uint8Array ? { path, data } : path);
             }));
         }
--- a/src/models.js
+++ b/src/models.js
@@ -116,7 +116,7 @@ import { RawImage } from './utils/image.js';
 import { dynamic_time_warping, max, medianFilter } from './utils/maths.js';
 import { EosTokenCriteria, MaxLengthCriteria, StoppingCriteriaList } from './generation/stopping_criteria.js';
 import { LogitsSampler } from './generation/logits_sampler.js';
-import { apis, env } from './env.js';
+import { apis } from './env.js';
 
 import { WhisperGenerationConfig } from './models/whisper/generation_whisper.js';
 import { whisper_language_to_code } from './models/whisper/common_whisper.js';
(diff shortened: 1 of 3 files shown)


src/models.js
```diff
@@ -248,7 +248,7 @@ async function getSession(pretrained_model_name_or_path, fileName, options) {
         );
     }
 
-    const bufferOrPathPromise = getModelFile(pretrained_model_name_or_path, modelFileName, true, options, apis.IS_NODE_ENV);
+    const bufferOrPathPromise = getModelFile(pretrained_model_name_or_path, modelFileName, true, options, (apis.IS_NODE_ENV && (env.useFSCache || env.useCustomCache)));
```
xenova, line 251: Perhaps simpler to move this additional check to `getModelFile`?

src/utils/hub.js
```diff
@@ -638,7 +638,7 @@ export async function getModelFile(path_or_repo_id, filename, fatal = true, opti
     });
 
     if (result) {
-        if (return_path) {
+        if (!apis.IS_NODE_ENV && return_path) {
```
xenova, line 641: technically not necessary due to above checks, but adding it here again makes it more readable. Will keep 👍

src/models.js
PrafulB: I thought so too, but then `getModelFile` is called with `return_path` being true, meaning the function is expected to return a file path. Because that is not possible in the case where a cache is unavailable, `return_path` would need to be ignored and a buffer would be returned necessarily. Keeping the checks in the call to the function seemed like an okay way to avoid this scenario.

src/utils/hub.js
```diff
@@ -407,6 +407,7 @@ export async function getModelFile(path_or_repo_id, filename, fatal = true, opti
             throw Error("Invalid configuration detected: both local and remote models are disabled. Fix by setting `env.allowLocalModels` or `env.allowRemoteModels` to `true`.")
         }
     }
+    return_path &&= env.useFSCache || env.useCustomCache;
```
xenova, line 410: What is the reason for including `env.useCustomCache` here too?

xenova, line 410: Just need to make sure this doesn't break some use-case when `useCustomCache=true`

src/models.js
xenova: Yeah good point! Reverted my change :)

src/utils/hub.js
PrafulB: I think it actually is necessary here. Earlier, the code assumed that `result` would only have a value if we were in the browser. Cache was assumed to always be available in Node and the `response` variable would be populated from the cache, instead of `result`. This assumption was incorrect. The error inside this condition should only be thrown if `return_path` is true in a non-Node environment.

xenova:
@PrafulB I made some minor modifications to also allow for custom cache mechanisms to work with `Response` or `string` return values. i.e., the following cases now all work:
```js
import * as transformers from '@huggingface/transformers';

async function run() {
    const pipe = await transformers.pipeline(
        "sentiment-analysis",
        "Xenova/distilbert-base-uncased-finetuned-sst-2-english",
        { dtype: "fp32" }
    );
    const result = await pipe("I love transformers.js!");
    console.log(result);
}

// Default
await run();

// No file cache
transformers.env.useFSCache = false;
await run();
transformers.env.useFSCache = true;

// Custom cache that does nothing
class CustomCache {
    async put(key, value) {
        /* Simulate adding nothing to the cache (e.g., no more storage available) */
    }
    async match(key) {
        /* Simulate a cache miss */
    }
}

transformers.env.useCustomCache = true;
transformers.env.customCache = new CustomCache();
await run();
```

I'm hoping to include this fix in the next version (v3.5), so if possible, can you confirm these changes meet your requirements still? Thanks!

xenova (approved):

PrafulB:
The changes look good to me. If I could make one final suggestion though? In the case where both `useFSCache` and `useCustomCache` are set to true (perhaps because `useFSCache` is true by default and the user forgot to unset it), the code defaults to the FileCache instead of the user-defined custom cache (see [utils/hub.js L437](https://github.com/huggingface/transformers.js/blob/32bd34ae4ac0cdc1dcdc2eb6e742439f35bf21fd/src/utils/hub.js#L437)). It should be the other way around, no? 
To that point, the test code you posted does not actually use the CustomCache implementation because you reset `useFSCache` to true after the second `run()` 😛.

PrafulB:
Created PR #1285 to track this and keep it separate from this one.

PR #1281, opened 2025-04-11
Stream reading a Response object may not be available while a progress callback is specified

Xmader:
The `body` property may not be available on a Response object to do streaming, e.g., when using a [fetch polyfill](https://github.com/github/fetch).

### DIFF
--- a/src/utils/hub.js
+++ b/src/utils/hub.js
@@ -578,12 +578,17 @@ export async function getModelFile(path_or_repo_id, filename, fatal = true, opti
             buffer = new Uint8Array(await response.arrayBuffer());
 
         } else if (
-            cacheHit // The item is being read from the cache
-            &&
-            typeof navigator !== 'undefined' && /firefox/i.test(navigator.userAgent) // We are in Firefox
+            (
+                // Due to bug in Firefox, we cannot display progress when loading from cache.
+                // Fortunately, since this should be instantaneous, this should not impact users too much.
+                cacheHit // The item is being read from the cache
+                &&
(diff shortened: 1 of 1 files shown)


xenova:
Thanks for the PR! Very interesting. 👀 Do you have a link to some documentation which states that the response body may be undefined in that (or some other) polyfill?

Xmader:
Hi @xenova, Thanks for the reply!
I couldn't find anywhere in the documentation of that polyfill, but I found a few issues (https://github.com/JakeChampion/fetch/issues/1454, https://github.com/JakeChampion/fetch/issues/746, https://github.com/JakeChampion/fetch/issues/198#issuecomment-494361060)  saying  `Response.body` is undefined and would unlikely be supported ever.

---

Another polyfill https://github.com/developit/unfetch says [it only supports a subset](https://github.com/developit/unfetch#response-methods-and-attributes) of the Response class, `response.body` is not included.

xenova:
Thanks for the info 👍 And just for interest's sake, what environment are you running in that requires a fetch polyfill?

lindboe:
> And just for interest's sake, what environment are you running in that requires a fetch polyfill?

@xenova Most React Native apps use a fetch polyfill. (Some use expo/fetch which is a native module implementation.)

I am not using this library, just came across this while researching some changes to the fetch polyfill, so I figured I would answer.

xenova:
As of Transformers.js v4, we expose a custom `env.fetch` option, which you can use if you need to support this.

PR #1285, opened 2025-04-15
Use custom cache over FSCache if specified

PrafulB:
Currently, if both `useFSCache` and `useCustomCache` env options are set to true (perhaps because `useFSCache` is true by default and the user forgot to unset it), the code defaults to instantiating a FileCache instead of using the custom cache instance passed in by the user when getting model files. This commit changes the default to be the custom cache if specified.

### DIFF
--- a/src/utils/hub.js
+++ b/src/utils/hub.js
@@ -451,7 +451,9 @@ export async function getModelFile(path_or_repo_id, filename, fatal = true, opti
     }
 
     if (!cache && env.useFSCache) {
-        // TODO throw error if not available
+        if (!apis.IS_FS_AVAILABLE) {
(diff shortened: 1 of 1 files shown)


xenova:
Thanks! Should we maybe take this one step further and prefer this over browser cache too? (otherwise, similarly, the user would need to disable browser cache before enabling custom cache)

PrafulB:
Of course! Totally missed that. 👍🏼 👍🏼

PrafulB:
Possibly outside the scope of this PR, but saw a TODO about throwing an error on fs cache being unavailable, so added that in my last commit. Can be excluded if needed.

xenova (approved):
Thanks again!

PR #1288, opened 2025-04-16
fix: `RawImage.fromURL` error when input file url

himself65:
Upstream: https://github.com/run-llama/LlamaIndexTS/pull/1855

`fs.existSync(path)` will assume path is a posix/win32 format if you give a string, so you will need pass `url.pathname`

### DIFF
--- a/src/utils/hub.js
+++ b/src/utils/hub.js
@@ -1,7 +1,7 @@
 
 /**
  * @file Utility functions to interact with the Hugging Face Hub (https://huggingface.co/models)
- * 
+ *
  * @module utils/hub
  */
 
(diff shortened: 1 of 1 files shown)


himself65:
> Hi @himself65 👋 Can you give an example of a URL which has this problem? The if clause should not allow invalid URLs to pass this step.

Yes, we were using `file:` protocol in node.js, and file protocol is a standard in node.js. And this is a regression because our test is working in <=4.0.0

```ts
const imageUrl = new URL(
  "../../fixtures/img/llamaindex-white.png",
  import.meta.url,
);

const nodes = [
  new ImageNode({
    image: new URL(
      "../../fixtures/img/llamaindex-white.png",
      import.meta.url,
    ),
    image: imageUrl,
  }),
];
```

xenova:
That's a valid use-case, thanks for reporting! This will be merged in v3.5.1 💪

src/utils/hub.js
```diff
@@ -215,8 +215,7 @@ function isValidHfModelId(string) {
 export async function getFile(urlOrPath) {
 
     if (env.useFS && !isValidUrl(urlOrPath, ['http:', 'https:', 'blob:'])) {
-        return new FileResponse(urlOrPath.toString());
-
+        return new FileResponse(urlOrPath instanceof URL ? urlOrPath.pathname : urlOrPath);
```
xenova, line 218: Should we perhaps check that the URL uses a valid `file://` schema? Is there any other case where `.pathname` would be valid (on a non-file URL?)

himself65: update the code

xenova (approved):
Thanks!

src/utils/hub.js
xenova: Thanks!

PR #1293, opened 2025-04-23
WebGPU fixes

xenova:
Makes multi-session management much easier, adding a global web inference lock to ensure only one session can run at any given time, preventing "Session has already started" errors.


src/models.js
```diff
@@ -433,17 +438,28 @@ async function sessionRun(session, inputs) {
...
-            }]));
+                const unpacked = {
+                    type: tensor.type,
+                    dims: tensor.dims,
+                    location: tensor.location,
+                }
+                if (unpacked.location !== "gpu-buffer") {
+                    // Only return the data if it's not a GPU buffer
+                    unpacked.data = tensor.data;
+                }
```
xenova, line 460: Needed, otherwise another error is thrown because data is on GPU, not CPU, so we can't access it with `.data` without awaiting it's contents (not needed for an error message)

PR #1297, opened 2025-04-27
Attempt to slim model if onnx check model fails

xenova:
It seems there is a bug in `onnx.checker.check_model` which throws an error when attempting to perform shape inference on models which run correctly in onnxruntime:
```
InferenceError: [ShapeInferenceError] Inference error(s): (op_type:If, node name: optimum::if): [ShapeInferenceError] Inference error(s): (op_type:Add, node name: /model/decoder/embed_positions/Add): [ShapeInferenceError] Inferred shape and existing shape differ in rank: (1) vs (0)
```

Passing the model through onnxslim seems to resolve the issue.

### DIFF
--- a/scripts/quantize.py
+++ b/scripts/quantize.py
@@ -220,6 +220,12 @@ def quantize_fp16(
         disable_shape_infer=disable_shape_infer,
         op_block_list=blocked_ops,
     )
+    try:
+        # Check the model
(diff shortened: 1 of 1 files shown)


xenova:
No longer used anymore; https://github.com/huggingface/transformers.js/pull/1382

PR #1300, opened 2025-04-28
Add support for qwen3

xenova:
Example usage:
```js

  "text-generation",
  "onnx-community/Qwen3-0.6B-ONNX",
  { dtype: "q4f16", device: "cpu" },
);

const messages = [
  { role: "user", content: "If 5 brog 5 is 1, and 4 brog 2 is 2, what is 3 brog 1?" },
];

// Generate a response
    max_new_tokens: 1024,
    do_sample: true,
    top_k: 20,
    temperature: 0.7,
    streamer: new TextStreamer(generator.tokenizer, { skip_prompt: true, skip_special_tokens: true}),
});
```

<details>

<summary>See example output</summary>

```
<think>
Okay, let's see. The problem is about a relationship between numbers and quantities. The user provided two examples: If 5 brog 5 is 1, and 4 brog 2 is 2, and asks what is 3 brog 1. 

First, I need to understand what "brog" means here. Maybe "brog" signifies a specific relationship or something else. Let me think... In some contexts, especially math problems, "brog" could be a substitution or conversion. For example, if "x brog y" translates to something like "x multiplied by y" in mathematical terms. Let me test that.

If "x brog y" equals x multiplied by y, then "5 brog 5" would be 5 * 5 = 25. But the problem says "5 brog 5 is 1" which contradicts that. So maybe that's not the case. 

Alternatively, maybe "brog" is a form of translation or a special term. For example, "x brog y" could mean "x divided by y". Let me check: If 5 brog 5 is 1, then 5 divided by 5 is 1, which fits. Similarly, if 4 brog 2 is 2, then 4 divided by 2 is 2, which matches. That seems to fit. 

So, if "x brog y" means x divided by y, then 3 brog 1 would be 3 divided by 1, which is 3. So the answer should be 3. 

Wait, but let me confirm. Another way to think about it: "brog" as a ratio. So "x brog y" equals x/y. Then 3 brog 1 is 3/1 = 3. That seems consistent with the given examples. The first example: 5 brog 5=1 implies 5/5=1, which is correct. The second example: 4 brog 2=2 implies 4/2=2, correct again. So this logic holds. Then the answer should be 3.
</think>

If **x brog y** translates to **x divided by y**, then:

- **5 brog 5** = $ \frac{5}{5} = 1 $
- **4 brog 2** = $ \frac{4}{2} = 2 $

Therefore, **3 brog 1** = $ \frac{3}{1} = 3 $.

**Answer:** 3
```

</details>

https://huggingface.co/spaces/webml-community/qwen3-webgpu

https://github.com/user-attachments/assets/d5262390-b70b-4310-b4f3-98be0af79cca


jincdream:
cant use webgpu？
image
image

xenova:
@jincdream does the demo work for you: https://huggingface.co/spaces/webml-community/qwen3-webgpu?

If not, could you provide information about your environment (browser, OS, etc.)

Issue #1301, opened 2025-04-29
docs: Improve Visibility of Recommended Options During create-next-app Initialization


subigya-js:
While initializing a new Next.js project using the command:
```
npx create-next-app@latest
```
users are prompted to choose between multiple setup options (e.g., TypeScript, Tailwind CSS, ESLint, App Router, etc.).

Currently, the recommended choices are displayed using bold text. However, in terminal environments—especially those with dark themes—the bold formatting is not always easily distinguishable from regular text. This can lead to users missing the recommended configuration options.

### Suggestion:
To improve user experience and visibility, it would be helpful to apply a more attentive and distinctive text color (such as green or cyan) to the recommended choices. This visual enhancement would make it easier for users to identify and follow best practices when creating a new Next.js project.

### How to discover:
Visit the [Next.js Installation](https://huggingface.co/docs/transformers.js/en/tutorials/next) part of the transformer.js documentation and reference to ***"Step 1 of Client-side inference"***

### Screenshot:

<img width="828" alt="Image" src="https://github.com/user-attachments/assets/1b40c395-4825-4cab-9823-e1dfa3e04d5d" />

xenova:
(see response [here](https://github.com/huggingface/transformers.js/pull/1302#pullrequestreview-2872849356))

PR #1302, opened 2025-04-29
docs: Fix formatting for project setup prompts in Next.js tutorial

subigya-js:
#1301 
Updated the installation part and gave green text color to the recommended option.
Screenshot 2025-04-29 at 12 15 44 PM


xenova (changes requested):
Thanks for the PR :)

```
<span style="color: green;"><b>No</span></b>
```

Unfortunately, this is invalid html/md due to the interleaving of `<span>` and `<b>` tags. Could you fix that?

PR #1306, opened 2025-05-02
ONNX Runtime improvements

xenova:


src/env.js
```diff
@@ -28,11 +28,16 @@ import url from 'url';
 
 const VERSION = '3.5.2';
 
+const IS_PROCESS_AVAILABLE = typeof process !== 'undefined';
+const IS_NODE_ENV = IS_PROCESS_AVAILABLE && process?.release?.name === 'node';
```
sroussey, line 32: This will be true for Bun, which does not have webgpu support.

xenova: The native EP doesn't rely on `navigator.gpu`, so we can still run it in the same way:
```js

// Create a feature-extraction pipeline
const extractor = await pipeline(
  "feature-extraction",
  "Xenova/all-MiniLM-L6-v2",
  { device: "webgpu" }
);

// Compute sentence embeddings
const sentences = ["Hello world", "This is an example sentence"];
const output = await extractor(sentences, { pooling: "mean", normalize: true });
console.log(output.tolist());
```

output:
```sh
$ bun run testing/bun.js 
dtype not specified for "model". Using the default dtype (fp32) for this device (webgpu).
2025-06-19 16:19:03.276 bun[51733:2036459] 2025-06-19 16:19:03.276206 [W:onnxruntime:, session_state.cc:1276 VerifyEachNodeIsAssignedToAnEp] Some nodes were not assigned to the preferred execution providers which may or may not have an negative impact on performance. e.g. ORT explicitly assigns shape related ops to CPU to improve perf.
2025-06-19 16:19:03.276 bun[51733:2036459] 2025-06-19 16:19:03.276244 [W:onnxruntime:, session_state.cc:1278 VerifyEachNodeIsAssignedToAnEp] Rerunning with verbose output on a non-minimal build will show node assignments.
[
  [
    -0.03447728604078293, 0.031023163348436356, 0.006734936963766813, ...
  ], [
    0.06765685230493546, 0.06349590420722961, 0.04871312156319618, ...
  ]
]
```

xenova:
Merging into v4 branch 🔥

PR #1322, opened 2025-05-27
fix: resolve webpack import.meta warning in browser extension

devlux76:
- Add parser configuration to disable import.meta parsing in webpack
- Suppress critical dependency warnings from @huggingface/transformers
- Add Node.js fallbacks for browser compatibility (fs, path, url)
- Ensures clean build without warnings while maintaining ML functionality
- Extension now builds successfully for both development and production

Resolves webpack 5 compatibility issue with Transformers.js library when bundling for browser extension environment.

### DIFF
--- a/examples/extension/webpack.config.js
+++ b/examples/extension/webpack.config.js
@@ -22,6 +22,26 @@ const config = {
         path: path.resolve(__dirname, 'build'),
         filename: '[name].js',
     },
+    resolve: {
+        fallback: {
+            "fs": false,
+            "path": false,
+            "url": false,
+        }
+    },
+    module: {
+        parser: {
+            javascript: {
+                importMeta: false,
+            },
+        },
+    },
+    ignoreWarnings: [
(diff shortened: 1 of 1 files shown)


vindex10:
Hi! @xenova I just checked, and it is still an issue. I ran current "transformers.js-examples" and got this build error:

[build.log](https://github.com/user-attachments/files/23287137/build.log)

I also tried to apply changes suggested in this PR. the PR itself works, but it is missing the lines below when compared to the current "-extensions" solution.

```
    alias: {
      "@huggingface/transformers": path.resolve(
        __dirname,
        "node_modules/@huggingface/transformers",
      ),
    },
```

when this is added (to match current "transformers.js-examples"), the error in the build.log above appears again.

vindex10:
there is actually a pull request in transformer.js-examples to fix this example:

https://github.com/huggingface/transformers.js-examples/pull/71

PR #1325, opened 2025-05-27
Improve typescript exports

xenova:
- Expose commonly-used types


xenova:
Added these for now, will continue to grow list depending on what devs need!

Issue #1332, opened 2025-06-02
TypeError: Cannot read properties of undefined (reading 'create')


A-Ghattas:
### System Info

deno 2.2.9
https://cdn.jsdelivr.net/npm/@huggingface/transformers
Windows 11

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension
- [x] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

I have seen the issue https://github.com/huggingface/transformers.js/issues/741
It didn't help at all.

My code:

```js
import {
  env,
  pipeline,
} from "https://cdn.jsdelivr.net/npm/@huggingface/transformers";

// Preparation for Deno runtime
env.useBrowserCache = false;
env.allowLocalModels = false;

export const test= async (text: string) => {
    const answerer = await pipeline(
      "question-answering",
      "Xenova/distilbert-base-uncased-distilled-squad"
    );
    const question = "Who's a nice puppet?";
    const context = "Jim Henson was a nice puppet.";
    const output = await answerer(question, context);

    return output;
};
```

I followed multiple tutorials including this one
https://supabase.com/blog/hugging-face-supabase
that says edge functions and deno support the transformers.js library
I successfully made the same code work with node but I can't make it work with deno or with the supabase edge function
I have the error 
```
error: TypeError: Cannot read properties of undefined (reading 'create')
    at _ (https://cdn.jsdelivr.net/npm/@huggingface/transformers:1:454916)
    at https://cdn.jsdelivr.net/npm/@huggingface/transformers:1:507508
    at eventLoopTick (ext:core/01_core.js:178:7)
    at async Promise.all (index 0)
    at async L (https://cdn.jsdelivr.net/npm/@huggingface/transformers:1:504311)
    at async Promise.all (index 0)
    at async st.from_pretrained (https://cdn.jsdelivr.net/npm/@huggingface/transformers:1:518423)
    at async pu.from_pretrained (https://cdn.jsdelivr.net/npm/@huggingface/transformers:1:563218)
    at async Promise.all (index 1)
    at async https://cdn.jsdelivr.net/npm/@huggingface/transformers:1:670826
```

### Reproduction

1. Create deno app
2. Run the function test
3. Enjoy the error

xenova:
[...]

PR #1333, opened 2025-06-02
Improve detection & usage of deno/bun

xenova:
Closes https://github.com/huggingface/transformers.js/issues/1332

- Uses `node:` prefixes for built-in libraries, ensuring standardization across node, bun, and deno runtimes
- Disable browser cache by default in Deno runtime (because the web cache API is available, but our URLs are not valid because we are using relative file paths due to the filesystem access we have)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ __pycache__
 node_modules
 .cache
 .DS_STORE
+deno.lock
 
 # Do not track build artifacts/generated files
 /dist
--- a/package.json
+++ b/package.json
@@ -16,6 +16,10 @@
         "default": "./dist/transformers.node.cjs"
       }
     },
+    "worker": {
+      "types": "./types/transformers.d.ts",
+      "default": "./dist/transformers.js"
+    },
     "default": {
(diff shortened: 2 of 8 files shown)


xenova:
For posterity, I was working on improving cloudflare worker integration, but I wasn't able to get it working. Here are things I learned (mainly for future me):


1. You can use conditional exports that cloudflare workers will use specifically:
```json
  "exports": {
    ...
    "worker": {
      "types": "./types/transformers.d.ts",
      "default": "./dist/transformers.js"
    },
    ...
  },
```

2. Cloudflare workers don't support relative URLs (e.g., `new URL('.')` is invalid)
3. Cloudflare workers don't support .node addon binaries (https://github.com/cloudflare/workerd/discussions/1905#discussioncomment-9769971). This is unlikely to change in future.
4. Cloudflare workers disable wasm code generation (https://github.com/cloudflare/next-on-pages/issues/704) - this is a security consideration and is unlikely to change in future.
5. You can detect cloudflare worker environment simply with `typeof globalThis.Cloudflare !== 'undefined'`

Issue #1343, opened 2025-06-18
numpy break change np.fromstring binary mode removed


xiaoyao9184:
### System Info

pip show numpy
Name: numpy
Version: 2.3.0

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension
- [ ] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [x] Other (e.g., VSCode extension)

### Description

Convert model to ONNX

### Reproduction

python -m scripts.convert --quantize --model_id Qwen/Qwen2.5-0.5B-Instruct

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/ubuntu/code/transformers.js/scripts/convert.py", line 456, in <module>
    main()
  File "/home/ubuntu/code/transformers.js/scripts/convert.py", line 428, in main
    quantize(
  File "/home/ubuntu/code/transformers.js/scripts/quantize.py", line 309, in quantize
    quantize_fp16(
  File "/home/ubuntu/code/transformers.js/scripts/quantize.py", line 217, in quantize_fp16
    model_fp16 = float16.convert_float_to_float16(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/code/transformers.js/scripts/float16.py", line 255, in convert_float_to_float16
    value_info_block_list = process_tensor_in_node(
                            ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/code/transformers.js/scripts/float16.py", line 510, in process_tensor_in_node
    convert_tensor_float_to_float16(
  File "/home/ubuntu/code/transformers.js/scripts/float16.py", line 138, in convert_tensor_float_to_float16
    float32_list = np.fromstring(tensor.raw_data, dtype="float32")
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: The binary mode of fromstring is removed, use frombuffer instead
```

fidoriel:
just install numpy 2.2.

PR #1351, opened 2025-07-01
Pin numpy version for scripts

fidoriel:
This is not a definitive version but should fix current issues.
Closes https://github.com/huggingface/transformers.js/issues/1343

### DIFF
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -4,3 +4,4 @@ optimum@git+https://github.com/huggingface/optimum.git@b04feaea78cda58d79b8da67d
 onnx==1.17.0
 tqdm==4.67.1
 onnxslim==0.1.48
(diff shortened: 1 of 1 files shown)


scripts/requirements.txt
```diff
@@ -4,3 +4,4 @@ optimum@git+https://github.com/huggingface/optimum.git@b04feaea78cda58d79b8da67d
 onnx==1.17.0
 tqdm==4.67.1
 onnxslim==0.1.48
+numpy==2.2
```
xenova:
```suggestion
numpy==2.2.6
```

xenova: maybe 2.2.6 instead of 2.2.0?

xenova:
Thanks!

Issue #850, opened 2024-07-17
The scripts/convert.py script fails for a few reasons


lancejpollard:
### System Info

- Python 3.12.3
- Mac OSX M3

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension
- [X] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [X] Other (e.g., VSCode extension)

### Description

I am trying to run the model locally, as [it doesn't appear to work running a remote model in Node.js](https://stackoverflow.com/questions/78757326/how-to-use-huggingfaces-transformers-js-to-distill-messy-dictionary-definitions).

First I followed https://github.com/xenova/transformers.js/blob/main/scripts/convert.py (which is [linked in the README](https://github.com/xenova/transformers.js?tab=readme-ov-file#convert-your-models-to-onnx)):

```
$ python3 -m pip install -r requirements.txt
Collecting transformers==4.33.2 (from transformers[torch]==4.33.2->-r requirements.txt (line 1))
  Using cached transformers-4.33.2-py3-none-any.whl.metadata (119 kB)
ERROR: Could not find a version that satisfies the requirement onnxruntime<1.16.0 (from versions: 1.17.0, 1.17.1, 1.17.3, 1.18.0, 1.18.1)
ERROR: No matching distribution found for onnxruntime<1.16.0
```

So that `onnxruntime<1.16.0` does not seem to exist in `pip`.

Can you update that script?

Second, I tried just installing the latest versions of everything instead, by making this the `requirements.txt`:

```txt
transformers
onnxruntime
optimum
tqdm
onnx
```

But after I ran this:

```
$ python3 -m pip install -r requirements.txt
... successful installation stuff...
$ python3 -m convert --quantize --task summarization --model_id bart-large-cnn
```

I got an error:

```
TypeError: quantize_dynamic() got an unexpected keyword argument 'optimize_model'
```

Full stack trace:

```
Framework not specified. Using pt to export the model.
The task `text2text-generation` was manually specified, and past key values will not be reused in the decoding. if needed, please pass `--task text2text-generation-with-past` to export using the past key values.
Using the export variant default. Available variants are:
    - default: The default ONNX variant.
Some non-default generation parameters are set in the model config. These should go into a GenerationConfig file (https://huggingface.co/docs/transformers/generation_strategies#save-a-custom-decoding-strategy-with-your-model) instead. This warning will be raised to an exception in v4.41.
[...]

PR #1368, opened 2025-07-17
fix: update conversion script for onnx v1.18.0 and Python 3.13.2 compatibility

tshasan:
This PR updates the conversion scripts to be compatible with ONNX v1.18.0 and Python 3.13.2.
## Changes Made
- **Updated ONNX dependency**: Bumped from `onnx==1.17.0` to `onnx==1.18.0` in `requirements.txt`
- **Fixed import statements**: Changed from relative imports to absolute imports in `convert.py` and `quantize.py`
- **Added missing imports**: Added `ORTQuantizer` and `QuantizationConfig` imports that are now required
- **Updated deprecated imports**: Fixed import paths that changed in newer versions of optimum
The previous import structure was causing issues with newer Python versions and the ONNX dependency was outdated. These changes ensure the conversion scripts work properly in modern Python environments.

### DIFF
--- a/scripts/convert.py
+++ b/scripts/convert.py
@@ -1,4 +1,3 @@
-
 import json
 import os
 import shutil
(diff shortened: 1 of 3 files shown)


xenova:
No longer needed; see https://github.com/huggingface/transformers.js/issues/850#issuecomment-4000183624

PR #1381, opened 2025-07-30
Optimize tensor.slice()

Honry:
The performance of executing `tensor.slice()` is super poor, especially for the 'logits' tensor with large dimensions.

```
const logits = outputs.logits.slice(null, -1, null);`
```

This is because currently implementation of the `slice` method manually iterates through each element and calculate indices which is a big time consuming if the tensor shape is large.

For cases like `slice(null, -1, null)`, where the slicing operation is contiguous along certain dimensions, which can be optimized by bulk copy by using `TypeArray.subarray()` and `TypeArray.set()`.

### DIFF
--- a/src/utils/tensor.js
+++ b/src/utils/tensor.js
@@ -464,7 +464,7 @@ export class Tensor {
             } else if (Array.isArray(this_data)) {
                 // If this.data is a plain array, use slice
                 const slicedData = this_data.slice(start, end);
-                for (let i = 0; i < slicedData.length; i++) {
+                for (let i = 0; i < slicedData.length; ++i) {
                     data[i] = slicedData[i];
                 }
             } else {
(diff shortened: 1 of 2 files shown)


Honry:
@xenova, PTAL, thanks!

On my test device, this can improve the execution time for `const logits = outputs.logits.slice(null, -1, null);` from 11ms to 0.3ms for Qwen model (with logits shape: `[batch_size,sequence_length,151936]`).

cc/ @ibelem

xenova:
Oh wow, this looks like a great PR! Running tests and reviewing now!

Thanks @Honry

xenova:
Very impressive: ran this with qwen3 and got a pretty substantial TPS increase 🙌 

Before: 45.087279331416916 tokens per second
After: 50.945617704855884 tokens per second

+13% speed boost to generate 512 tokens 🔥 

Merging now!

PR #1382, opened 2025-07-31
Transformers.js V4: Native WebGPU EP, repo restructuring, and more!

xenova:

PR #1386, opened 2025-08-06
Address deprecation in fromstring to frombuffer

hmaarrfk:
Deprecated since numpy 1.14 according to
https://github.com/numpy/numpy/pull/28254

This should be the correct fix for https://github.com/huggingface/transformers.js/issues/1343

And after this is merged, one can likely just revert this https://github.com/huggingface/transformers.js/pull/1351

### DIFF
--- a/scripts/float16.py
+++ b/scripts/float16.py
@@ -135,7 +135,7 @@ def convert_tensor_float_to_float16(tensor, min_positive_val=1e-7, max_finite_va
         # convert raw_data (bytes type)
         if tensor.raw_data:
             # convert n.raw_data to float
-            float32_list = np.fromstring(tensor.raw_data, dtype="float32")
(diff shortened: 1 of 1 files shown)


hmaarrfk:
was hoping that the tests could validate this code since I haven't tested it locally. marking this as draft for now.

hmaarrfk:
i guess the larger problem is that tensors could be bfloat16 or something like that.

hmaarrfk:
Seems like this code was taken from
https://github.com/microsoft/onnxconverter-common/blob/master/onnxconverter_common/float16.py#L125

that moved from fromstring to frombuffer 2 months ago in
https://github.com/microsoft/onnxconverter-common/pull/314

xenova:
Closed by https://github.com/huggingface/transformers.js/pull/1382

Issue #1191, opened 2025-02-12
Speeding up `Tensor.tolist()`


tarekziade:
There might be a way to speed up Tensor.tolist() by simplifying its code 

below is a side-by-side comparison of my version with the 3.0.2 one we ship in FF

```
var transformers = ChromeUtils.importESModule("chrome://global/content/ml/transformers-dev.js");

class CustomTensor extends transformers.Tensor {
    tolist() {
        const shape = this.dims;
        const data = this.data;
        if (shape.length === 1) return [...data];
        const result = new Array(shape[0]);
        const rowSize = shape[1];
        
        for (let i = 0; i < shape[0]; i++) {
            const row = new Array(rowSize);
            for (let j = 0; j < rowSize; j++) {
                row[j] = data[i * rowSize + j];
            }
            result[i] = row;
        }
        return result;
    }
}

async function test_tolist_speed(iterations = 10) {
    const numTokens = 16;
    const embeddingDim = 768;
    const totalElements = numTokens * embeddingDim;
    const data = Array.from({ length: totalElements }, () => Math.random() * 2 - 1);
    const shape = [numTokens, embeddingDim]; // [16, 768] for a sentence embedding
    
    const tensor = new transformers.Tensor('float32', data, shape);
    const customTensor = new CustomTensor('float32', data, shape);

    let totalTimeDefault = 0;
    let totalTimeOptimized = 0;
    let allEqual = true;

    function arraysEqual(arr1, arr2) {
        if (!Array.isArray(arr1) || !Array.isArray(arr2) || arr1.length !== arr2.length) return false;
        return arr1.every((val, index) => Array.isArray(val) ? arraysEqual(val, arr2[index]) : val === arr2[index]);
    }

    for (let i = 0; i < iterations; i++) {
        let start = performance.now();
        const defaultList = tensor.tolist();
        totalTimeDefault += performance.now() - start;
        
        start = performance.now();
        const optimizedList = customTensor.tolist();
        totalTimeOptimized += performance.now() - start;
        
        if (!arraysEqual(defaultList, optimizedList)) {
            allEqual = false;
        }
    }
    console.debug(`Avg tolist() time (default): ${(totalTimeDefault / iterations).toFixed(4)} ms`);
    console.debug(`Avg tolist() time (optimized): ${(totalTimeOptimized / iterations).toFixed(4)} ms`);
    console.debug(`Lists are equal: ${allEqual}`);
}


await test_tolist_speed(); 
```

And the results:

[...]

PR #1395, opened 2025-08-19
fix: speedup tolist function

obenjiro:
## Summary
- implement stride-based `reshape` for faster nested array construction
- add tensor tests for 4D `tolist`

Fixes https://github.com/huggingface/transformers.js/issues/1191 

## The reshape-based approach
The old implementation built the nested structure through a series of reshapes:
- For each dimension, the flat array was chunked into smaller arrays.
- Each chunk was then reshaped again for the next dimension.
- This meant multiple passes over the data: one to create the first level, another for the second, and so on.

Each reshape step:
- Allocates new arrays for every chunk.
- Invokes slice or map operations, producing copies or calling callbacks for every element.
- Re-reads the same data multiple times. For an n‑dimensional tensor, the data might be iterated n separate times.
- Incurs significant JavaScript function‑call overhead, because reduce, map, and friends invoke callbacks for each element.

This results in substantial CPU time (multiple passes and callback dispatches) and memory churn (allocating and discarding intermediate arrays). The complexity effectively grows with both the number of elements and the number of dimensions.

## Recursive construction with precomputed strides
The optimized version precomputes the strides—the number of elements to skip along each dimension—and builds the nested array recursively:
- Starting at dimension 0, allocate the output array for that level.
- For each index at this level, compute the offset into the flat buffer using the stride.
- Recurse into the next dimension until the innermost dimension is reached, where elements are read directly from the typed array.

| Shape              | Current tolist (ms) | Optimized tolist (ms) | Speedup  |
|--------------------|----------------------|-----------------------|----------|
| [16, 768]          | 1.003                | 0.319                 | 3.15×    |
| [32, 768]          | 1.948                | 0.187                 | 10.42×   |
| [8, 16, 64]        | 0.645                | 0.049                 | 13.14×   |
| [8, 32, 64]        | 1.792                | 0.259                 | 6.92×    |
| [4, 8, 16, 32]     | 1.797                | 0.096                 | 18.73×   |
| [2, 4, 8, 16, 32]  | 3.537                | 0.203                 | 17.43×   |

### DIFF
--- a/src/utils/tensor.js
+++ b/src/utils/tensor.js
@@ -212,7 +212,7 @@ export class Tensor {
      * @returns {Array}
      */
     tolist() {
-        return reshape(this.data, this.dims)
+        return reshape(this.data, this.dims);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


obenjiro:
@xenova sorry for bothering you. Would you kindly look into this PR?

Issue #1372, opened 2025-07-21
[deno] OffscreenCanvas not supported by this browser for server-side worker


korywka:
### System Info

transformers: 3.6.2
deno: 2.4.1

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension
- [x] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

Using Deno and running transformers.js inside WebWorker tries to call `OffscreenCanvas`:

```
if (IS_BROWSER_OR_WEBWORKER) {
    // Running in browser or web-worker
    createCanvasFunction = (/** @type {number} */ width, /** @type {number} */ height) => {
        if (!self.OffscreenCanvas) {
            throw new Error('OffscreenCanvas not supported by this browser.');
        }
        return new self.OffscreenCanvas(width, height)
    };
    loadImageFunction = self.createImageBitmap;
    ImageDataClass = self.ImageData;

}
```

But Deno doesn't support `OffscreenCanvas`. Can we skip this condition for Deno? This issue started after update of the package version. On old version of transformers.js it was running without any error for Deno. Thanks.

### Reproduction

calling `const image = await RawImage.fromURL(...);`

korywka:
Looks like node also doesn't support `OffscreenCanvas`. So should the condition only for browser be enough?

korywka:
Any ideas? Don't you mind a PR to remove web worker environment for OffscreenCanvas?

xenova:
Hi there 👋 OffscreenCanvas is quite useful when running in-browser via web workers, so a PR to fully remove it sounds a bit excessive.

That said, perhaps adding a check here for Deno could work, since in that case, we'll just disallow image manipulation operations.

Also, if you could provide a code snippet to reproduce the error you're facing, that would be great!

korywka:
@xenova sorry for the delay, here is the minimal demo:

https://github.com/korywka/transformersjs-workers

xenova:
> But Deno doesn't support OffscreenCanvas. Can we skip this condition for Deno? This issue started after update of the package version. On old version of transformers.js it was running without any error for Deno. Thanks.

Now in Transformers.js v4, the default deno runtime (when imported by `npm:@huggingface/transformers`) will use the correct node-like runtime, and will use `sharp` for image operations. See https://github.com/huggingface/transformers.js/pull/1546.

[...]

PR #1417, opened 2025-09-17
Check for Canvas API before using instead of environment

korywka:
Issue: https://github.com/huggingface/transformers.js/issues/1372

The `IS_WEBWORKER_ENV` is `true` for Deno Worker runtime:
https://github.com/huggingface/transformers.js/blob/699dcb564ac1be64df2269451d641da567693908/src/env.js#L33

So it is not reliable condition to check for HTML/Offscreen-Canvas API usage.

One way to fix it is to go from env-based conditions to feature-based.

If we add Deno condition to exclude Canvas API, it would work for now, but I think it will be broken for Node Workers (https://github.com/nodejs/node/issues/43583) and [Bun Web Workers](https://bun.com/docs/api/workers).

Please note that this PR has been tested only with Deno, not other environments. It should be tested further, closed, or used as a starting point.

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -29,11 +29,12 @@ import url from 'node:url';
 const VERSION = '3.7.3';
 
 // Check if various APIs are available (depends on environment)
-const IS_BROWSER_ENV = typeof window !== "undefined" && typeof window.document !== "undefined";
-const IS_WEBWORKER_ENV = typeof self !== "undefined" && (['DedicatedWorkerGlobalScope', 'ServiceWorkerGlobalScope', 'SharedWorkerGlobalScope'].includes(self.constructor?.name));
+const IS_BROWSER_ENV = typeof window !== 'undefined' && typeof window.document !== 'undefined';
+const IS_WEBWORKER_ENV = typeof self !== 'undefined' && (['DedicatedWorkerGlobalScope', 'ServiceWorkerGlobalScope', 'SharedWorkerGlobalScope'].includes(self.constructor?.name));
 const IS_WEB_CACHE_AVAILABLE = typeof self !== "undefined" && 'caches' in self;
 const IS_WEBGPU_AVAILABLE = typeof navigator !== 'undefined' && 'gpu' in navigator;
 const IS_WEBNN_AVAILABLE = typeof navigator !== 'undefined' && 'ml' in navigator;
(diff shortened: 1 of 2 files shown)


xenova:
Thanks for the PR! This mostly looks good to me. pinging @nico-martin for another review :)

I tried with bun workers, but the existing code seems to work for that. Sounds like Deno has this issue, which your PR aims to fix. If you can provide some reproduction that makes it fail for Deno, I can test that in Bun.

Update: Bun seems to not use a name (i.e., `Object`) for the webworker environment, meaning the check for is web worker is always false.

Also, when both sharp and canvas is available, I'm not sure what we should default to, but I think sharp would be more performant. Open to suggestions here.

nico-martin:
I like the idea of going from runtime-detection to feature-detection.
However I was not able to reproduce the error you described:

```typescript
// index.ts
const worker = new Worker(new URL("./worker.ts", import.meta.url).href);

await new Promise((resolve, reject) => {
  worker.postMessage("hello");

  worker.onmessage = (event) => {
    console.log("message received in main", event.data);
    resolve(event.data);
  };

  worker.onerror = (error) => {
    console.error("Worker error:", error);
    reject(error);
  };
});

console.log("Worker completed");
worker.terminate();

export {};
```

```typescript
// worker.ts

declare var self: Worker;

self.onmessage = async (event: MessageEvent) => {
  console.log("message received in worker", event.data);
  const segmenter = await pipeline("background-removal", "Xenova/modnet");
  const url =
    "https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/portrait-of-woman_small.jpg";
  const output = await segmenter(url);
  postMessage(output);
};
```

I then added a console.log in transformers.js to see what `IS_WEBWORKER_ENV` is set to. Turns out it's `false`, as expected. 

@korywka, could you send me an example where I can reproduce the error? Or have I misunderstood the problem?

korywka:
@nico-martin Hello. Yep, I gave the link in the linked issue: https://github.com/huggingface/transformers.js/issues/1372#issuecomment-3303351819

nico-martin:
Thanks!
I have been able to reproduce the error with bun and verified your fix. I also tested your changes with the example above:
✅ Node v22.17.1
✅ Deno 2.2.6
✅ Bun 1.2.22
✅ Browser (Chrome 141)

lgtm!

nico-martin (approved):

PR #1421, opened 2025-09-27
Fix Docs: grammar in text-to-audio description

P1N2O:


xenova:
The original sentence is grammatically correct, so we'll keep it as is. Thanks for opening though!

PR #1431, opened 2025-10-03
Temperature zero issue

nico-martin:
added a check that sets `generation_config.do_sample` to `false` if `temperature=0`
Screenshot 2025-10-03 at 12 38 04


src/models.js
```diff
@@ -1427,6 +1427,12 @@ export class PreTrainedModel extends Callable {
             processors.push(new ClassifierFreeGuidanceLogitsProcessor(generation_config.guidance_scale));
         }
 
+
+        if (generation_config.temperature === 0 && generation_config.do_sample) {
+          console.warn('do_sample automatically set to false. When temperature: 0, the model uses deterministic sampling (always selecting the most likely token), which is incompatible with do_sample: true.');
```
xenova:
```suggestion
          console.warn('`do_sample` changed to false because `temperature: 0` implies greedy sampling (always selecting the most likely token), which is incompatible with `do_sample: true`.');
```

xenova: image

examples/webgpu-chat/package-lock.json
xenova, line 1: we can remove this file -- all examples will be moved to https://github.com/huggingface/transformers.js-examples in transformers.js v4, so this folder will be deleted

PR #1433, opened 2025-10-06
added contributing.md

nico-martin:
@xenova, here is my idea of a CONTRIBUTING.md
Couple of things to discuss (marked as [TODO] in the file)

- [ ] whats the minimum Nodejs version? Is it 18?
- [ ] Do you have thoughts about the Pull Request checklist?

I've also seen that 16 of the 1926 tests failed is that a problem? If not we should not enforce the tests in the PR checklist. Otherwise we maybe should look into it

I also had to make some changes to the webpack config. Mainly because the dev-server failed in my setup because of an error (Error: EMFILE: too many open files, watch) and I had to add CORS headers so the test-application can load transformers.js from HTTP without CORS error.

Lastly I think it makes sense to introduce the CONTRIBUTING.md in v4 so prettier works as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ node_modules
 .cache
 .DS_STORE
 deno.lock
+.idea
 
 # Do not track build artifacts/generated files
 /dist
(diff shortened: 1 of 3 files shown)


xenova:
(This was superseded by https://github.com/huggingface/transformers.js/pull/1382)

Issue #1409, opened 2025-09-08
Getting  TS2304: Cannot find name 'PretrainedProcessorOptions'. error


ShibleNK:
### System Info

transformers.js - 3.7.2
github codespaces 
node  - 22.17.0
Next.js 15 for frontend
Node.js for backend
typescript - 5.8.3

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension
- [ ] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

I was trying to use the gemma-3-1b-it-ONNX-GQA for giving prompt to generate text. 

### Reproduction

package.json
```
{
  "devDependencies": {
    "@huggingface/transformers": "^3.7.2"
  }
}
```

tsconfig.json
```
{
  "compilerOptions": {
    "target": "ES6",
    "baseUrl": "./src",
    "rootDir": "./src",
    "module": "commonjs",
    "paths": {
      "@/*": ["*"]
    },
    "typeRoots": ["./src/types", "./node_modules/@types"],
    "strict": true,
    "outDir": "./dist",
    "esModuleInterop": true
  },
  "include": ["./src/**/*"]
}
```

main.ts
```
import { pipeline } from "@huggingface/transformers";
```

executing tsc returns -

```../node_modules/.pnpm/@huggingface+transformers@3.7.2/node_modules/@huggingface/transformers/types/models/auto/processing_auto.d.ts:32:77 - error TS2304: Cannot find name 'PretrainedProcessorOptions'.

32     static from_pretrained(pretrained_model_name_or_path: string, options?: PretrainedProcessorOptions): Promise<Processor>;
                                                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~

../node_modules/.pnpm/@huggingface+transformers@3.7.2/node_modules/@huggingface/transformers/types/models/mgp_str/processing_mgp_str.d.ts:52:5 - error TS2416: Property 'batch_decode' in type 'MgpstrProcessor' is not assignable to the same property in base type 'Processor'.
  Type '([char_logits, bpe_logits, wp_logits]: Tensor[]) => { generated_text: string[]; scores: number[]; char_preds: string[]; bpe_preds: string[]; wp_preds: string[]; }' is not assignable to type '(batch: number[][] | Tensor, decode_args?: any) => string[]'.
    Types of parameters '__0' and 'batch' are incompatible.
[...]

Issue #1337, opened 2025-06-10
3.5.2 typescript build errors


jozefchutka:
### System Info

- transformers.js: 3.5.2

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension
- [ ] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

Running a very basic typescript build using transformers.js 3.5.2 fails.

### Reproduction

package.json:

```json
{
	"devDependencies": {
		"@huggingface/transformers": "^3.5.2"
	}
}
```

tsconfig.json:

```json
{
	"compilerOptions": {
	  "target": "ESNEXT",
	  "lib": ["ESNext", "WebWorker"],
	  "module": "ESNext",
	  "moduleResolution": "bundler",
	  "baseUrl": "./",
	  "paths": {}
	}
  }
```

Main.mts

```ts
import { AutomaticSpeechRecognitionConfig, env, pipeline, ProgressCallback } from "@huggingface/transformers";
```

runnig `tsc` command fails with the following errors:

```
node_modules/@huggingface/transformers/types/models/auto/processing_auto.d.ts:32:76 - error TS2304: Cannot find name 'PretrainedProcessorOptions'.

32     static from_pretrained(pretrained_model_name_or_path: string, options: PretrainedProcessorOptions): Promise<Processor>;
                                                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@huggingface/transformers/types/models/mgp_str/processing_mgp_str.d.ts:52:5 - error TS2416: Property 'batch_decode' in type 'MgpstrProcessor' is not assignable to the same property in base type 'Processor'.
  Type '([char_logits, bpe_logits, wp_logits]: Tensor[]) => { generated_text: string[]; scores: number[]; char_preds: string[]; bpe_preds: string[]; wp_preds: string[]; }' is not assignable to type '(batch: number[][] | Tensor, decode_args?: any) => string[]'.
    Types of parameters '__0' and 'batch' are incompatible.
      Type 'number[][] | Tensor' is not assignable to type 'Tensor[]'.
        Type 'number[][]' is not assignable to type 'Tensor[]'.
          Type 'number[]' is missing the following properties from type 'Tensor': dims, type, data, size, and 47 more.

[...]

PR #1436, opened 2025-10-10
Type error fixes

nico-martin:
The following errors from https://github.com/huggingface/transformers.js/issues/1409 are adressed by this PR:
-  Added a type definition (@typedef) that tells TypeScript what PretrainedProcessorOptions is
-  Changed the parameter type from "an array of Tensors" to "exactly 3 Tensors" (char, bpe, wp)
- The sharp import error (needs esModuleInterop)

Closes https://github.com/huggingface/transformers.js/issues/1337
Closes https://github.com/huggingface/transformers.js/issues/1409


src/models/mgp_str/processing_mgp_str.js
```diff
@@ -119,8 +119,7 @@ export class MgpstrProcessor extends Processor {
      * - bpe_preds: The list of BPE decoded sentences.
      * - wp_preds: The list of wp decoded sentences.
      */
-    // @ts-expect-error The type of this method is not compatible with the one
-    // in the base class. It might be a good idea to fix this.
+    // @ts-expect-error The type of this method is not compatible with the one in the base class.
```
xenova, line 122: ah thanks! Probably why it wasn't being registered correctly.

tsconfig.json
```diff
@@ -13,7 +13,8 @@
     "declaration": true,
     "declarationMap": true,
     "noEmit": false,
-    "emitDeclarationOnly": true
+    "emitDeclarationOnly": true,
+    "esModuleInterop": true
```
xenova, line 17: what does this do (specifically needed for this PR)? 👀 https://www.typescriptlang.org/tsconfig/#esModuleInterop

.gitignore
```diff
@@ -5,6 +5,7 @@ node_modules
 .DS_STORE
 deno.lock
 .idea
+demo
```
xenova:
```suggestion
```

nico-martin: You're right. I removed it.

tsconfig.json
nico-martin: I got an error file investigation this issue:
```
../types/utils/image.d.ts:140:8 - error TS1259: Module '"/path/transformers.js/node_modules/sharp/lib/index"' can only be default-imported using the 'esModuleInterop' flag

140 import sharp from 'sharp';
```
The problem is that sharp is using commonJS (require). This will ensure that interop between CommonJS ed es modules works better. As far as I know it doesn't affect runtime behavior. It only helps TypeScript understand CommonJS modules better during type checking and generation.

xenova: gotcha! thanks :)

PR #1449, opened 2025-11-01
[TimeSeries] add AutoModelForForecasting with chronos2 

kashif:
Onnx conversion scripts: https://github.com/amazon-science/chronos-forecasting/pull/359

### DIFF
--- a/src/configs.js
+++ b/src/configs.js
@@ -179,6 +179,7 @@ function getNormalizedConfig(config) {
 
         // Encoder-decoder models
         case 't5':
+        case 'chronos2':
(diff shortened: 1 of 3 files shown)


xenova:
Very cool! Thanks for the PR 🥳 

Correct me if I'm wrong, but `transformers` doesn't currently have a time series pipeline, right? Similarly, I was unable to find the `AutoModelForForecasting` (and other) classes. Is this custom to chronos models?

Ideally, having a couple of time-series models compatible with the pipeline API would be useful for standardization (and to prevent breaking changes in future).

PR #1455, opened 2025-11-14
ImageProcessor preprocess uses image_std for fill value

NathanKolbas:


xenova (approved):
Thanks! Silly typo 😅 

(luckily this was almost never triggered because by default, image_std is an array.)

PR #1464, opened 2025-11-22
fixed backend loading path

nico-martin:
The env.backends.onnx is set at the end of the file while env.backends.onnx.versions?.web is requested already on line 197.
That means the check wether the jsdelivr file should be requested is always `false`.

This PR ensures that https://cdn.jsdelivr.net is used if not in service worker.


nico-martin:
FYI, the reason it did work in dev when using transformers.js via http `import { pipeline } from "http://localhost:8080/dist/transformers.js";` is that in this case, the wasm file is actually at the correct place `http://localhost:8080/dist/ort-wasm-simd-threaded.mjs`. But when using it via npm it could not resolve correctly.

xenova (approved):
Thanks!

PR #1465, opened 2025-11-26
V4 apertus

nico-martin:


sroussey:
Is there a timeframe for v4?

I'd love to get my abort controller stuff in there.

Also, I have a website that does a workflows with a decent UI using transformers.js and would be nice to make announcement at same time. Local users, local models, local runtime.

xenova (approved):
LGTM ✅

PR #1466, opened 2025-11-27
[v4] Switch build system to use esbuild

nico-martin:


xenova:
Amazing -- I will be testing this out in the next few days! 😍

package.json
```diff
@@ -56,8 +56,10 @@
   "homepage": "https://github.com/huggingface/transformers.js#readme",
   "dependencies": {
     "@huggingface/jinja": "^0.5.1",
+    "esbuild": "^0.27.0",
     "onnxruntime-node": "1.24.0-dev.20251104-75d35474d5",
     "onnxruntime-web": "1.24.0-dev.20251104-75d35474d5",
+    "rimraf": "^6.1.2",
```
xenova: should be devDependencies

scripts/esbuild/build.mjs
```diff
@@ -0,0 +1,146 @@
...
+    sourcemap: true,
+    minify: true,
+    external: externalModules,
+    plugins,
+    legalComments: "none",
+    logOverride: {
+      // Suppress import.meta warning for CJS builds - it's handled gracefully in the code
+      "empty-import-meta": "silent",
+    },
+  });
```
xenova, line 69: minimize duplication by defining a list of defaults which we unpack unto each config.

scripts/esbuild/build.mjs
```diff
@@ -0,0 +1,146 @@
...
+  reportSize(path.join(outdir, minFile));
+}
+
+console.log("\nBuilding transformers.js with esbuild...\n");
+
+const startTime = performance.now();
+
+try {
+  console.log("=== CLEAN ===");
+  execSync(`rimraf ${DIST_FOLDER}`, { stdio: "inherit" });
```
xenova: if we are going to call rimraf using execSync, we might as well just use native fs functions to clear the dist directory. -1 dependency :)

scripts/esbuild/build/constants.mjs
```diff
@@ -0,0 +1,13 @@
+export const DIST_FOLDER = "dist";
+export const NODE_IGNORE_MODULES = ["onnxruntime-web"];
+export const NODE_EXTERNAL_MODULES = [
+  "onnxruntime-common",
+  "onnxruntime-node",
+  "sharp",
+  "node:fs",
+  "node:path",
+  "node:url",
+];
```
xenova, line 11: the `node:...` list may grow in future. Is there a way to specify a pattern like `/^node:.*/`?

xenova: (this may only be possible in the section of code where we use it)

scripts/esbuild/build/plugins/ignoreModulesPlugin.mjs
```diff
@@ -0,0 +1,33 @@
...
+      return {
+        contents: `
+          const noop = () => {};
+          const emptyObj = {};
+          export default emptyObj;
+          export const Readable = { fromWeb: noop };
+          export const pipeline = noop;
+          export const createWriteStream = noop;
+          export const createReadStream = noop;
+        `
```
xenova: Will this be added one, or for each module we will ignore? 👀

scripts/esbuild/build/plugins/postBuildPlugin.mjs
```diff
@@ -0,0 +1,41 @@
...
+  let completed = false;
+  return {
+    name: "post-build",
+    setup(build) {
+      build.onEnd(() => {
+        if (completed) return;
+        completed = true;
+
+        const ORT_JSEP_FILE = 'ort-wasm-simd-threaded.jsep.mjs';
+        const ORT_BUNDLE_FILE = 'ort.bundle.min.mjs';
```
xenova: [WIP] How do we ensure this is bundled with the main library, enabling https://github.com/huggingface/transformers.js/pull/1471

scripts/esbuild/dev.mjs
```diff
@@ -0,0 +1,97 @@
...
+const outdir = path.join(rootDir, DIST_FOLDER);
+const startTime = performance.now();
+
+// Ensure output directory exists
+if (!existsSync(outdir)) {
+  mkdirSync(outdir, { recursive: true });
+}
+
+console.log("=== CLEAN ===");
+execSync(`rimraf ${outdir}`, { stdio: "inherit" });
```
xenova: same comment as above.

scripts/esbuild/dev.mjs
```diff
@@ -0,0 +1,97 @@
...
+  external: ["onnxruntime-common", "onnxruntime-web"],
+  plugins: [
+    ignoreModulesPlugin(WEB_IGNORE_MODULES),
+    stripNodePrefixPlugin(),
+    rebuildPlugin("Web"),
+  ],
+  logOverride: {
+    "empty-import-meta": "silent",
+  },
+});
```
xenova, line 42: same comment as above RE: duplication.

Maybe we can also consolidate dev and build logic to reduce duplication

xenova (changes requested):
Thanks for the PR! Good start 👍

Issue #1236, opened 2025-03-14
Disable Warnings


shiffman:
### Feature request

Hi @xenova, we were just chatting about this IRL! Woohoo! I'm working on a set of examples for transformers.js + p5.js. The target user is a beginner in the [p5.js web editor](https://editor.p5js.org/). I'd like to be able to disable certain warnings as they how up as large red blocks in the console and beginners can be quite thrown off and confused by them! See:

<img width="1647" alt="Image" src="https://github.com/user-attachments/assets/aa2928c1-8f35-4bae-bcb6-eb2ec14073ba" />

### Motivation

This is to help beginners to transformers.js, specifically with the p5.js web editor, but perhaps it would be helpful to others as well! Thank you!!

### Your contribution

I'm happy to help submit a PR for this, would love any tips or to be pointed to a direction in the code as to where this feature would be best!

rakesh-163:
Is there any update on this? In browser extensions, these show up as errors by Chrome and thus confusing the users. Happy to submit a PR too if someone who understands the codebase could sketch out a high-level approach for a fix.

obenjiro:
@rakesh-163 @shiffman Have you tried setting `env.backends.onnx.logLevel = 'error';` ?

PR #1468, opened 2025-11-30
[v4] suppress console.error while creating InferenceSession

nico-martin:
I invested some time into #1236 and from what I see there are three ways how we could solve this:
1. suppress console.error while creating InferenceSession
2. while copying the wasm .mjs file we could overwrite the line that cases the console.error
3. change the wasm library .mjs file.

for 3. We (or the ONNX maintainers) would have to make changes directly in their library. This would certainly be the cleanest solution, but we would be dependent on third parties. 2. I would not recommend this, as it could disrupt the copying process if changes are made within the ONNX library.
That's why solution one would be my favorite. It looks a bit messy or hacky, but honestly, I think it's the simplest solution.

### DIFF
--- a/src/backends/onnx.js
+++ b/src/backends/onnx.js
@@ -46,6 +46,9 @@ const DEVICE_TO_EXECUTION_PROVIDER_MAPPING = Object.freeze({
     'webnn-cpu': { name: 'webnn', deviceType: 'cpu' }, // WebNN CPU
 });
 
+/** @type {Array<'verbose' | 'info' | 'warning' | 'error' | 'fatal'>} */
+const LOG_LEVELS = ['verbose', 'info', 'warning', 'error', 'fatal'];
+
 /**
  * The list of supported devices, sorted by priority/performance.
  * @type {import("../utils/devices.js").DeviceType[]}
@@ -150,8 +153,6 @@ let webInitChain = Promise.resolve();
  */
 export async function createInferenceSession(buffer_or_path, session_options, session_config) {
 
-    /** @type {Array<'verbose' | 'info' | 'warning' | 'error' | 'fatal'>} */
-    const LOG_LEVELS = ['verbose', 'info', 'warning', 'error', 'fatal'];
     /** @type {0|1|2|3|4} */
     const logSeverityLevel =
         typeof session_options.logSeverityLevel !== 'number' ||


xenova:
Thanks for the PR! I recall that for V3, the inference session logging options would not work
```
[logSeverityLevel](https://onnxruntime.ai/docs/api/js/interfaces/InferenceSession.SessionOptions.html#logSeverityLevel)?: 0 | 2 | 1 | 3 | 4;
    [logVerbosityLevel](https://onnxruntime.ai/docs/api/js/interfaces/InferenceSession.SessionOptions.html#logVerbosityLevel)?: number;
```

but perhaps they work now? https://onnxruntime.ai/docs/api/js/interfaces/InferenceSession.SessionOptions.html

---

We can then set a global default logging behaviour, without needing to override global console methods.

nico-martin:
Unfortunately this does not work. Even with `logSeverityLevel:4, logVerbosityLevel:4` I still get this console.error
Screenshot 2025-12-02 at 09 58 15

I think it does make sense to set the `logSeverityLevel: 4`.

But for the `Unknown CPU vendor. cpuinfo_vendor value: 0` error, I had to dig deeper.
This is actually thrown here: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/common/cpuid_info_vendor.cc#L239
I think thats a) expected behaviour because when running in WebAssembly in a browser, the code doesn't have access to CPU vendor information and b) not an issue at all.

This warning actually bypasses the session-specific settings.
But after a lot of further digging in the onnxrunntime I found out there is a `ONNX_WEB.env.logLevel`. So when this is set to `fatal`, it wont show up in the console :)

nico-martin:
My solution now sets the default logging level to 4 (fatal). But it can be overwritten with `session_options.logSeverityLevel`

```javascript
const segmenter = await pipeline("background-removal", "Xenova/modnet", {
  device: "webgpu",
  session_options: {
    logSeverityLevel: 2, // warning
  },
});
```

Tis will not only set the onnxrunntime session_options, but also the `ONNX_WEB.env.logLevel`

src/backends/onnx.js
```diff
@@ -149,6 +149,21 @@ let webInitChain = Promise.resolve();
  * @returns {Promise<import('onnxruntime-common').InferenceSession & { config: Object}>} The ONNX inference session.
  */
 export async function createInferenceSession(buffer_or_path, session_options, session_config) {
+
+    /** @type {Array<'verbose' | 'info' | 'warning' | 'error' | 'fatal'>} */
+    const LOG_LEVELS = ['verbose', 'info', 'warning', 'error', 'fatal'];
+    /** @type 0|1|2|3|4 */
```
xenova:
```suggestion
    /** @type {0|1|2|3|4} */
```

src/backends/onnx.js
```diff
@@ -149,6 +149,21 @@ let webInitChain = Promise.resolve();
...
+    const logSeverityLevel =
+        typeof session_options.logSeverityLevel !== 'number' ||
+        session_options.logSeverityLevel < 0 ||
+        session_options.logSeverityLevel > 4
+            ? 4
+            : session_options.logSeverityLevel;
+
+    ONNX_WEB.env.logLevel = LOG_LEVELS[logSeverityLevel];
+
+    session_options = { ...session_options, logSeverityLevel };
```
xenova, line 166: I think we can move the `ONNX_WEB.env.logLevel` logic to happen during init (similar to how we set default WASM paths (as a good default).

This should hopefully minimize any session_options specific code a user would need to set.

src/models.js
```diff
@@ -241,6 +241,8 @@ async function getSession(pretrained_model_name_or_path, fileName, options, is_d
 
     // Overwrite `executionProviders` if not specified
     session_options.executionProviders ??= executionProviders;
+    // Set `logSeverityLevel` to 4 (fatal) if not specified
+    session_options.logSeverityLevel ??= 4;
```
xenova, line 245: this should be good. From my testing, a value of 3 will hide
```
2025-12-03 00:44:48.937 node[68874:20600323] 2025-12-03 00:44:48.937479 [W:onnxruntime:, session_state.cc:1316 VerifyEachNodeIsAssignedToAnEp] Some nodes were not assigned to the preferred execution providers which may or may not have an negative impact on performance. e.g. ORT explicitly assigns shape related ops to CPU to improve perf.
2025-12-03 00:44:48.937 node[68874:20600323] 2025-12-03 00:44:48.937528 [W:onnxruntime:, session_state.cc:1318 VerifyEachNodeIsAssignedToAnEp] Rerunning with verbose output on a non-minimal build will show node assignments.
```

but as you say in https://github.com/huggingface/transformers.js/pull/1468#issuecomment-3601187542, a value of 4 is necessary to hide everything else when running in browser.

src/backends/onnx.js
xenova: can move constant to top of file

src/backends/onnx.js
nico-martin: I actually think it makes sense to set the `ONNX_WEB.env.logLevel` here where I create a new session. Als a transformers.js user I dont really care if the log is a "onnxruntime session log" or a "onnxruntime general log". All I want to do is to create a pipeline and then be able to set the logLevel.
My solutions basically abstracts this decision away by only providing one setting and thats basically the settings that has always been available in the `session_options` (although very view people actually used that setting in the past I guess).

xenova: Sure, I agree with the point you made about a user not caring what kind of log is 👍 

I guess one thing that this solution will prevent is a user modifying the logLevel themselves (which they can do with `env.backends.onnx.env.logLevel = ...`), because it will be overridden each time. Ideally, a default will be set once, and then a user can update these on a per-session (logSeverityLevel) or per-runtime (env.logLevel) basis.

xenova:
Thanks again!

src/backends/onnx.js
xenova: maybe another option is to set a transformers.js-level logLevel, which would set both env and session-level

PR #1471, opened 2025-12-01
[v4] added wasm cache

nico-martin:
This adds caching of the wasm Binary.
I also added an env.cacheKey so developers can modify the cacheKey. By default it will be `transformers-cache` but they are free to set something related to their app.

Note: this will only cache the wasm file. So there will still be a request to https://cdn.jsdelivr.net/ for the mjs file. In my opinion, this should be cached via the service-worker if the applications requires full offline support.

### DIFF
--- a/src/backends/utils/cacheWasm.js
+++ b/src/backends/utils/cacheWasm.js
@@ -7,42 +7,41 @@ import { getCache } from '../../utils/cache.js';
  */
 async function loadAndCacheFile(url) {
     const fileName = url.split('/').pop();
+
+    /** @type {import('../../utils/cache.js').CacheInterface|undefined} */
+    let cache;
     try {
-        const cache = await getCache();
+        cache = await getCache();
 
         // Try to get from cache first
         if (cache) {
-            try {
-                const result = await cache.match(url);
-                if (result) {
-                    return result;
-                }
-            } catch (e) {
-                console.warn(`Error reading ${fileName} from cache:`, e);
+            const result = await cache.match(url);
+            if (result) {
+                return result;
             }
         }
-
-        // If not in cache, fetch it
-        const response = await fetch(url);
-
-        if (!response.ok) {
-            throw new Error(`Failed to fetch ${fileName}: ${response.status} ${response.statusText}`);
-        }
-
-        // Cache the response for future use
-        if (cache) {
-            try {
-                await cache.put(url, response.clone());
-            } catch (e) {
-                console.warn(`Failed to cache ${fileName}:`, e);
-            }
-        }
-
-        return response;
     } catch (error) {
-        console.warn(`Failed to load ${fileName}:`, error);
-        return null;
+        console.warn(`Failed to load ${fileName} from cache:`, error);
     }
+
+    // If not in cache, fetch it
+    const response = await fetch(url);
+
+    if (!response.ok) {
+        throw new Error(`Failed to fetch ${fileName}: ${response.status} ${response.statusText}`);
+    }
+
+    // Cache the response for future use
+    if (cache) {
+        try {
+            await cache.put(url, response.clone());
+        } catch (e) {
+            console.warn(`Failed to cache ${fileName}:`, e);
+        }
+    }
+
+    return response;
+
 }
 
 /**


xenova:
> So there will still be a request to https://cdn.jsdelivr.net/ for the mjs file. In my opinion, this should be cached via the service-worker if the applications requires full offline support.

Indeed, caching the mjs file would be necessary to ensure full offline support, and imo is essential before adding a caching feature like this PR proposes. Any ideas for how we could do this? Perhaps bundling `ort-wasm-simd-threaded.jsep.mjs` with `transformers.js` file?

xenova (commented):
Very nice + clean PR! Now we just need to figure out how to completely remove the ort-wasm-simd-threaded.jsep.mjs dependency.

nico-martin:
So after another deep-dive into onnxruntime I figured out that it's actually no problem at all to load the wasm factory (.mjs) as a blob, which allows us to load it from cache.
The only problem is that inside the factory there are URLs that try to resolve relative with "import.meta.url". If we just replace this with the actual baseURL ist works just fine. 
https://github.com/huggingface/transformers.js/blob/v4-cache-wasm-file/src/backends/utils/cacheWasm.js#L76

On to of that I also did some refactoring of the hub.js. My goal is to keep large files that only have a handfull of exported methods as clean as possible by extracting some heloper functions and constants into their separate files.

I also wanted to improve the caching (which is now used not only in the hub.js but also in the backends/onnx.js) so I created a helper function also also an interface "CacheInterface" that any given cache has to implement.

src/backends/onnx.js
```diff
@@ -141,6 +142,85 @@ const IS_WEB_ENV = apis.IS_BROWSER_ENV || apis.IS_WEBWORKER_ENV;
...
+
+/**
+ * Ensures the WASM binary is loaded and cached before creating an inference session.
+ * Only runs once, even if called multiple times.
+ *
+ * Note: This caches the WASM binary via the wasmBinary option, which allows it to work offline.
+ * However, the MJS loader file still needs to be fetched from the network (or bundled).
+ * For full offline support including the MJS file, you need to either:
+ * 1. Use a Service Worker to cache all network requests
+ * 2. Bundle onnxruntime-web with your application
```
xenova: We can update this, right?

src/env.js
```diff
@@ -152,9 +152,12 @@ const localModelPath = RUNNING_LOCALLY ? path.join(dirname__, DEFAULT_LOCAL_MODE
  * @property {boolean} useFSCache Whether to use the file system to cache files. By default, it is `true` if available.
  * @property {string|null} cacheDir The directory to use for caching files with the file system. By default, it is `./.cache`.
  * @property {boolean} useCustomCache Whether to use a custom cache system (defined by `customCache`), defaults to `false`.
- * @property {Object|null} customCache The custom cache to use. Defaults to `null`. Note: this must be an object which
+ * @property {import('./utils/cache.js').CacheInterface|null} customCache The custom cache to use. Defaults to `null`. Note: this must be an object which
  * implements the `match` and `put` functions of the Web Cache API. For more information, see https://developer.mozilla.org/en-US/docs/Web/API/Cache.
- * If you wish, you may also return a `Promise<string>` from the `match` function if you'd like to use a file path instead of `Promise<Response>`.
```
xenova, line 157: we allow string/path retrieval too, as this is more efficient with onnxruntime-node (specify path instead of buffer/response)

src/env.js
```diff
@@ -185,6 +188,9 @@ export const env = {
 
     useCustomCache: false,
     customCache: null,
+
+    useWasmCache: IS_WEB_CACHE_AVAILABLE || IS_FS_AVAILABLE,
```
xenova, line 192: wouldn't this also match the onnxruntime-node case? would we download files unnecessarily?

nico-martin: Yes. Haven't tested it yet but if onnxruntime-node also uses the `ONNX_ENV.wasm.wasmBinary` and `ONNX_ENV.wasm.wasmPaths` caching should work there too.

src/env.js
nico-martin: True. I will fix this

src/backends/utils/cacheWasm.js
```diff
@@ -0,0 +1,83 @@
+import { getCache } from '../../utils/cache.js';
+
+/**
+ * Loads and caches a file from the given URL.
+ * @param {string} url The URL of the file to load.
+ * @returns {Promise<Response|import('../../utils/hub/FileResponse.js').default|null|string>} The response object, or null if loading failed.
+ */
+async function loadAndCacheFile(url) {
+    const fileName = url.split('/').pop();
```
xenova, line 9: is this robust enough?

src/utils/hub.js
```diff
@@ -585,14 +253,15 @@ export async function getModelFile(path_or_repo_id, filename, fatal = true, opti
         /** @type {Uint8Array} */
         let buffer;
 
-        if (!options.progress_callback) {
+        if (!options.progress_callback && typeof response !== 'string') {
```
xenova: we do the `typeof response !== 'string'` check multiple times here... and I imagine we can wrap the whole block in this check?

src/backends/utils/cacheWasm.js
nico-martin: I think so. Also if not its no big deal since the fileName is only used in the Error/console messages.

src/utils/hub.js
nico-martin: yes, definitely!

nico-martin:
I like the CacheRegistry! The only problem is that we normally don't know upfront all the files a model will load/expect (altough I think it would be great to add that as well).
I think that opens up a completely new topic. I will keep this on my radar but not implement it here in the PR.

src/backends/utils/cacheWasm.js
```diff
@@ -0,0 +1,83 @@
...
+    try {
+        const cache = await getCache();
+
+        // Try to get from cache first
+        if (cache) {
+            try {
+                return await cache.match(url);
+            } catch (e) {
+                console.warn(`Error reading ${fileName} from cache:`, e);
```
xenova: if `await cache.match(url)` returns undefined (i.e., it is not in cache), then we return `undefined` from this function... and the fetch below is never called (meaning, it is never cached).

xenova:
```suggestion
        if (cache) {
            try {
                const result = await cache.match(url);
                if (result) {
                    return result;
                }
            } catch (e) {
                console.warn(`Error reading ${fileName} from cache:`, e);
```

seems to fix it

xenova: made this change.

xenova (approved):
Thanks! 🚀  Had time today to finish the review, and it works well! I had to make one small adjustment (only return when `await cache.match(url)` matches... not when `undefined`), and it's good to go :)

I also like the refactor out of the monolithic hub.js file.

I tested it on the recent chatterbox webgpu demo, and it now runs fully offline thanks to this PR! 🔥

xmcp:
I encountered the below exception after this PR. It was fine at the previous commit (aab23267eb13bcb85460c6ce8db44b36b0250f50).

image

The inner (caught) exception is:

```
TypeError: Invalid base URL
    at blob:https://localhost:5173/49689b73-721b-4544-9d07-d0402577925e:100:240
    at ortWasmThreaded (blob:https://localhost:5173/49689b73-721b-4544-9d07-d0402577925e:100:478)
    at https://localhost:5173/@fs/.../transformers.js/dist/transformers.web.js:2628:28592
    at new Promise (<anonymous>)
    at Xt (https://localhost:5173/@fs/.../transformers.js/dist/transformers.web.js:2628:28380)
    at async Ns (https://localhost:5173/@fs/.../transformers.js/dist/transformers.web.js:2630:28936)
    at async yr.init (https://localhost:5173/@fs/.../transformers.js/dist/transformers.web.js:2630:33169)
    at async ec (https://localhost:5173/@fs/.../transformers.js/dist/transformers.web.js:2624:1454)
    at async La (https://localhost:5173/@fs/.../transformers.js/dist/transformers.web.js:2624:1733)
    at async a.create (https://localhost:5173/@fs/.../transformers.js/dist/transformers.web.js:2624:19895)
```

The `blob:...:100:240` (the first line of the stack trace) points to this statement:

```
Va ??= g.locateFile ? g.locateFile ? g.locateFile("ort-wasm-simd-threaded.asyncify.wasm", ja) : ja + "ort-wasm-simd-threaded.asyncify.wasm" : (new URL("/@fs/.../transformers.js/dist/ort-wasm-simd-threaded.asyncify.wasm",import.meta.url)).href;
```

Seems that we still have remaining `import.meta.url`s in the imported mjs file.

xmcp:
Btw I managed to implement mjs&wasm cache in this way, which works fine with previous versions (both aab23267eb13bcb85460c6ce8db44b36b0250f50 and v3 from npm). I'm not sure why it does not encounter the above exception.

```javascript
import {pipeline, env} from "@huggingface/transformers";

import onnx_wasm from "../node_modules/@huggingface/transformers/dist/ort-wasm-simd-threaded.asyncify.wasm?url";
import onnx_mjs from "../node_modules/@huggingface/transformers/dist/ort-wasm-simd-threaded.asyncify.mjs?url";

async function init() {
    let cache = await caches.open('transformers-cache');

    let wasm_file = await cache.match(onnx_wasm);
    if(!wasm_file) {
        await cache.add(onnx_wasm);
        wasm_file = await cache.match(onnx_wasm);
    }

    let mjs_file = await cache.match(onnx_mjs);
    if(!mjs_file) {
        await cache.add(onnx_mjs);
        mjs_file = await cache.match(onnx_mjs);
    }

    env.localModelPath = '/models';
    env.allowRemoteModels = false;
    env.allowLocalModels = true;
    env.backends.onnx.wasm.wasmPaths = {
        wasm: URL.createObjectURL(await wasm_file.blob()),
        mjs: URL.createObjectURL(await mjs_file.blob()),
    };

    return await pipeline(
        ...
    );
}
```

nico-martin:
Hi @xmcp
Thanks for reaching out!

> Btw I managed to implement mjs&wasm cache in this way, which works fine with previous versions (both [aab2326](https://github.com/huggingface/transformers.js/commit/aab23267eb13bcb85460c6ce8db44b36b0250f50) and v3 from npm). I'm not sure why it does not encounter the above exception.

In your set up that makes total sense. The import statement makes sure the .wasm and the .mjs files are copied to your dist/ folder and onnx_wasm and onnx_mjs are then links to those files.
You then use the caches api to cache the files and pass them as an internal reference URL to `env.backends.onnx.wasm.wasmPaths`. So far so good.

Hardcoding the wasmPaths is ok, but there could be one issue. We only use the asyncify version for non-iOS devices because we experienced some issues in the past (maybe @xenova could elaborate on this).
https://github.com/huggingface/transformers.js/blob/v4/src/backends/onnx.js#L285-L293
In your example, all clients will us the asyncify version.

But back to your main issue: I think the problem is that now your code tries to cache the request, but then our code tries to cache it again. And now the caches API tries to cache a blob URL and I am not sure if that works :D.
Could you try to set `env.useWasmCache = false` right where you set all the other env properties?

nico-martin:
I also created a little check for that case. @xmcp could you verify if that solves your problem?
https://github.com/huggingface/transformers.js/compare/v4...v4-cache-wasm-file-blob-fix?expand=1

xmcp:
I can confirm that setting `env.useWasmCache = false;` solves the exception. The patch at `v4-cache-wasm-file-blob-fix` does not fix the exception if I pass normal URLs (not the blob one) to `env.backends.onnx.wasm.wasmPaths`. I will try to diagnose this.

xmcp:
I think I know the problem. In my code snippet, `onnx_wasm` and `onnx_mjs` are relative URLs (`/assets/.../xxx.mjs` in prod or `/@fs/.../xxx.mjs` in dev). Therefore, the `loadWasmFactory` function replaced all `import.meta.url`s to relative path, which does not work for the second argument of a URL constructor.

Making the baseUrl absolute solves the problem. Here is a patch for that:

```diff
--- a/src/backends/utils/cacheWasm.js
+++ b/src/backends/utils/cacheWasm.js
@@ -74,7 +74,7 @@ export async function loadWasmFactory(libURL) {
     try {
         let code = await response.text();
         // Fix relative paths when loading factory from blob, overwrite import.meta.url with actual baseURL
-        const baseUrl = libURL.split('/').slice(0, -1).join('/');
+        const baseUrl = new URL(libURL.split('/').slice(0, -1).join('/'), location.href).href;
         code = code.replace(/import\.meta\.url/g, `"${baseUrl}"`);
         const blob = new Blob([code], { type: 'text/javascript' });
         return URL.createObjectURL(blob);
```

(As a bonus, it also urlencodes the baseUrl correctly, so it will no longer crash the program if libURL contains double quotes.)

nico-martin:
Thanks for taking the time to investigate!
But I think that will break on the server (NodeJS, Bun, Deno). I'll dig a bit deeper and come back to you.

nico-martin:
@xmcp 
could you try this?
https://github.com/huggingface/transformers.js/compare/v4...v4-cache-wasm-file-blob-fix?expand=1

xmcp:
Yes this version should work. Btw the `isBlobURL(url) || isRemoteURL(url)` check seems redundant because the URL class works fine for absolute URLs.

image

nico-martin:
Good catch! I removed it.PR is open :)
https://github.com/huggingface/transformers.js/pull/1489

xenova:
PR is merged ✅

PR #1478, opened 2025-12-05
Adapt script module for optimum-onnx

Disk-MTH:
Export script:
- Update requirements.txt
- Adapt to new numpy and onnxruntime versions


xenova:
No longer used in https://github.com/huggingface/transformers.js/pull/1382; instead, we recommend using optimum-onnx directly (or onnxruntime-genai for LLMs)

PR #1480, opened 2025-12-12
Add support for WhisperForAudioClassification

xenova:
Which means we can run https://huggingface.co/pipecat-ai/smart-turn-v3 in the browser, a new voice activity detection model 🥳


P0u4a:
@xenova This would be awesome. Is anything (other than the merge conflict) holding it back?

PR #1481, opened 2025-12-12
Fix fp16/q4f16 dtype test failures on Node.js 20+

neonwatty:
## Problem

Preparing a different PR I noticed that running tests locally on Node.js 20+ causes the Llama fp16/q4f16 dtype tests to fail:

```
TypeError: Tensor.data must be a typed array (4) for float16 tensors, but got typed array (0).
```

## Cause

`onnxruntime-common` detects `Float16Array` (available in Node 20+) and uses it for float16 tensors. However, `onnxruntime-node`'s native binding was written before `Float16Array` existed and only accepts `Uint16Array`.

## Fix

This PR adds a workaround that hides `Float16Array` before onnxruntime loads, forcing it to use `Uint16Array` instead.

There's an upstream PR to add `Float16Array` support to onnxruntime-node (https://github.com/microsoft/onnxruntime/pull/26742), but it hasn't merged yet (and onnxruntime PRs can take many months to merge).  Once it does, this workaround can be removed.

## Test plan

- [x] `npm test` passes locally on Node.js 20+
- [x] fp16/q4f16 dtype tests now pass

### DIFF
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -128,7 +128,7 @@ export default {
   // runner: "jest-runner",
 
   // The paths to modules that run some code to configure or set up the testing environment before each test
-  // setupFiles: [],
+  setupFiles: ["./tests/setup.js"],
 
   // A list of paths to modules that run some code to configure or set up the testing framework before each test
   // setupFilesAfterEnv: [],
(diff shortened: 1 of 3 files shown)


xenova:
Hi 👋 thanks for the PR. Preferably, we'd be able to fix this for normal running, not just unit tests. Any ideas? It seems like any call to `new Tensor(...)` for type=float16 produces `Float16Array` values, even if the the data passed in is Uint16Array.

xenova:
I did a bit of digging and https://github.com/huggingface/transformers.js/pull/1482 seems to fix it. How does that look?

neonwatty:
Perfection -  #1482 is a better fix - closing this one.

PR #1486, opened 2025-12-15
Fix word-level timestamp overflow in Whisper chunked transcription

### LINKED ISSUE #1357: whisper-large-v3-turbo_timestamped has broken timestamps
### System Info

transformers.js version 3.6.1

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension

### Description

Running `whisper-large-v3-turbo_timestamped` produces broken timestamps.

### Reproduction

Run the following code (takes ~30 seconds) and follow console log:

```html
<script type="module">
const { env, pipeline } = await import(`https://cdn.jsdelivr.net/npm/@huggingface/transformers@3.6.1/dist/transformers.min.js`);
env.allowLocalModels = false;

const buffer = await (await fetch("tos.pcm")).arrayBuffer();
const audio = new Float32Array(buffer);

const pipe = await pipeline("automatic-speech-recognition",
	"onnx-community/whisper-large-v3-turbo_timestamped",
	{dtype:"fp16", device:"webgpu"});

	chunk_length_s: 30,
	stride_length_s: 5,
	return_timestamps: "word",
	language: "en"});

console.log(result.chunks.map(chunk => `${chunk.timestamp[0]} -> ${chunk.timestamp[1]} ${chunk.text}`))
</script>
```

Input is 60 second .pcm file, the console prints:

```
...
66: "40.82 -> 40.98  by"
67: "40.98 -> 41.1  these"
68: "41.1 -> 41.44  giant"
69: "41.44 -> 42.02  robotic"
70: "42.02 -> 69.98  claws--"
71: "69.98 -> 69.98  Oh,"
72: "69.98 -> 69.98  whatever,"
73: "69.98 -> 69.98  we're"
74: "69.98 -> 69.98  done!"
75: "69.98 -> 69.98  We're"
76: "69.98 -> 69.98  done!"
77: "69.98 -> 69.98  Robot's"
78: "69.98 -> 69.98  memory"
79: "69.98 -> 69.98  synced"
80: "69.98 -> 69.98  and"
81: "69.98 -> 69.98  locked."
```

There are two problems here:
1. timestamps are not precise
2. timestamps goes beyond the source duration

Attached the tos.pcm as .zip

[tos.zip](https://github.com/user-attachments/files/21099492/tos.zip)

neonwatty:
This PR aims to fix #1357

## The Problem

#1357 reports a problem I've had in usage as well - that when using `chunk_length_s: 30` with `return_timestamps: "word"`, timestamps can exceed the actual audio duration.  For example, a 60s audio file produces timestamps up to 69.98s.

*Root cause:** Digging in here's what I found: the model outputs timestamps up to ~29.98s (the maximum representable value given `time_precision = 30/1500 = 0.02`).  For final chunks shorter than 30s, these raw timestamps are added to the accumulated `time_offset`, causing overflow.

## Proposed Solution

I think this simple solution works: clamp word-level timestamps to the actual `chunk_len` from stride metadata before applying `time_offset`:

```javascript
if (current_chunk_len !== null) {
    raw_start = Math.min(raw_start, current_chunk_len);
    raw_end = Math.min(raw_end, current_chunk_len);
}
```

## Why not match Python's approach?

There's a similar reported issue upstream with Python transformers; this fix crops cross-attention matrices before DTW alignment ([PR #25607](https://github.com/huggingface/transformers/pull/25607)). 

To my knowledge, here in JS we receive pre-computed `token_timestamps` from ONNX models, so we cannot modify the DTW computation. 

Hence clamping at the tokenizer level seems to be the appropriate fix here.

## Testing

- Added unit test for issue #1357
- Validated edge cases: exactly 30s, <30s, and long multi-chunk audio
- All 1891 existing tests pass
- Segment-level timestamps (`return_timestamps: true`) unaffected

### DIFF
--- a/src/tokenizers.js
+++ b/src/tokenizers.js
@@ -3690,8 +3690,14 @@ export class WhisperTokenizer extends PreTrainedTokenizer {
             let last_timestamp = null;
             let first_timestamp = timestamp_begin;
 
+            // Track chunk length for clamping token timestamps (issue #1357)
+            // This prevents timestamps from exceeding the actual audio duration
+            // when the model outputs timestamps near the 30s boundary
+            let current_chunk_len = null;
+
             if ("stride" in output) {
                 const [chunk_len, stride_left, stride_right] = output.stride;
+                current_chunk_len = chunk_len;
 
                 // Offset the timings to account for the other `model_outputs`.
                 time_offset -= stride_left;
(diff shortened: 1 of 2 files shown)


xenova:
Superseded by https://github.com/huggingface/transformers.js/pull/1594.

PR #1489, opened 2025-12-18
V4 cache wasm file blob fix

nico-martin:
fixes an issue described in the conversation here:
https://github.com/huggingface/transformers.js/pull/1471

### DIFF
--- a/src/backends/utils/cacheWasm.js
+++ b/src/backends/utils/cacheWasm.js
@@ -1,4 +1,5 @@
 import { getCache } from '../../utils/cache.js';
+import { isValidUrl } from '../../utils/hub/utils.js';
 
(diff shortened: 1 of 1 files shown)


xenova (commented):
Thanks! Could we use `isValidUrl`? https://github.com/huggingface/transformers.js/blob/b125e82b86f62cf9f0f77217601e04a5ff7c6e7f/src/utils/hub.js#L177 for this?

Example usage:
https://github.com/huggingface/transformers.js/blob/b125e82b86f62cf9f0f77217601e04a5ff7c6e7f/src/utils/hub.js#L217

nico-martin:
Yes, you're right. I changed it!

PR #1497, opened 2026-01-07
fixing text color to display token ids correctly

DecodersLord:
Current tokenizer example displays white text over white background while displaying the token-ids which is difficult to see. (need to highlight the text inside textbox)

Screenshot 2026-01-07 220705

Proposed fix is to change text color to black to see it better.

Screenshot 2026-01-07 220407

### DIFF
--- a/examples/tokenizer-playground/src/App.jsx
+++ b/examples/tokenizer-playground/src/App.jsx
@@ -146,7 +146,7 @@ function App() {
         </div>
       </div>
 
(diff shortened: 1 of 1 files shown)


DecodersLord:
@xenova Please have a look. I took the liberty to submit a PR for this fairly small and simple bug I came across.

nico-martin:
Hi @DecodersLord, with v4 we moved the -examples out of the main repo into its own repo.
Could you apply your fix here?
https://github.com/huggingface/transformers.js-examples/tree/main/the-tokenizer-playground

DecodersLord:
@nico-martin Thank you for the update. 

The example repo seems to be displaying the tokens with correct coloring. Let me know if I can contribute to some other existing issues.

PR #1498, opened 2026-01-09
[v4] Refactor models.js into separate per-model modeling files

nico-martin:


src/models/_base/pre-trained-model.js
xenova, line 1: we can rename to `modeling_utils.js`
https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L1094

src/models/_base/pre-trained-model.js
```diff
@@ -0,0 +1,1277 @@
...
+    MultiModality: 8,
+    Phi3V: 9,
+    AudioTextToText: 10,
+    AutoEncoder: 11,
+    ImageAudioTextToText: 12,
+    Supertonic: 13,
+    Chatterbox: 14,
+};
+
+const MODEL_TYPE_CONFIG = {
```
xenova, line 101: useful refactor, thanks!

src/models/_base/output.js
xenova, line 1: rename to `modeling_outputs.js`
from https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_outputs.py

src/models/_base/pre-trained-model.js
```diff
@@ -0,0 +1,1277 @@
...
+    TemperatureLogitsWarper,
+    ClassifierFreeGuidanceLogitsProcessor,
+} from '../../generation/logits_process.js';
+import { GenerationConfig } from '../../generation/configuration_utils.js';
+import { EosTokenCriteria, MaxLengthCriteria, StoppingCriteriaList } from '../../generation/stopping_criteria.js';
+import { LogitsSampler } from '../../generation/logits_sampler.js';
+import { pick } from '../../utils/core.js';
+import { ModelOutput } from './output.js';
+
+export const MODEL_TYPES = {
```
xenova, line 83: do we need to export `MODEL_TYPES`? Ideally, the only thing export from this file (and accessible by users) should be `PreTrainedModel`

src/transformers.js
```diff
@@ -25,11 +25,11 @@ export * from './utils/tensor.js';
 export * from './utils/maths.js';
 
 export { FeatureExtractor } from './base/feature_extraction_utils.js';
-export * from './models/feature_extractors.js';
+export * from './models/feature-extractors.js';
 export * from './models/auto/feature_extraction_auto.js';
 
 export { ImageProcessor } from './base/image_processors_utils.js';
-export * from './models/image_processors.js';
+export * from './models/image-processors.js';
```
xenova: ideally kept with underscores, to align with python lib. Also, if we were going to change to hyphens (probably more used in JS ecosystem), we'd need to change it everywhere else.

src/models/modeling_utils.js
xenova, line 1: duplicate of src/models/_base/pre-trained-model.js

src/models/albert/modeling_albert.js
xenova, line 1: very clean!

src/models/auto/feature_extraction_auto.js
```diff
@@ -1,7 +1,7 @@
 import { FEATURE_EXTRACTOR_NAME, GITHUB_ISSUE_URL } from '../../utils/constants.js';
 import { getModelJSON } from '../../utils/hub.js';
 import { FeatureExtractor } from '../../base/feature_extraction_utils.js';
-import * as AllFeatureExtractors from '../feature_extractors.js';
+import * as AllFeatureExtractors from '../feature-extractors.js';
```
xenova: as mentioned elsewhere, we can keep using underscores here

src/models/bit/index.js
```diff
@@ -0,0 +1 @@
+export * from './image_processing_bit.js';
```
xenova: missing file

src/models/chinese_clip/index.js
```diff
@@ -0,0 +1,2 @@
+export * from './modeling_chinese_clip.js';
+export * from './image_processing_chinese_clip.js';
```
xenova, line 2: missing file

src/models/clap/index.js
```diff
@@ -0,0 +1,2 @@
+export * from './modeling_clap.js';
+export * from './feature_extraction_clap.js';
```
xenova: missing file

src/models/convnext/index.js
```diff
@@ -0,0 +1,2 @@
+export * from './modeling_convnext.js';
+export * from './image_processing_convnext.js';
```
xenova: missing file

src/models/dinov3_vit/index.js
```diff
@@ -0,0 +1,2 @@
+export * from './modeling_dinov3_vit.js';
+export * from './image_processing_dinov3_vit.js';
```
xenova: missing file

src/models/ernie4_5/modeling_ernie4_5.js
```diff
@@ -0,0 +1,7 @@
+import { PreTrainedModel } from '../_base/pre-trained-model.js';
+
+export class Ernie4_5_PretrainedModel extends PreTrainedModel {}
+
+export class Ernie4_5_Model extends Ernie4_5_PretrainedModel {}
+
+export class Ernie4_5_ForCausalLM extends Ernie4_5_PretrainedModel {}
```
xenova:
```suggestion
export class Ernie4_5PretrainedModel extends PreTrainedModel {}

export class Ernie4_5Model extends Ernie4_5PretrainedModel {}

export class Ernie4_5ForCausalLM extends Ernie4_5PretrainedModel {}
```

This was a change that the python team made just before release, causing a mismatch with us (whoops). So, we will use the latest correct name. Not a big issue because not many users of this model via the class name (mostly pipeline).
https://github.com/huggingface/transformers/blob/7a2bf25f36264f3f5409d2743722e4f6b3504e86/src/transformers/models/ernie4_5/modeling_ernie4_5.py#L337

src/models/janus/index.js
```diff
@@ -0,0 +1,2 @@
+export * from './image_processing_janus.js';
+export * from './processing_janus.js';
```
xenova: missing files

src/models/registry.js
xenova, line 1: very nice! This solves a circular import issue I faced when trying to do this refactor a while ago.

src/models/feature-extractors.js
xenova, line 1: rather keep the underscore (to align with the python lib, and also to ensure consistency with current standard)

src/models/image_processors.js
```diff
@@ -32,6 +32,7 @@ export * from './rt_detr/image_processing_rt_detr.js';
 export * from './sam/image_processing_sam.js';
 export * from './sam2/image_processing_sam2.js';
 export * from './sam3/image_processing_sam3.js';
+export * from './sapiens/image_processing_sapiens.js';
```
xenova, line 35: good catch!

src/models/image_processors.js
```diff
@@ -40,3 +41,4 @@ export * from './vit/image_processing_vit.js';
 export * from './vitmatte/image_processing_vitmatte.js';
 export * from './vitpose/image_processing_vitpose.js';
 export * from './yolos/image_processing_yolos.js';
+export * from './auto/image_processing_auto.js';
```
xenova: do we need to export the auto image processor here? I think it's better to keep in a different file (because the AutoImageProcessor isn't a ImageProcessor like the others, it's a class which returns an ImageProcessor from the `from_pretrained` function.

src/models/modeling_utils.js
xenova: can remove most likely.

src/models/modeling_outputs.js
xenova, line 1: duplicate of src/models/_base/output.js

src/models/model-mapping-names.js
```diff
@@ -0,0 +1,79 @@
+/**
+ * @file Model mapping names without class references to avoid circular dependencies.
```
xenova: ah nice 👍. I assume we do a lookup later when using these mappings. That works for me. For reference, the reason why we had to create a mapping of "class name string" to "class name" was because minifiers would end up changing the name of the class, causing the exports to be renamed to random names.

This solution you have proposed shouldn't have this problem.

src/models/model-mapping-names.js
xenova, line 1: we're missing a lot of `MODEL_FOR_*` definitions here. Is it perhaps split into another file?

src/models/model-mapping-names.js
xenova: oh wait, I see... this is just a subset for the model validation. Hopefully we can simplify this and move into the same file as the rest? May be annoying to maintain two parallel lists.

https://github.com/huggingface/transformers.js/blob/b125e82b86f62cf9f0f77217601e04a5ff7c6e7f/src/models.js#L7929-L8456

Perhaps the solution is to just create a string to string mapping, and do the class lookup later?

src/models/utils.js
xenova, line 1: since these are all modeling utils, perhaps we can move to the (new) modeling_utils.js file? Not sure if it will cause circular dependency issues though. Hopefully not.

src/models/session.js
xenova, line 1: very nice to separate this out too. thanks!

src/models.js
xenova, line 1: *chef's kiss*

Image

src/utils/tensor.js
```diff
@@ -1583,3 +1583,64 @@ export function quantize_embeddings(tensor, precision) {
...
+
+/**
+ * Creates a boolean tensor with a single value.
+ * @param {boolean} value The value of the tensor.
+ * @returns {Tensor} The boolean tensor.
+ * @private
+ */
+export function boolTensor(value) {
+    return new Tensor('bool', [value], [1]);
+}
```
xenova: this is more of a helper function that we don't really want to expose to users in global import scope. Let's make sure we don't.

xenova: the tensor functions in the global score are closely aligned with the naming of pytorch functions (e.g., `ones_like`, `full`), but it does clutter things up a lot. This will definitely be a great candidate for subpaths. `import { ones } from @huggingface/transformers/tensor` maybe? Or alternatively, just create static functions like `Tensor.ones`. Not too sure. Not important for this refactor.

src/utils/model-loader.js
xenova, line 1: another useful refactor. thanks!

xenova (changes requested):
First pass done! ✅ This will be a super impactful refactor, making model addition and maintenance far easier... (models.js was getting far too unmanageable).

src/models/_base/pre-trained-model.js
nico-martin: We also need them in src/models/registry.js. But since we only export them from this file but not from src/transformers.js, they should not be accessible by users..

src/models/chinese_clip/index.js
nico-martin: The file should be there: https://github.com/huggingface/transformers.js/blob/v4-models-js-3/src/models/chinese_clip/image_processing_chinese_clip.js

src/models/utils.js
nico-martin: all good. no circular dependencies so far..

src/models/model-mapping-names.js
nico-martin: Most of the MODEL_FOR_ mappings are in the src/models/registry.js. Because the circular dependency problem. But let me find a better solution..

src/models/image_processors.js
nico-martin: you're right. Thats not needed

src/models/chinese_clip/index.js
xenova: oh whoops yes my bad -- this was a comment I thought I had deleted 🥷

kungfooman:
For anyone having trouble with `npm i` for:

```
    "onnxruntime-node": "1.24.0-dev.20251104-75d35474d5",
    "onnxruntime-web": "1.24.0-dev.20251104-75d35474d5",
```

like:

```
npm error code 1
npm error path /var/www/transformers.js/node_modules/onnxruntime-node
npm error command failed
npm error command sh -c node ./script/install
npm error /var/www/transformers.js/node_modules/onnxruntime-node/script/install-utils.js:129
npm error       throw new Error(`Failed to find specific package versions for ${packageName} in ${index}`);
npm error             ^
npm error
npm error Error: Failed to find specific package versions for Microsoft.ML.OnnxRuntime.Gpu.Linux in https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/nuget/v3/index.json
npm error     at nugetPackageUrlResolver (/var/www/transformers.js/node_modules/onnxruntime-node/script/install-utils.js:129:13)
npm error     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
npm error     at async resolvePackage (/var/www/transformers.js/node_modules/onnxruntime-node/script/install-utils.js:203:14)
npm error     at async installPackages (/var/www/transformers.js/node_modules/onnxruntime-node/script/install-utils.js:93:19)
npm error
npm error Node.js v18.20.4
```

You have to use: `npm i --onnxruntime-node-install-cuda=skip`

xenova (approved):
Just did a final sweep over the PR. Great stuff! 🥳

PR #1504, opened 2026-01-22
V4 move to pnpm workspaces

nico-martin:
This PR is a proposal to move to pnpm workspaces.

**why workspaces?**
Right we have only one package: `@huggingface/transformers`
So everything we ship can be used from this one package. In the future we also want to be able to ship small, domain specific packages. For example a framework specific package `@huggingface/transformers-react`or maybe abstractions like a text-generation package that is compatible with the Responses API `@huggingface/transformers-responses`.
Using workspaces will allow us to ship those small packages that depend on `@huggingface/transformers` without the overhead of individual repos.

**why pnpm?**
Pnpm is a package manager for NodeJS that brings a lot of benefits especially in how it handles disk space.
pnpm stores each package version once in a global store and hard-links it to all workspaces that need it, which avoids the disk space waste of npm duplicating packages everywhere. The symlink-based node_modules structure means packages can only import what they explicitly declare as dependencies, preventing the "works on my machine" issues where code accidentally relies on hoisted packages.
The workspace:* protocol links local packages during development and converts to real versions when you publish, and installs are faster because pnpm fetches packages in parallel and reuses everything from the global store. You just need a pnpm-workspace.yaml file to set it up.
Also pnpm seems to be the standard for workspaces in the npm ecosystem, since big repos like [Vercel AI SDK](https://github.com/vercel/ai), [langchainjs](https://github.com/langchain-ai/langchainjs) and others are using it.

### DIFF
--- a/packages/transformers/src/transformers.js
+++ b/packages/transformers/src/transformers.js
@@ -1,14 +1,22 @@
 /**
- * @file Entry point for the Transformers.js library.
- * Only the exports from this file are available to the end user.
+ * @file Entry point for the Transformers.js library. Only the exports from this file
+ * are available to the end user, and are grouped as follows:
+ *
+ * 1. [Environment variables](./env)
+ * 2. [Pipelines](./pipelines)
+ * 3. [Models](./models)
+ * 4. [Tokenizers](./tokenizers)
+ * 5. [Processors](./processors)
+ * 6. [Configs](./configs)
  *
  * @module transformers
  */
 
+// Environment variables
 export { env } from './env.js';
 
+// Pipelines
 export * from './pipelines.js';
-export { PretrainedConfig, AutoConfig } from './configs.js';
 
 // Models
 export * from './models/models.js';
@@ -29,6 +37,9 @@ export * from './models/auto/image_processing_auto.js';
 // Processors
 export * from './models/processors.js';
 export * from './models/auto/processing_auto.js';
+
+// Configs
+export { PretrainedConfig, AutoConfig } from './configs.js';
 
 // Additional exports
 export * from './generation/streamers.js';


package.json
```diff
@@ -1,88 +1,29 @@
 {
-  "name": "@huggingface/transformers",
+  "name": "@huggingface/transformers-monorepo",
   "version": "4.0.0-alpha.0",
```
xenova: can remove/set version to something else? Unless we are intentional about other releases (e.g., react-transformers) to keep the exact same version. In which case, a global version makes sense I guess.

packages/transformers/package.json
```diff
@@ -0,0 +1,89 @@
...
+  "author": "Hugging Face",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/huggingface/transformers.js/issues"
+  },
+  "homepage": "https://github.com/huggingface/transformers.js#readme",
+  "dependencies": {
+    "@huggingface/jinja": "^0.5.3",
+    "@huggingface/tokenizers": "^0.1.0",
+    "onnxruntime-common": "1.24.0-dev.20251104-75d35474d5",
```
xenova: is it necessary to add? onnxruntime-web and onnxruntime-node already have this as a dependency.

packages/transformers/src/pipelines/zero-shot-audio-classification.js
xenova, line 1: whoops! thanks!

.github/workflows/publish.yml
xenova, line 1: I'd definitely like to align the publishing workflow with how we do it in tokenizers.js and other huggingface.js libraries: https://github.com/huggingface/tokenizers.js/blob/main/.github/workflows/publish.yml

allowing per-package publishing via Actions tab is far simpler than the current approach. It would require a few more adjustments as some of the documentation needs updates of version numbers (e.g., jsdelivr cdn link). Can be an update we make down the line.

.github/workflows/tests.yml
xenova, line 1: I think we should define per-package github workflow tests. Kind of like https://github.com/huggingface/huggingface.js/tree/main/.github/workflows. WDYT?

packages/transformers/scripts/esbuild/build/plugins/postBuildPlugin.mjs
```diff
@@ -0,0 +1,58 @@
...
+        // 1. Remove unnecessary files
+        const file = path.join(distDir, ORT_BUNDLE_FILE);
+        if (existsSync(file)) unlinkSync(file);
+
+        // 2. Copy unbundled JSEP file
+        // In a workspace setup, node_modules might be at the monorepo root
+        const possiblePaths = [
+          path.join(rootDir, "node_modules/onnxruntime-web/dist"),
+          path.join(rootDir, "../../node_modules/onnxruntime-web/dist"),
+        ];
```
xenova: Is there a better/programmatic way to determine where the node_modules is located?

packages/transformers/scripts/esbuild/build/constants.mjs
```diff
@@ -20,7 +20,7 @@ export const OUT_DIR = path.join(ROOT_DIR, DIST_FOLDER);
 export const getEsbuildDevConfig = (rootDir) => ({
   bundle: true,
   treeShaking: true,
-  logLevel: "info",
+  logLevel: "silent",
```
xenova: any specific info that was too much for dev mode? I'd imagine it would still be useful to see these logs.

packages/transformers/src/models/vits/tokenization_vits.js
xenova, line 1: I think we can fix these issues by updating the tsconfig.json:
```diff
-    "module": "nodenext",
-    "moduleResolution": "nodenext",
+    "module": "esnext",
+    "moduleResolution": "bundler",
```

Afterwards, we can revert most of these changes.

packages/transformers/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+  // Only include files in the src directory
+  "include": ["src/**/*"],
+  "compilerOptions": {
+    // Tells the compiler to check JS files
+    "checkJs": true,
+    "target": "esnext",
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
```
xenova:
```suggestion
    "module": "esnext",
    "moduleResolution": "bundler",
```
(see vits tokenizer comments)

.gitignore
```diff
@@ -10,6 +10,10 @@ deno.lock
 /dist
 /types
```
xenova: these are probably unnecessary now. same for other root files

packages/transformers/src/models/vits/tokenization_vits.js
xenova: [updated this myself, but you can check if fixes the issue]

package.json
nico-martin: Yes, It should not be connected to the version of `@huggingface/transformers`

.github/workflows/tests.yml
nico-martin: You mean publish workflows per package? Definitely makes sense.

packages/transformers/scripts/esbuild/build/plugins/postBuildPlugin.mjs
nico-martin: You're right. In our current setup, only the first path is used.

.gitignore
nico-martin: You're right.

/dist can be removed. It's never created at the repository root.
/types should be kept. If someone runs tsc or npx tsc --build from the repository root (intentionally or accidentally), it would generate types/ at the root based on the root tsconfig.json.

.github/workflows/tests.yml
xenova: Yep exactly. This can be a follow-up PR.

xenova (commented):
PR's looking pretty good imo! 🙌 I made a couple of fixes for documentation generation, and I've now tested it with hf-doc-builder, and everything seems to be correct. In case you want to run it locally, you can:
1. install https://github.com/huggingface/doc-builder from source (`pip install --upgrade git+https://github.com/huggingface/doc-builder`)
2. in `packages/transformers`, run `pnpm docs-api`, then `doc-builder build transformers.js ./docs/source/ --not_python_module --build_dir ./docs/build/`, then `doc-builder preview transformers.js ./docs/source/ --not_python_module`.

<details>
<summary>You should then see something like this</summary>

Image

</details>

Some things you may be able to check out:
1. Our root directory is filled with lots of config files. Can we find a way to reduce this count? e.g., are both tsconfig files necessary (do we need one in the root, if we have on a per-project level)?
2. We should add a solid CONTRIBUTING.md, probably following from your work in https://github.com/huggingface/transformers.js/pull/1433, including how the project is structured, the build process, opening PRs, etc.
3. We dynamically build the Transformers.js README.md, (`pnpm readme`), but we don't want to commit duplicates of this file. On the other hand, when publishing, we would need the README.md inside of the packages/transformers folder. On that note, we would probably eventually need to make the monorepo README different from the transformers.js README (where the monorepo README explains the project structure).
4. Some pages (e.g., api/tokenizers) contain many unnecessary sections (because these are exported by the module). Perhaps we can find a way with https://www.npmjs.com/package/jsdoc-to-markdown to filter definitions to only be visible when this is a global export of transformers.js
5. Documentation improvements, as a whole, would be most welcomed (in a separate PR), and the doc builder would then be able to include code snippets inside the generated docs.

nico-martin:
@xenova, I added a CONTRIBUTING.md and did some changes to the config files. Unfortunately I dont think we can remove more files.
For the readme/documentation stuff I think we should open another PR.

PR #1505, opened 2026-01-22
[v4] Finalize pipeline types

xenova:
Ensures correct dynamic pipeline types for all those left.

I also created a unit test for this, to check that the test file compiles with no issues.


src/models/supertonic/modeling_supertonic.js
xenova, line 1: This adds batching support for supertonic as well as for the text-to-audio pipeline. I included it in this PR because the implementation and types are now consistent.

Issue #1479, opened 2025-12-09
[RFC] Move from global `env` to session-based configuration


nico-martin:
## Summary

Currently, Transformers.js uses a global `env` object for configuration. I propose moving to a session-based approach where configuration is passed directly to pipeline instances, while maintaining backward compatibility with the existing `env` object as defaults.

## Current Behavior
```javascript
import { env, pipeline } from "@huggingface/transformers";

env.remoteHost = "https://myhost.com";

const pipe = await pipeline('sentiment-analysis', 'Xenova/bert-base-multilingual-uncased-sentiment');
// All pipelines now use myhost.com
```

All pipelines share the same global configuration through the `env` object.

## Proposed Behavior
```javascript
import { pipeline } from "@huggingface/transformers";

const pipe = await pipeline('sentiment-analysis', 'Xenova/bert-base-multilingual-uncased-sentiment', {
  remoteHost: "https://myhost.com",
});
// Only this pipeline uses myhost.com
```

Configuration is passed per-pipeline, allowing different instances to use different settings.

## Motivation

1. **Client-side context**: The `env` concept feels server-oriented, where configuration is typically set once at startup. On the client, where you have long-running processes and lots of user interaction, a session-based approach makes more sense.

2. **Better encapsulation**: Different pipeline instances might need different configurations (e.g., loading some models from a local cache, others from a CDN, others from a private host).

3. **Industry standard**: Other similar libraries follow this pattern:
   - MediaPipe.js
   - TensorFlow.js
   - WebLLM
   
   They all use session/instance-based configuration rather than global state.

4. **Semantic clarity**: An "environment" (`env`) should represent static values based on the environment an app runs in (e.g., `IS_BROWSER_ENV`, `IS_FS_AVAILABLE`). Most of the current `env` properties are actually runtime configuration options, not environment detection.

## Backward Compatibility

The existing `env` object would serve as the default configuration:
```javascript
import { env, pipeline } from "@huggingface/transformers";

env.remoteHost = "https://myhost.com";

// Uses env.remoteHost as default
const pipe1 = await pipeline("text-generation", "gpt2");

// Overrides the default for this instance
const pipe2 = await pipeline("text-generation", "gpt2", {
  remoteHost: "https://mysecondhost.com",
});
```

This would maintain full backward compatibility while enabling the new pattern.

## Implementation Considerations

[...]

PR #1507, opened 2026-01-29
feat: add configurable log levels via env.logLevel

### LINKED ISSUE #117: [Feature request] Logging Level and Progress Bar for Model Downloads
Right now, the output can be quite lengthy and verbose, see:

image

Could it be possible to expose logger options to granularly control the output as well as offering a visual feedback for the status of the model download from HuggingFace?

taronsung:
## Summary

Adds configurable logging levels to control library verbosity, as requested in #117.

## Changes

- Add `LogLevel` enum with NONE, ERROR, WARNING, INFO, DEBUG levels
- Add `env.logLevel` configuration (defaults to WARNING)
- Create `logger` utility that respects the configured log level
- Update `console.log/warn` calls in `pipelines.js` to use `logger`
- Export `LogLevel` and `logger` from main entry point

## Usage

```javascript
import { env, LogLevel, logger } from '@huggingface/transformers';

// Set log level to suppress all logs
env.logLevel = LogLevel.NONE;

// Set log level to show only errors
env.logLevel = LogLevel.ERROR;

// Set log level to show warnings and above (default)
env.logLevel = LogLevel.WARNING;

// Set log level to show info messages and above
env.logLevel = LogLevel.INFO;

// Set log level to show all messages including debug
env.logLevel = LogLevel.DEBUG;

// Use the logger directly
logger.info('Model loaded successfully');
logger.warn('Deprecated method used');
logger.error('Failed to load model');
logger.debug('Token count:', tokens.length);
```

## Notes

- This PR focuses on the infrastructure and updates `pipelines.js` as a starting point
- Other files with `console.warn/error` calls can be migrated in follow-up PRs
- The `TextStreamer` output in `streamers.js` was intentionally not changed as it's user-facing streaming output, not logging

Closes #117

### DIFF
--- a/packages/transformers/src/env.js
+++ b/packages/transformers/src/env.js
@@ -152,13 +152,6 @@ const localModelPath = RUNNING_LOCALLY ? path.join(dirname__, DEFAULT_LOCAL_MODE
  * // Disable all logging
  * env.logLevel = LogLevel.NONE;
  *
- * @readonly
- * @enum {number}
- * @property {0} NONE - No logging output
- * @property {1} ERROR - Only error messages (value: 1)
- * @property {2} WARNING - Errors and warnings (value: 2)
- * @property {3} INFO - Errors, warnings, and info messages (value: 3)
- * @property {4} DEBUG - All messages including debug output (value: 4)
  */
 export const LogLevel = Object.freeze({
     /** All messages including debug output (value: 10) */


xenova:
Hi @taronsung 👋 Thanks for the PR! I do like this feature, and it's definitely something I think we can implement. Since we're in the process of a relatively large refactor (v4: https://github.com/huggingface/transformers.js/tree/v4), could you update the base to be that branch? From the looks of it, it'd just entail moving the new files into the `packages/transformers` folder... all else should stay mostly similar.

We could also take inspiration from the python transformers library... maybe implementing methods like `warning_once` ([for example](https://github.com/huggingface/transformers/blob/e23f92b30853a4e776b1512b7d61a9513fcdc3c9/src/transformers/utils/logging.py#L319)) to only log once in case of repeated calls. WDYT? Edit: Actually, maybe not worth it because browser runtimes typically collapse duplicate logs anyway.

packages/transformers/src/transformers.js
```diff
@@ -13,7 +13,8 @@
  */
 
 // Environment variables
-export { env } from './env.js';
+export { env, LogLevel } from './env.js';
+export { logger } from './utils/logger.js';
```
nico-martin: Is there a reason you want to export the logger? In my opinion that should stay in the library and should not be user facing.

nico-martin: @taronsung Could you take another look at it and see if you have any further suggestions? Otherwise, we would like to merge the PR. Thank you again for bringing up this topic!

packages/transformers/src/env.js
```diff
@@ -134,12 +134,52 @@ const DEFAULT_CACHE_DIR = RUNNING_LOCALLY ? path.join(dirname__, '/.cache/') : n
...
+    /** Only error messages (value: 1) */
+    ERROR: 1,
+    /** Errors and warnings (value: 2) */
+    WARNING: 2,
+    /** Errors, warnings, and info messages (value: 3) */
+    INFO: 3,
+    /** All messages including debug output (value: 4) */
+    DEBUG: 4,
+});
+
```
xenova: Probably a nit because this won't be visible externally, but we should probably align with how the python logging library does it: https://docs.python.org/3/library/logging.html#logging-levels

(which I think is standard across many other libraries I've seen... and for onnxruntime)

<table class="docutils align-default">
<thead>
<tr class="row-odd"><th class="head"><p>Level</p></th>
<th class="head"><p>Numeric value</p></th>
<th class="head"><p>What it means / When to use it</p></th>
</tr>
</thead>
<tbody>
<tr class="row-even"><td><dl class="py data">
<dt class="sig sig-object py" id="logging.NOTSET">
<span class="sig-prename descclassname"><span class="pre">logging.</span></span><span class="sig-name descname"><span class="pre">NOTSET</span></span><a class="headerlink" href="#logging.NOTSET" title="Link to this definition">¶</a></dt>
<dd></dd></dl>

</td>
<td><p>0</p></td>
<td><p>When set on a logger, indicates that
ancestor loggers are to be consulted
to determine the effective level.
If that still resolves to
<code class="xref py py-const docutils literal notranslate"><span class="pre">NOTSET</span></code>, then all events
are logged. When set on a handler,
all events are handled.</p></td>
</tr>
<tr class="row-odd"><td><dl class="py data">
<dt class="sig sig-object py" id="logging.DEBUG">
<span class="sig-prename descclassname"><span class="pre">logging.</span></span><span class="sig-name descname"><span class="pre">DEBUG</span></span><a class="headerlink" href="#logging.DEBUG" title="Link to this definition">¶</a></dt>
<dd></dd></dl>

</td>
<td><p>10</p></td>
<td><p>Detailed information, typically only
of interest to a developer trying to
diagnose a problem.</p></td>
</tr>
<tr class="row-even"><td><dl class="py data">
<dt class="sig sig-object py" id="logging.INFO">
<span class="sig-prename descclassname"><span class="pre">logging.</span></span><span class="sig-name descname"><span class="pre">INFO</span></span><a class="headerlink" href="#logging.INFO" title="Link to this definition">¶</a></dt>
<dd></dd></dl>

</td>
<td><p>20</p></td>
<td><p>Confirmation that things are working
as expected.</p></td>
</tr>
<tr class="row-odd"><td><dl class="py data">
<dt class="sig sig-object py" id="logging.WARNING">
<span class="sig-prename descclassname"><span class="pre">logging.</span></span><span class="sig-name descname"><span class="pre">WARNING</span></span><a class="headerlink" href="#logging.WARNING" title="Link to this definition">¶</a></dt>
<dd></dd></dl>

</td>
<td><p>30</p></td>
<td><p>An indication that something
unexpected happened, or that a
problem might occur in the near
future (e.g. ‘disk space low’). The
software is still working as
expected.</p></td>
</tr>
<tr class="row-even"><td><dl class="py data">
<dt class="sig sig-object py" id="logging.ERROR">
<span class="sig-prename descclassname"><span class="pre">logging.</span></span><span class="sig-name descname"><span class="pre">ERROR</span></span><a class="headerlink" href="#logging.ERROR" title="Link to this definition">¶</a></dt>
<dd></dd></dl>

</td>
<td><p>40</p></td>
<td><p>Due to a more serious problem, the
software has not been able to
perform some function.</p></td>
</tr>
<tr class="row-odd"><td><dl class="py data">
<dt class="sig sig-object py" id="logging.CRITICAL">
<span class="sig-prename descclassname"><span class="pre">logging.</span></span><span class="sig-name descname"><span class="pre">CRITICAL</span></span><a class="headerlink" href="#logging.CRITICAL" title="Link to this definition">¶</a></dt>
<dd></dd></dl>

</td>
<td><p>50</p></td>
<td><p>A serious error, indicating that the
program itself may be unable to
continue running.</p></td>
</tr>
</tbody>
</table>

packages/transformers/src/transformers.js
```diff
@@ -13,7 +13,7 @@
  */
 
 // Environment variables
-export { env } from './env.js';
+export { env, LogLevel } from './env.js';
```
xenova, line 16: Minimal API surface change - nice!

xenova:
main comment is [here](https://github.com/huggingface/transformers.js/pull/1507#discussion_r2823616461); next, for some reason I'm unable to merge main into this branch. @taronsung could you rebase? alternatively, @nico-martin if you have permission, could you ensure the PR is on latest main? thanks!

packages/transformers/src/env.js
nico-martin: I looked at the 4 most used logging libraries in JS and if they are ascending (higher number = more logs) or descending (higher number = less logs):
https://www.npmjs.com/package/pino (ascending)
https://www.npmjs.com/package/winston (descending)
https://www.npmjs.com/package/log4js (ascending)
https://www.npmjs.com/package/bunyan (ascending)

So most use our approach implemented in LogLevel so I would like to keep it.

nico-martin:
@xenova, I merged origin/main into this branch so its now up to date with main. Imo its ready to be merged :D

packages/transformers/src/env.js
xenova: Thanks for investigating! Sounds good to me. As mentioned before, it's a small implementation detail the user shouldn't have to worry about... but good to know how others do it!

xenova: Hmm, I just checked [pino](https://github.com/pinojs/pino/blob/47c7991b09ff783f1b7f42c1032d1e4e4bf88af5/lib/constants.js#L6-L13) and they do seem to do it descending (higher number = fewer logs):
```
const DEFAULT_LEVELS = {
  trace: 10,
  debug: 20,
  info: 30,
  warn: 40,
  error: 50,
  fatal: 60
}
```

[winston](https://github.com/winstonjs/winston/blob/ed45345f01b8ceb1d436e4791d95469c5213a0cf/README.md?plain=1#L133-L141) is ascending (higher number = more logs)
```
  error: 0,
  warn: 1,
  info: 2,
  http: 3,
  verbose: 4,
  debug: 5,
  silly: 6
```

[log4js](https://github.com/log4js-node/log4js-node/blob/7904872ef4368044791b988a2e15141a7487437a/lib/levels.js#L95-L105) is descending (higher number = fewer logs)
```
  ALL: { value: Number.MIN_VALUE, colour: 'grey' },
  TRACE: { value: 5000, colour: 'blue' },
  DEBUG: { value: 10000, colour: 'cyan' },
  INFO: { value: 20000, colour: 'green' },
  WARN: { value: 30000, colour: 'yellow' },
  ERROR: { value: 40000, colour: 'red' },
  FATAL: { value: 50000, colour: 'magenta' },
  MARK: { value: 9007199254740992, colour: 'grey' }, // 2^53
  OFF: { value: Number.MAX_VALUE, colour: 'grey' },
 ```

[bunyan](https://github.com/trentm/node-bunyan/blob/5c2258ecb1d33ba34bd7fbd6167e33023dc06e40/bin/bunyan#L69-L74) is descending (higher number = fewer logs)
```
var TRACE = 10;
var DEBUG = 20;
var INFO = 30;
var WARN = 40;
var ERROR = 50;
var FATAL = 60;
```

Comment (nico-martin) [reply packages/transformers/src/env.js]: You're right. Somehow, the ascending/descending confused me. I'll change it so that it's closer to Python and the common JS libraries.

File: packages/transformers/src/backends/onnx.js [hunk 4]
```diff
@@ -229,8 +251,8 @@ export async function createInferenceSession(buffer_or_path, session_options, se
     await ensureWasmLoaded();
     const load = () =>
         InferenceSession.create(buffer_or_path, {
-            // Set default log level, but allow overriding through session options
-            logSeverityLevel: DEFAULT_LOG_LEVEL,
+            // Set default log severity level, but allow overriding through session options
+            logSeverityLevel: getOnnxLogSeverityLevel(env.logLevel ?? 30),
```
xenova: We might as well use our own log level dict and then use LogLevel.WARNING here (instead of a magic number)

packages/transformers/src/backends/onnx.js
```diff
@@ -268,7 +290,7 @@ export function isONNXTensor(x) {
 
 /** @type {import('onnxruntime-common').Env} */
 const ONNX_ENV = ONNX?.env;
-ONNX_ENV.logLevel = LOG_LEVELS[DEFAULT_LOG_LEVEL];
+ONNX_ENV.logLevel = ONNX_LOG_LEVEL_NAMES[getOnnxLogSeverityLevel(env.logLevel ?? 30)];
```
xenova: same here

packages/transformers/src/env.js
```diff
@@ -134,12 +134,52 @@ const DEFAULT_CACHE_DIR = RUNNING_LOCALLY ? path.join(dirname__, '/.cache/') : n
...
+ * // Disable all logging
+ * env.logLevel = LogLevel.NONE;
+ *
+ * @readonly
+ * @enum {number}
+ * @property {0} NONE - No logging output
+ * @property {1} ERROR - Only error messages (value: 1)
+ * @property {2} WARNING - Errors and warnings (value: 2)
+ * @property {3} INFO - Errors, warnings, and info messages (value: 3)
+ * @property {4} DEBUG - All messages including debug output (value: 4)
```
xenova:
```suggestion
```

whoops

fs-eire:
Just saw this PR. I need to mention that `ONNX_ENV.logLevel` only takes effect when creating the `OrtEnv`, which is a singleton that lives forever. If set `ONNX_ENV.logLevel` to `"warning"` and create a session and then change it to `"verbose"`, it won't work.

Also, if setting `ONNX_ENV.logLevel` to `"warning"` but setting `logSeverityLevel` in session options to 0 (verbose), it will not output any logs and will cause the inference become slower. The reason is logSeverityLevel setting to 0 will cause the log string to be created and push to the log sink (which is avoid if setting logSeverityLevel to 3), but the actual logger won't output the log according to `ONNX_ENV.logLevel`.

A little bit confusing but need to be careful for this part. @xenova

xenova:
Thanks for the info @fs-eire 👍 Do you have any suggestions to improve this? Luckily it shouldn't affect most users, and the default we have chosen are much better than what they used to be (outputting far too many logs).

This is similar to a problem I described to @nico-martin, which makes session-level settings a bit more difficult to implement (since ORT uses global settings for many things, like proxying). Ref: https://github.com/huggingface/transformers.js/issues/1479

fs-eire:
based on the current limitation, we probably have something a "production" mode and a "diagnose" mode, where in production mode we can set ORT env log level to error, and in diagnose mode we set ORT env log level to verbose (a little bit slower but it's fine since it's diagnose mode). This should match most users' expectation.

nico-martin:
It's not ideal, but I think it's quite rare for ORT variables to be changed by the user.
My main point with https://github.com/huggingface/transformers.js/issues/1479 is not mecessarily about ORT variables, but about session specific variables. Like `remoteHost`, `remotePathTemplate`, etc. Those are things I am concerned because they are not linked to the environment, but rather to a specific session. Especially ehen the ecosystem grows and more dependencies use Transformers.js as a dependency. Then all of a sudden the user has no idea why installing the `feature-extraction`-package influences the `simple-web-llm`-package (one might set a different `remoteHost`).

TLDR: I can live with certain global dependencies that we cannot eliminate for good reasons. :D

PR #1509, opened 2026-01-29
fix: force light theme styles for output tokenids visibility

KongaraLikhith:
**Description:** This PR addresses the visibility issue reported in the Agents Course (Issue #507).

**Problem:** When the course website is in dark mode, the tokenizer (token_ids) output becomes unreadable due to color inheritance inside the iframe.

**Solution:** Instead of adding dark mode support, I have forced the output container to remain in "light mode" (white background, black text). This ensures the playground remains consistent and readable regardless of the theme used in the environment where it is embedded.

**Changes:**
Updated _className_ of the output div in **App.jsx** to explicitly use _bg-white_ and _text-black._

### DIFF
--- a/examples/tokenizer-playground/src/App.jsx
+++ b/examples/tokenizer-playground/src/App.jsx
@@ -146,7 +146,7 @@ function App() {
         </div>
       </div>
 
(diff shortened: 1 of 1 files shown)


DecodersLord:
Duplicate PR #1497

KongaraLikhith:
Thanks for letting me know! I see that PR #[1497] covers this fix. I'll close this one to avoid redundancy

PR #1510, opened 2026-01-30
Fix model fetching from workers.dev domains

sebastian-zm:
Add referrerPolicy: 'no-referrer' to browser fetch calls to prevent Hugging Face infrastructure from blocking requests based on referrer.

To test: `curl https://huggingface.co/Xenova/bert-base-uncased/resolve/main/config.json`
And also: `curl -s https://huggingface.co/Xenova/bert-base-uncased/resolve/main/config.json -H "Referer: https://workers.dev/"`
And also: `curl -s https://huggingface.co/Xenova/bert-base-uncased/resolve/main/config.json -H "Referer: https://example.com/"`

Weird stuff happens when adding a referer from a workers.dev subdomain.

See live example at `https://transformers-js-demo.cloudflare-huddle586.workers.dev/` where the model errors. However, see example at `https://transformers-js-demo-2.cloudflare-huddle586.workers.dev/` using my fork of transformers.js where it actually works with my changes.

### DIFF
--- a/src/utils/hub.js
+++ b/src/utils/hub.js
@@ -245,7 +245,7 @@ export async function getFile(urlOrPath) {
         // Running in a browser-environment, so we use default headers
         // NOTE: We do not allow passing authorization headers in the browser,
         // since this would require exposing the token to the client.
-        return fetch(urlOrPath);
+        return fetch(urlOrPath, { referrerPolicy: 'no-referrer' });
     }
 }
 


sebastian-zm:
I created a discussion to see if the block can be fixed at https://discuss.huggingface.co/t/cannot-access-from-workers-dev-subdomains/172959

nico-martin:
Hi @sebastian-zm
Thanks for the report.

The issue here is that Hugging Face blocks requests from certain origins (like `workers.dev` subdomains) that have been used to mirror the HF website 1:1. This is a platform-level policy to prevent abuse.

Unfortunately, we can't merge the PR that sets `referrerPolicy: 'no-referrer'` globally in Transformers.js, as that would be a workaround that affects all users and could have unintended side effects.

**Recommended workaround for your use case:**

You can monkeypatch `fetch` in your own project to strip the referrer before Transformers.js makes requests:

```js
const originalFetch = window.fetch;
window.fetch = (input, init = {}) => {
  return originalFetch(input, { ...init, referrerPolicy: 'no-referrer' });
};
```

Alternatively, you can self-host the models (e.g. serve them from your own Cloudflare Worker or R2 bucket) to avoid the restriction entirely. This is also the more robust long-term solution.

Sorry for the inconvenience!

Issue #1345, opened 2025-06-25
Add a `supported` API


tarekziade:
### Feature request

It would be nice to add in the APIs something to know if a model is supported, given a hub, model version, id and revision and quantization level. 

I guess this would consist of listing all the files that will get downloaded by Transformers.js before we actually start it.

I am thinking something along those lines:

- grab the config.json
- determine the model architecture
- compute the files required and their URL
- check if present in cache, if not do HEAD calls

### Motivation

This is to avoid starting a session if we know it'll fail (missing quantization, or missing version etc). 

### Your contribution

I am happy to do the patch if I am mentored

xenova:
This will definitely be a useful addition to the library! I think the best way to achieve something like this would be to abstract the logic [here](https://github.com/huggingface/transformers.js/blob/7b450425ee35a55ed2e98b9b582b02141fd2b0c0/src/models.js#L1147-L1285) to define a mapping from `MODEL_TYPES` to required files, and then use that to determine whether a model is supported or not.

Although, we'd need to pair it with an API call to see if these files actually do exist on the Hub (e.g., something like [this](https://huggingface.co/api/models/onnx-community/whisper-tiny-ONNX))

PR #1511, opened 2026-02-03
[v4] Improve download progress tracking (model cache registry and define which files will be loaded for pipelines)

nico-martin:
# Improved Download Progress Tracking

## Problem

Transformers.js couldn't reliably track total download progress because:
- File lists weren't known before downloads started
- File sizes were inconsistent (compressed vs uncompressed)
- No cache awareness before initiating downloads

### New Exported Functions
- **`get_files()`**: Determines required files before downloading
- **`get_model_files()` / `get_tokenizer_files()` / `get_processor_files()`**: Helper functions to identify files for each component
- **`get_file_metadata()`**: Fetches file metadata using Range requests without downloading full content
  - Returns `fromCache` boolean to identify cached files
  - Ensures consistent uncompressed file sizes
- **`is_cached()`**: Checks if all files from a model are already in cache

### Enhanced Progress Tracking
- **`readResponse()` with `expectedSize`**: Falls back to metadata when `content-length` header is missing
- **`total_progress` callback**: Provides aggregate progress across all files

## Review
One thing I am not super confident is the [get_model_files function](https://github.com/huggingface/transformers.js/blob/v4-cache-handler/packages/transformers/src/utils/hub/get_model_files.js). I tried to test it with different model architectures, but maybe I missed some that load files that are not in that function. @xenova, could you smoke-test some models and write mie the models that fail?

Easiest way to do that is:
```javascript
import {
  get_files,
  pipeline,

const expectedFiles = await get_files(
  "onnx-community/gemma-3-270m-it-ONNX",
  {
    dtype: "fp32",
    device: "webgpu",
  }
);
const loadedFiles = new Set();
const pipe = await pipeline(
  "text-generation",
  "onnx-community/gemma-3-270m-it-ONNX",
  {
    dtype: "fp32",
    device: "webgpu",
    progress_callback: (e) => {
      if (e.file) loadedFiles.add(e.file);
    },
  }
);

console.log(
  "SAME FILES:",
  expectedFiles.sort().join(",") === Array.from(loadedFiles).sort().join(",")
);
```

---

Closes https://github.com/huggingface/transformers.js/issues/1345

### DIFF
--- a/packages/transformers/docs/source/_toctree.yml
+++ b/packages/transformers/docs/source/_toctree.yml
@@ -72,8 +72,8 @@
           title: Core
         - local: api/utils/hub
           title: Hub
-        - local: api/utils/ModelRegistry
-          title: ModelRegistry
+        - local: api/utils/model_registry
+          title: Model Registry
         - local: api/utils/image
           title: Image
         - local: api/utils/audio
--- a/packages/transformers/src/utils/model_registry/ModelRegistry.js
+++ b/packages/transformers/src/utils/model_registry/ModelRegistry.js
@@ -91,7 +91,7 @@
  * // }
  * ```
  *
- * @module utils/ModelRegistry
+ * @module utils/model_registry
  */
 
 import { get_files } from './get_files.js';


packages/transformers/src/transformers.js
```diff
@@ -52,6 +52,14 @@ export { load_video, RawVideo, RawVideoFrame } from './utils/video.js';
...
 export { softmax, log_softmax, dot, cos_sim } from './utils/maths.js';
 
+// Hub utilities for cache and file management
+export { get_file_metadata } from './utils/hub.js';
+export { get_files } from './utils/hub/get_files.js';
+export { get_model_files } from './utils/hub/get_model_files.js';
+export { get_tokenizer_files } from './utils/hub/get_tokenizer_files.js';
+export { get_processor_files } from './utils/hub/get_processor_files.js';
+export { is_cached } from './utils/hub/is_cached.js';
+
```
xenova: can we find a way to export a single Object/static class, which encapsulates all these methods?

e.g.,
```js
import { ClassName } from '@huggingface/transformers'; // no idea what to call it yet haha

const x = await ClassName.get_files(...);
```

I feel like this could be more future proof.

No idea what to call this wrapper class, but something along the lines of `CacheRegistry`, `FileRegistry`, `HubRegistry` could work? Maybe we can find a similar abstraction in the transformers/huggingface_hub library?

packages/transformers/src/utils/dtypes.js
```diff
@@ -45,6 +45,7 @@ export const DATA_TYPES = Object.freeze({
 });
 /** @typedef {keyof typeof DATA_TYPES} DataType */
 
+export const DEFAULT_DEVICE_DTYPE = DATA_TYPES.fp32;
```
xenova, line 48: currently, we do a bit of a funny thing when loading models:
1. if wasm, and no dtype set in config, we use q8 (8-bit) model as it's pretty fast on CPU
2. if node cpu, and no dtype set in config, we use fp32 model
3. if webgpu, and no dtype set in config, we use fp32 model

the main reason is that many models on WASM can encounter an out of memory issue if using fp32 on CPU in the browser. Something I think we can do is to do a scan of models on the hub and specify the default dtype there, especially on a per-device basis. We can check in a separate PR whether we have a good way to support per-device dtypes based on configs.

packages/transformers/src/utils/core.js
```diff
@@ -31,6 +31,28 @@
...
  */
 
+/**
+ * @typedef {Object} FileLoadingProgress
+ * @property {number} loaded The number of bytes loaded for this file.
+ * @property {number} total The total number of bytes for this file.
+ */
+
+/**
+ * @typedef {Object.<string, FileLoadingProgress>} FilesLoadingMap
```
xenova:
```suggestion
 * @typedef {Record<string, FileLoadingProgress>} FilesLoadingMap
```

right?

packages/transformers/src/utils/hub/is_cached.js
```diff
@@ -0,0 +1,50 @@
...
+ * @param {import('../../configs.js').PretrainedConfig} [options.config] Pre-loaded config
+ * @param {import('../dtypes.js').DataType|Record<string, import('../dtypes.js').DataType>} [options.dtype] Override dtype
+ * @param {import('../devices.js').DeviceType|Record<string, import('../devices.js').DeviceType>} [options.device] Override device
+ * @returns {Promise<boolean>} True if all files are cached, false otherwise
+ *
+ * @example
+ * import { is_cached } from '@huggingface/transformers';
+ *
+ * const cached = await is_cached('Xenova/gpt2');
+ * console.log(cached ? 'All files cached!' : 'Some files need downloading');
```
xenova: maybe we can return which files we still need to download? Or maybe provide a separate method which does this?

packages/transformers/src/utils/cache/get_model_files.js
```diff
@@ -0,0 +1,200 @@
...
+                modelType = MODEL_TYPES.Seq2Seq;
+            }
+        } else {
+            if (architectures.some((arch) => arch.includes('CausalLM') || arch.includes('LMHead'))) {
+                modelType = MODEL_TYPES.DecoderOnly;
+            } else {
+                modelType = MODEL_TYPES.EncoderOnly;
+            }
+        }
+    }
```
xenova, line 82: can we check which models are not correctly set? Ideally, we shouldn't need manual checking here.

packages/transformers/src/utils/cache/get_model_files.js
```diff
@@ -0,0 +1,200 @@
...
+ */
+export async function get_model_files(
+    modelId,
+    { config = null, dtype: overrideDtype = null, device: overrideDevice = null } = {},
+) {
+    if (!config) {
+        // Use dynamic import to avoid circular dependency
+        const { AutoConfig } = await import('../../configs.js');
+        config = await AutoConfig.from_pretrained(modelId);
+    }
```
xenova: i remember having issues when bundling (especially for CDN serving) when doing dynamic imports. Can we find a way to avoid this?

packages/transformers/src/utils/model_registry/get_model_files.js
```diff
@@ -0,0 +1,200 @@
...
+    // Add model files based on model type
+    if (modelType === MODEL_TYPES.DecoderOnly) {
+        add_model_file('model', 'model');
+        files.push('generation_config.json');
+    } else if (modelType === MODEL_TYPES.Seq2Seq || modelType === MODEL_TYPES.Vision2Seq) {
+        add_model_file('model', 'encoder_model');
+        add_model_file('decoder_model_merged');
+        // Note: generation_config.json is only loaded for generation models (e.g., T5ForConditionalGeneration)
+        // not for base models (e.g., T5Model). Since we can't determine the specific class here,
+        // we include it as it's loaded for most use cases.
```
xenova, line 118: if generation_config.json isn't found, do we throw an error, or gracefully ignore it? 👀 

Luckily, as you said, users very rarely load the non-generation variants of these models.

packages/transformers/src/utils/cache/get_model_files.js
nico-martin: There was a circular dependency issue in the beginning. But it does not seem to be the case any more. So I moved the import to the top.

packages/transformers/src/transformers.js
nico-martin: Good point. I think the perfect solution would be to have a different pattern for the pipeline:
```
const textGenerator = new Pipeline({
  "text-generation",
  "onnx-community/gemma-3-270m-it-ONNX",
  {
    dtype: "fp32",
    device: "webgpu",
  }
});

const isCached = await textGenerator.isCached();
const files = await textGenerator.getFiles();

await textGenerator.load({ progress_callback: console.log })

const output = await textGenerator.generate([
  {
    role: "system",
    content: "You are a helpful assistant",
  },
  {
    role: "user",
    content: "How are you?",
  },
], {
  max_new_tokens: 1024,
})
```

Then those methods would be on the `pipe`. But I definitely see your point. The whole feature should be encapsuled in one export. I'll think of a cleaner solution.

nico-martin: How about this? https://github.com/huggingface/transformers.js/blob/v4-cache-handler/packages/transformers/src/utils/cache/ModelRegistry.js

packages/transformers/src/utils/hub/is_cached.js
nico-martin: You're right. I improved the return value. And I also added a is_pipeline_cached that allows to check is_cached by pipeline. I thought it makes sense to have a get_pipeline_files that also checks if the pipeline has a processor and a tokenizer before it checks if those files are present.

packages/transformers/src/utils/cache/get_model_files.js
nico-martin: Yes, we can check! I added logging to detect models not in MODEL_TYPE_MAPPING:  
 ```
 console.warn(                                                                   
   `[get_model_files] Architecture(s) not found in MODEL_TYPE_MAPPING: [${archList}] ` + 
   `for model type '${config.model_type}'. Using heuristic detection. ` +
   `Consider adding to packages/transformers/src/models/registry.js`
);                                                                              
```
I also improved the lookup to check model_type before falling back to heuristics. This fixed models like modnet (no architectures array) and briaai/RMBG-1.4 which were triggering the fallback unnecessarily.

File: packages/transformers/src/utils/model_registry/get_model_files.js (hunk 7 above)
Comment (nico-martin) [reply packages/transformers/src/utils/model_registry/get_model_files.js:118]: Good question. I mean I wouldn't throw an error. If there is no `generation_config.json` I don't see a problem. It would expect a file, but the expected fize will be 0. So the total would still be correct. On the `progress` callback it will then load all the other files up to 100%, then its done. I think its fine.

Comment (nico-martin) [reply packages/transformers/src/utils/model_registry/get_model_files.js:118]: In fact, its better to add a file if we are not sure. The only downside could be a 404 on the metadata check. But I dont think thats a problem.

File: packages/transformers/docs/source/_toctree.yml [hunk 8]
```diff
@@ -72,6 +72,10 @@
           title: Core
         - local: api/utils/hub
           title: Hub
+        - local: api/utils/cache
+          title: Cache & File Management
+        - local: api/utils/pipeline-tasks
+          title: Pipeline Tasks
```
xenova: Cache management makes sense, but do we need to add these as external/visible docs?

packages/transformers/src/utils/model_registry/clear_cache.js
```diff
@@ -0,0 +1,164 @@
...
+ * @property {boolean} deleted - Whether the file was successfully deleted
+ * @property {boolean} wasCached - Whether the file was cached before deletion
+ */
+
+/**
+ * @typedef {Object} CacheClearResult
+ * @property {number} filesDeleted - Number of files successfully deleted
+ * @property {number} filesCached - Number of files that were in cache
+ * @property {FileClearStatus[]} files - Array of files with their deletion status
+ */
```
xenova, line 24: Are these exposed externally to the user, or just for development internally?

Also, is this implementation (mainly, the names and structure of output) inspired by other similar cache implementations? (e.g., browser APIs or other libraries?) Just curious :D

packages/transformers/src/utils/cache/clear_cache.js
```diff
@@ -0,0 +1,164 @@
...
+ * @property {number} filesDeleted - Number of files successfully deleted
+ * @property {number} filesCached - Number of files that were in cache
+ * @property {FileClearStatus[]} files - Array of files with their deletion status
+ */
+
+/**
+ * Internal helper to clear cached files.
+ *
+ * Uses parallel processing (Promise.all) to delete multiple files concurrently for optimal performance.
+ * This is safe because each file deletion is an independent operation with no shared state or side effects.
```
xenova:
```suggestion
```

do we need that comment? 😅

packages/transformers/src/utils/cache/clear_cache.js
```diff
@@ -0,0 +1,164 @@
...
+            const cached = await checkCachedResource(cache, localPath, proposedCacheKey);
+            const wasCached = !!cached;
+
+            // Try to delete using both possible cache keys
+            let deleted = false;
+            if (wasCached) {
+                // Try deleting with proposedCacheKey first (most likely to work)
+                const deletedWithProposed = await cache.delete(proposedCacheKey);
+
+                // If that didn't work and keys are different, try localPath
```
xenova: update comment; "If that didn't work"

According to the code logic, we will still delete using `localPath` even if `proposedCacheKey` existed and is now deleted.

packages/transformers/src/utils/cache/clear_cache.js
```diff
@@ -0,0 +1,164 @@
...
+ * @param {string} [options.revision] - Model revision (default: 'main')
+ * @param {import('../../configs.js').PretrainedConfig} [options.config] - Pre-loaded config
+ * @param {import('../dtypes.js').DataType|Record<string, import('../dtypes.js').DataType>} [options.dtype] - Override dtype
+ * @param {import('../devices.js').DeviceType|Record<string, import('../devices.js').DeviceType>} [options.device] - Override device
+ * @param {boolean} [options.include_tokenizer=true] - Whether to clear tokenizer files
+ * @param {boolean} [options.include_processor=true] - Whether to clear processor files
+ * @returns {Promise<CacheClearResult>} Object with deletion statistics and file status
+ *
+ * @example
+ * import { clear_cache } from '@huggingface/transformers';
```
xenova: outdated?

packages/transformers/src/utils/cache/clear_cache.js
```diff
@@ -0,0 +1,164 @@
...
+ * @param {Object} [options] - Optional parameters
+ * @param {string} [options.cache_dir] - Custom cache directory
+ * @param {string} [options.revision] - Model revision (default: 'main')
+ * @param {import('../../configs.js').PretrainedConfig} [options.config] - Pre-loaded config
+ * @param {import('../dtypes.js').DataType|Record<string, import('../dtypes.js').DataType>} [options.dtype] - Override dtype
+ * @param {import('../devices.js').DeviceType|Record<string, import('../devices.js').DeviceType>} [options.device] - Override device
+ * @returns {Promise<CacheClearResult>} Object with deletion statistics and file status
+ *
+ * @example
+ * import { clear_pipeline_cache } from '@huggingface/transformers';
```
xenova: also outdated.

packages/transformers/src/utils/cache/clear_cache.js
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * // Clear cached files for a specific pipeline
+ * const result = await clear_pipeline_cache('text-generation', 'Xenova/gpt2');
+ * console.log(`Deleted ${result.filesDeleted} of ${result.filesCached} cached files`);
+ * result.files.forEach(f => {
+ *     console.log(`${f.file}: ${f.deleted ? '✓ deleted' : f.wasCached ? '✗ failed' : 'not cached'}`);
+ * });
+ *
+ * @example
+ * // Background removal (only clears model files, no tokenizer/processor)
```
xenova: shouldn't processor be removed here too? (`Xenova/modnet` has an image processor component, which does load preprocessor_config.json)

packages/transformers/src/utils/model_registry/get_file_metadata.js
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Makes a Range request to get file metadata without downloading full content.
+ *
+ * We use a Range request (first byte only) instead of HEAD because:
+ * 1. Ensures we get the uncompressed file size, which is critical for accurate progress tracking
+ * 2. When compression is used, HEAD request content-length reflects compressed size
+ * 3. But during actual download, fetch API decompresses transparently and reports uncompressed bytes
+ * 4. This mismatch causes progress bar inconsistencies when tracking multiple files
+ * 5. Range requests typically aren't compressed, and content-range header shows true uncompressed size
```
xenova, line 23: very nice!

packages/transformers/src/utils/cache/get_file_metadata.js
```diff
@@ -0,0 +1,186 @@
...
+        }
+        return fetch(urlOrPath, { method: 'GET', headers });
+    } else {
+        // Running in a browser-environment
+        const headers = new Headers();
+        // Request only the first byte to get metadata without downloading the full file
+        headers.set('Range', 'bytes=0-0');
+        return fetch(urlOrPath, { method: 'GET', headers });
+    }
+}
```
xenova: a lot of logic here seems duplicated; can we consolidate with the hub/model file fetching code?

packages/transformers/src/utils/cache/get_file_metadata.js
```diff
@@ -0,0 +1,186 @@
...
+                    return {
+                        exists: true,
+                        size: size ? parseInt(size, 10) : undefined,
+                        contentType: contentType || undefined,
+                        fromCache: false,
+                    };
+                }
+            } catch (e2) {
+                // Ignore
+            }
```
xenova: if this path is triggered, will we need to do the download another time when we try access the rest of the file?
1. metadata check -> FAIL -> download whole file
2. download whole file

packages/transformers/src/utils/cache/get_model_files.js
```diff
@@ -0,0 +1,218 @@
...
+import { AutoConfig } from '../../configs.js';
+
+/**
+ * Returns the list of files that will be loaded for a model based on its configuration.
+ *
+ * This function reads configuration from the model's config.json on the hub.
+ * If dtype/device are not specified in the config, you can provide them to match
+ * what the pipeline will actually use.
+ *
+ * @param {string} modelId The model id (e.g., "Xenova/llama-2-7b")
```
xenova:
```suggestion
 * @param {string} modelId The model id (e.g., "Xenova/gpt2")
```

https://huggingface.co/Xenova/llama-2-7b doesn't exist :)

Also, we should use `onnx-community` models more often for code samples (as I upload all the new models there)

packages/transformers/src/utils/cache/get_model_files.js
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    // Fall back to heuristic detection if not found in mapping
+    if (!foundInMapping) {
+        // Log warning for models not in MODEL_TYPE_MAPPING so they can be added
+        const archList = architectures.length > 0 ? architectures.join(', ') : '(none)';
+        console.warn(
+            `[get_model_files] Architecture(s) not found in MODEL_TYPE_MAPPING: [${archList}] ` +
+                `for model type '${config.model_type}'. Using heuristic detection. ` +
+                `Consider adding to packages/transformers/src/models/registry.js`,
```
xenova: bit of a weird error message to show end users (no "packages/transformers/src/models/registry.js" for them to add to. Maybe ask them to open an issue (GH link in constants)

packages/transformers/src/utils/cache/get_model_files.js
```diff
@@ -0,0 +1,218 @@
...
+            } else {
+                modelType = MODEL_TYPES.Seq2Seq;
+            }
+        } else {
+            if (architectures.some((arch) => arch.includes('CausalLM') || arch.includes('LMHead'))) {
+                modelType = MODEL_TYPES.DecoderOnly;
+            } else {
+                modelType = MODEL_TYPES.EncoderOnly;
+            }
+        }
```
xenova: We should just remove all of this and fallback to EncoderOnly always. The reason is that the other model formats (like Vision2Seq) require a strict file setup and is exceedingly unlikely to be valid fallbacks if the user doesn't have the config set up correctly. Similarly, we are checking `modelName = config.model_type`, so if that is set to musicgen or whisper, then those conditions should have triggered a successful detection.

packages/transformers/src/utils/model_registry/get_model_files.js
```diff
@@ -0,0 +1,218 @@
...
+        files.push('generation_config.json');
```
xenova, line 119: Following up on this. Prompted by this discussion, I went ahead and added the ability to differentiate: https://github.com/huggingface/transformers.js/commit/a75f97702f8fd176b4c0704e449f8dde39f95a26

So, we can incorporate this here. If DecoderOnlyWithoutHead, do not load generation_config. If DecoderOnly, do load generation_config.

packages/transformers/src/utils/cache/is_cached.js
```diff
@@ -0,0 +1,129 @@
...
+ * @param {Object} [options] Optional parameters
+ * @param {string} [options.cache_dir] Custom cache directory
+ * @param {string} [options.revision] Model revision (default: 'main')
+ * @param {import('../../configs.js').PretrainedConfig} [options.config] Pre-loaded config
+ * @param {import('../dtypes.js').DataType|Record<string, import('../dtypes.js').DataType>} [options.dtype] Override dtype
+ * @param {import('../devices.js').DeviceType|Record<string, import('../devices.js').DeviceType>} [options.device] Override device
+ * @returns {Promise<CacheCheckResult>} Object with allCached boolean and files array with cache status
+ *
+ * @example
+ * import { is_cached } from '@huggingface/transformers';
```
xenova: outdated

packages/transformers/src/utils/cache/is_cached.js
```diff
@@ -0,0 +1,129 @@
...
+ * @param {Object} [options] - Optional parameters
+ * @param {string} [options.cache_dir] - Custom cache directory
+ * @param {string} [options.revision] - Model revision (default: 'main')
+ * @param {import('../../configs.js').PretrainedConfig} [options.config] - Pre-loaded config
+ * @param {import('../dtypes.js').DataType|Record<string, import('../dtypes.js').DataType>} [options.dtype] - Override dtype
+ * @param {import('../devices.js').DeviceType|Record<string, import('../devices.js').DeviceType>} [options.device] - Override device
+ * @returns {Promise<CacheCheckResult>} Object with allCached boolean and files array with cache status
+ *
+ * @example
+ * import { is_pipeline_cached } from '@huggingface/transformers';
```
xenova: outdated

packages/transformers/src/utils/cache/is_cached.js
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * // Check cache status
+ * const status = await is_pipeline_cached('text-generation', 'Xenova/gpt2');
+ * console.log(status.allCached ? 'Ready to use!' : 'Will download files');
+ * status.files.forEach(f => {
+ *     console.log(`${f.file}: ${f.cached ? '✓' : '✗'}`);
+ * });
+ *
+ * // Background removal (only needs model, no tokenizer/processor)
```
xenova: same comment here: this pipeline does have processor

xenova: do we need to share this example again here? Duplicated from before. better to expose+document on model registry class method, right?

packages/transformers/src/utils/model_registry/ModelRegistry.js
```diff
@@ -0,0 +1,219 @@
...
+import { get_processor_files } from './get_processor_files.js';
+import { is_cached, is_pipeline_cached } from './is_cached.js';
+import { get_file_metadata } from './get_file_metadata.js';
+import { clear_cache, clear_pipeline_cache } from './clear_cache.js';
+
+/**
+ * Static class for cache and file management operations.
+ * @hideconstructor
+ */
+export class ModelRegistry {
```
xenova, line 110: Move example code for each method from individual files to here.

packages/transformers/src/utils/model_registry/ModelRegistry.js
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * Get all files (model, tokenizer, processor) needed for a model.
+     *
+     * @param {string} modelId - The model id (e.g., "Xenova/bert-base-uncased")
+     * @param {Object} [options] - Optional parameters
+     * @param {import('../../configs.js').PretrainedConfig} [options.config=null] - Pre-loaded config
+     * @param {import('../dtypes.js').DataType|Record<string, import('../dtypes.js').DataType>} [options.dtype=null] - Override dtype
+     * @param {import('../devices.js').DeviceType|Record<string, import('../devices.js').DeviceType>} [options.device=null] - Override device
+     * @param {boolean} [options.include_tokenizer=true] - Whether to check for tokenizer files
+     * @param {boolean} [options.include_processor=true] - Whether to check for processor files
```
xenova, line 121: if user doesn't want tokenizer or processor files, they can just use `get_model_files` or `get_tokenizer_files` or `get_processor_files`, right? Do we need to expose this functionality (i.e., specifying custom options). Seems unlikely to be used. Also introduces non-standard options/parameter names.

packages/transformers/src/utils/cache/ModelRegistry.js
```diff
@@ -0,0 +1,219 @@
...
+     * @param {import('../devices.js').DeviceType|Record<string, import('../devices.js').DeviceType>} [options.device] - Override device
+     * @returns {Promise<import('./clear_cache.js').CacheClearResult>} Object with deletion statistics and file status
+     *
+     * @example
+     * // Clear cached files for a specific pipeline
+     * const result = await ModelRegistry.clear_pipeline_cache('text-generation', 'Xenova/gpt2');
+     * console.log(`Deleted ${result.filesDeleted} of ${result.filesCached} cached files`);
+     *
+     * @example
+     * // Background removal (only clears model files)
```
xenova: same comment here.

packages/transformers/src/utils/hub/files.js
```diff
@@ -202,10 +202,27 @@ export class FileCache {
         }
     }
 
+    /**
+     * Deletes the cache entry for the given request.
+     * @param {string} request
+     * @returns {Promise<boolean>} A Promise that resolves to `true` if the cache entry was deleted, `false` otherwise.
+     */
+    async delete(request) {
+        let filePath = path.join(this.path, request);
```
xenova, line 211:
```suggestion
        const filePath = path.join(this.path, request);
```

packages/transformers/src/utils/core.js
```diff
@@ -31,6 +31,28 @@
...
+ */
+
+/**
+ * @typedef {Record<string, FileLoadingProgress>} FilesLoadingMap
+ * A mapping of file names to their loading progress. Each key is a file path and each value contains
+ * the loaded and total bytes for that file.
+ */
+
+/**
+ * @typedef {Object} ProgressTotalStatusInfo
```
xenova: I think renaming to `TotalProgressInfo` sounds better.

packages/transformers/src/utils/hub.js
```diff
@@ -205,6 +206,9 @@ export async function storeCachedResource(path_or_repo_id, filename, cache, cach
     }
 }
 
+// Re-export from cache folder for backward compatibility
+export { get_file_metadata } from './cache/get_file_metadata.js';
```
xenova: backwards compatibility of? 👀

packages/transformers/src/pipelines/index.js
xenova, line 1: we can maybe move this to the pipelines folder. Could go under an index.js or registry.js file.

packages/transformers/docs/source/_toctree.yml
nico-martin: No, you're right.

packages/transformers/src/utils/model_registry/clear_cache.js
nico-martin: they are not exported from `@huggingface/transformers` directly. But if a user uses `ModelRegistry.clear_cache` they should see correct typehints. Also I like to write the type system (whats the expected input/output) first and then the implementation (typescript habit). So even for only internal methods I mostly have typedefs.
No, the structure is basically what I thought would be good infos to have..

packages/transformers/src/utils/cache/clear_cache.js
nico-martin: // this is a comment :D

packages/transformers/src/utils/cache/clear_cache.js
nico-martin: yes, your right. I changed that!

packages/transformers/src/utils/cache/clear_cache.js
nico-martin: Yep, I removed the examples in packages/transformers/src/utils/cache/ files and only kept those in packages/transformers/src/utils/cache/ModelRegistry.js. Because those are the ones that are visible to the users.

packages/transformers/src/utils/model_registry/ModelRegistry.js
nico-martin: That's a good point. The problem is that it's not necessarily intended for users, but rather that we need it here: https://github.com/huggingface/transformers.js/blob/v4-cache-handler/packages/transformers/src/utils/cache/get_pipeline_files.js#L49-L50

get_pipeline_files is basically get_files but it first checks if we actually need atokenizer or a processor in this pipeline task.

Both include_tokenizer and include_processor are true by default. Would it be better to name them exclude_tokenizer and exclude_processor in a different object?

```
export async function get_files(
    modelId,
    { config = null, dtype = null, device = null } = {},
    { exclude_tokenizer = false, exclude_processor = false } = {},
) {
    const files = [];

    if (!exclude_tokenizer) {
        const tokenizerFiles = await get_tokenizer_files(modelId);
        files.push(...tokenizerFiles);
    }

    files.push(...(await get_model_files(modelId, { config, dtype, device })));

    if (!exclude_processor) {
        const processorFiles = await get_processor_files(modelId);
        files.push(...processorFiles);
    }

    return files;
}
```

That way its clear that `get_files` will always return all files, but if you want to for some reason (for example because the pipeline does not have a tokenizer) you can exclude it?

packages/transformers/src/utils/core.js
nico-martin: True. I thought it makes sense to keep the naming consistent with `ProgressStatusInfo`, so `ProgressTotalStatusInfo `. But `TotalProgressInfo` sounds better.

packages/transformers/src/utils/hub.js
nico-martin: backward compatibility from a previous state of this branch :)

packages/transformers/src/utils/model_registry/get_model_files.js
nico-martin: I added a case for DecoderOnlyWithoutHead. Do you think thats correct?

packages/transformers/src/utils/cache/get_file_metadata.js
nico-martin: I removed this block. The above try block already handles all the edcases I could think of. So if it fails (enters this catch block), its most likely because of a network error, timeout, etc which means the getFile would most likely fail as well. So it does not really make sense to keep it.

packages/transformers/src/utils/model_registry/clear_cache.js
tomaarsen: Heads up, if you post `@huggingface/transformers`, you'll ping everyone in that team on GitHub:
image

Probably best to avoid!

nico-martin: Sorry, I should have used backticks!

tomaarsen: All good!

packages/transformers/src/models/session.js
```diff
@@ -57,35 +45,18 @@ async function getSession(pretrained_model_name_or_path, fileName, options, is_d
...
-            // Defined by the config, and is not "auto"
-            dtype = config_dtype;
-        } else {
-            // Choose default dtype based on device, falling back to fp32
-            dtype = DEFAULT_DEVICE_DTYPE_MAPPING[selectedDevice] ?? DATA_TYPES.fp32;
-        }
+    const rawDtype = options.dtype ?? custom_config.dtype;
+    if (rawDtype && typeof rawDtype !== 'string' && !rawDtype.hasOwnProperty(fileName)) {
+        // dtype is a per-file object but this fileName is not in it
+        const fallback = resolveDtype(rawDtype, fileName, selectedDevice);
```
nico-martin: One questio here. We use `resolveDtype` to get the fallback. But afterwards we only use it for the logger.warn and then "throw it away" (selectedDtype from 57 is used afterwards). Do we actually need this or could we just:

```javascript
const selectedDtype = /** @type {import("../utils/dtypes.js").DataType} */ (
  resolveDtype(rawDtype, fileName, selectedDevice, custom_config.dtype)
);

if (rawDtype && typeof rawDtype !== 'string' && !rawDtype.hasOwnProperty(fileName)) {
  logger.warn(
    `dtype not specified for "${fileName}". Using the default dtype (${selectedDtype}) for this device (${selectedDevice}).`,
  );
}
```

xenova: Spot on! We can simplify this indeed :)

xenova (approved):
Huge PR! 🔥 Thanks so much @nico-martin.

PR #1516, opened 2026-02-06
V4 worker pipeline package

nico-martin:
# Add Web Worker Pipeline Support

[This PR](https://github.com/huggingface/transformers.js-examples/pull/78) reminded me that other libraries offer a fairly simple way to interact with web workers.
So this PR adds a `webWorkerPipeline` that can be used to replace the `pipeline` and a `webWorkerPipelineHandler` that handles the web worker requests.

## Subpackage

I deliberately implemented this feature as the first implementation of a subpackage. It uses the same package commands as the main library, so something like `pnpm dev` or `pnpm build` works for both packages.

We will incorporate deployment to npm at a later date.

## Usage

**Main Thread:**
```javascript
import { webWorkerPipeline } from '@huggingface/transformers-webworker';

const worker = new Worker('worker.js', { type: 'module' });

const pipe = await webWorkerPipeline(
  worker,
  'background-removal',
  'Xenova/modnet',
  {
    device: 'webgpu',
    progress_callback: (e) => {
      console.log('Loading:', e.file);
    },
  }
);

const result = await pipe(image);
```

**Worker Thread (worker.js):**
```javascript
import { webWorkerPipelineHandler } from '@huggingface/transformers-webworker';

const handler = webWorkerPipelineHandler();
self.onmessage = handler.onmessage;
```

### Function Callbacks

Function callbacks like `progress_callback` are automatically handled via a callback bridge and will execute in the main thread:

```typescript
const pipe = await webWorkerPipeline(worker, 'text-generation', 'model', {
  progress_callback: (progress) => {
    console.log('Loading:', progress);
  }
});
```

**Note:** `session_options` cannot contain GPU devices, WebNN contexts, or typed arrays as these are not serializable across worker boundaries.


sroussey:
Having done something similar, i have some notes.

1. just export pipeline with the same signature as the original (it can fail if the worker is not globally registered). This lets people use the same code and just change the imports and call a register worker fn.
2. The new Worker stuff is more finicky than you may think across vite/bun/rolldown/eslint/etc al
3. Not everything is text. Transferables are a thing, and certainly a desirable thing when sending from the worker where it likely no longer needs it, but iffy on the client side.  Not just typed arrays, but buffers backing a canvas, etc. 
4. This didn't make it to v4 release, but maybe it can go into the samples repo and have samples use it. Then it also doesn't need to support everything, just what the samples need.

PR #1520, opened 2026-02-11
Improve text streamer when `skip_special_tokens: false`

xenova:
Previously, streaming with `skip_special_tokens: false`, would result in long stretches of text between special tokens, as we never "reach an end of sentence".

Before:
```
<|channel|>analysis<|message|>We 
need 
to 
write 
a 
poem 
about 
Machine 
Learning. 
Should 
be 
creative, 
maybe 
with 
some 
technical 
references 
but 
poetic. 
Let's 
produce 
a 
nice 
poem.<|end|><|start|>assistant<|channel|>final<|message|>**When 
the 
Machine 
Learns 
to 
Dream**


In 
a 
quiet 
room 
of 
humming 
silicon,  
```

After:
```
<|channel|>
analysis
<|message|>
We 
need 
to 
write 
a 
poem 
about 
Machine 
Learning. 
Should 
be 
creative, 
maybe 
with 
some 
technical 
references 
but 
poetic. 
Let's 
produce 
a 
nice 
poem.
<|end|>
<|start|>
assistant
<|channel|>
final
<|message|>
**When 
the 
Machine 
Learns 
to 
Dream**


In 
a 
quiet 
room 
of 
humming 
silicon,  
```

---

The original implementation was based on the python TextStreamer, but imo this provides a significantly better UX, especially when dealing with parsing reasoning vs. output tokens.

(also, it's pretty old and the original logic is pretty simple)
image

cc'ing @ArthurZucker @Cyrilvallez, in case you want to port this (or similar) logic to python transformers.

### DIFF
--- a/packages/transformers/src/generation/streamers.js
+++ b/packages/transformers/src/generation/streamers.js
@@ -70,6 +70,9 @@ export class TextStreamer extends BaseStreamer {
         this.token_cache = [];
         this.print_len = 0;
         this.next_tokens_are_prompt = true;
+
+        // Track special token IDs for special handling during streaming.
+        this.special_ids = new Set(this.tokenizer.all_special_ids.map(BigInt));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ArthurZucker:
Happy to have in `transformers` if you want to open a PR 👀

xenova:
> Happy to have in `transformers` if you want to open a PR 👀

can do! 🤗

PR #1525, opened 2026-02-13
fix: improves typedef of Message

philnash:
Message is defined as an object with a role and content, both of which are strings. But there are examples throughout the codebase, when using a vision model, showing that you can pass an array of objects with a type property of "string" or "image" and a text property that is a string.

This expands the JSDoc typedef of Message to cover both cases.

### DIFF
--- a/packages/transformers/src/tokenization_utils.js
+++ b/packages/transformers/src/tokenization_utils.js
@@ -65,11 +65,9 @@ const SPECIAL_TOKEN_ATTRIBUTES = [
 ];
 
 /**
- * @typedef {{ type: 'text', text: string, source_lang_code?: string, target_lang_code?: string, [key: string]: any }} TextContent
+ * @typedef {{ type: 'text', text: string, [key: string]: any }} TextContent
  * @property {'text'} type The type of content (must be 'text').
  * @property {string} text The text content.
- * @property {string} [source_lang_code] The source language code.
- * @property {string} [target_lang_code] The target language code.
  */
 
 /**


philnash:
Hey @nico-martin, it's good to be here! I saw you were working on v4 so I had to check it out!

Not sure I understand, it looks to me like `{ type: "image" }` is also supported, it's [tested here for example](https://github.com/huggingface/transformers.js/blob/a1c658accd26f95af298fb641d8e94c9f424add6/packages/transformers/tests/models/idefics3/test_modeling_idefics3.js#L9).

I could narrow the type to:

```typescript
{ type: "text", text: string } | { type: "image" }
```

if that works better?

nico-martin:
Oh wait. You're right. I was only looking at the TextGeneration pipeline. There it makes sense to only allow text. But if you create your own "pipeline" an vision language model, then you could also pass an image in the content array.

However, I think we should distinguish between the different types of content so that it is clear, for example, that text generation models can only process text input, whereas VLMs can also process images.

Would you like to pursue this further, or is it okay with you if I dive in?

philnash:
I have had a look through and I'm afraid I'm a bit lost with how the processors work and how that applies to the model I'm trying to use. (The app I'm playing around with uses [granite-docling-258m-onnx](https://huggingface.co/onnx-community/granite-docling-258M-ONNX) and I'm not sure, for example, how it gets its settings.

I did find some interesting things though:

* The Janus processor [defines a `MultimodalConversation`](https://github.com/huggingface/transformers.js/blob/main/packages/transformers/src/models/janus/processing_janus.js#L22-L26) by combining `Message` with an `images` property of type `(RawImage | string | URL)[]`. This allows you to pass the images, but the original `Message` type still restricts the `contents` to `string`
* The `TextGeneration` pipeline could presumably also receive image inputs if the generator was a vision model that still only generated text. [The `_call` method for the `TextGeneration` pipeline doesn't seem to be typed at all](https://github.com/huggingface/transformers.js/blob/main/packages/transformers/src/pipelines/text-generation.js#L103), so the inputs can be `any`
* The `Idefics3Processor` has a bunch of type issues in its test file, [particularly that the `Message` doesn't fit the data that its tested with](https://github.com/huggingface/transformers.js/blob/main/packages/transformers/tests/models/idefics3/test_modeling_idefics3.js#L32)
* A list of `Message[]` is [redefined as a `Chat` in the `TextGeneration` pipeline](https://github.com/huggingface/transformers.js/blob/main/packages/transformers/src/pipelines/text-generation.js#L8), so sometimes when you're trying to find `Message[]` you need to look for a different type instead

I'm happy to help, but maybe don't have as much time, and definitely not as much experience, with the codebase, so please take it from here if you want!

nico-martin:
After some back and forth, we will stick with a very open implementation. We do have type definitions for autocomplete, but it allows for all kinds of properties.

packages/transformers/src/tokenization_utils.js
```diff
@@ -64,10 +64,32 @@ const SPECIAL_TOKEN_ATTRIBUTES = [
     // additional_special_tokens (TODO)
 ];
 
+/**
+ * @typedef {{ type: 'text', text: string, source_lang_code?: string, target_lang_code?: string, [key: string]: any }} TextContent
+ * @property {'text'} type The type of content (must be 'text').
+ * @property {string} text The text content.
+ * @property {string} [source_lang_code] The source language code.
+ * @property {string} [target_lang_code] The target language code.
+ */
```
xenova, line 71:
```suggestion
/**
 * @typedef {{ type: 'text', text: string, [key: string]: any }} TextContent
 * @property {'text'} type The type of content (must be 'text').
 * @property {string} text The text content.
 */
```

since `source_lang_code` and `target_lang_code` are very specific to translategemma, I think we can remove them from here. The generic `[key: string]: any` should be sufficient.

packages/transformers/src/tokenization_utils.js
```diff
@@ -64,10 +64,32 @@ const SPECIAL_TOKEN_ATTRIBUTES = [
...
+ * @property {'image'} type The type of content (must be 'image').
+ * @property {string | import('./utils/image.js').RawImage} [image] Optional URL or instance of the image.
+ *
+ * Note: This works for SmolVLM. Qwen2VL and Idefics3 have different implementations.
+ */
+
+/**
+ * @typedef {TextContent | ImageContent | { type: string & {}, [key: string]: any }} MessageContent
+ * Base type for message content. This is a discriminated union that can be extended with additional content types.
+ * Example: `@typedef {TextContent | ImageContent | AudioContent} MessageContent`
```
xenova, line 84: sounds good 👍  we can extend with AudioContent once we better support AudioLMs. We are working on this right now too!

xenova:
Thanks @philnash and @nico-martin!

Issue #1522, opened 2026-02-11
[v4.0.0-next.2] ProgressCallback type missing


Jerboas86:
### System Info

transformer.js: 4.0.0-next.2

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension
- [ ] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

`ProgressCallback` type is missing, it can be derived from `PretrainedModelOptions` type like that `type ProgressCallback = Required<PretrainedModelOptions>['progress_callback'];`

But since it is a type that will be probably used by end user, i think it should be exported.

### Reproduction

1- pnpm add @hugginface/transformers@4.0.0-next.2
2- importing `ProgressCallback` fails

xenova:
Thanks for the report. You're right, we should be exporting this. Will be fixed in `@hugginface/transformers@4.0.0-next.4`

PR #1528, opened 2026-02-16
added ProgressCallback and ProgressInfo as type exports

nico-martin:
Due to some refactorings, types for ProgressCallback and ProgressInfo are not longer exported. This PR readds those exports

Closes https://github.com/huggingface/transformers.js/issues/1522


xenova:
thanks!

Issue #1534, opened 2026-02-20
Types not exported


sroussey:
### System Info

version next

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension
- [ ] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

In v3, things like FillMaskSingle, FillMaskOutput where exported, but in v4 they are not. They are export in the file, but not reexported in the barrel file

### Reproduction

Create a new project, include @huggingface/transformers@next and try and `import {FillMaskSingle} from "@huggingface/transformers"`

sroussey:
<img width="305" height="502" alt="Image" src="https://github.com/user-attachments/assets/8fded0ad-7c8c-4569-9b5f-fb2419457161" />

xenova:
Thanks for reporting! Closed by your PR in https://github.com/huggingface/transformers.js/pull/1535.

PR #1535, opened 2026-02-20
Add type definitions for various pipeline outputs in transformers.js

sroussey:
https://github.com/huggingface/transformers.js/issues/1534

### DIFF
--- a/packages/transformers/src/pipelines.js
+++ b/packages/transformers/src/pipelines.js
@@ -577,43 +577,27 @@ export {
 
 // Export pipeline output types
 /**
- * @typedef {import('./pipelines/fill-mask.js').FillMaskSingle} FillMaskSingle
  * @typedef {import('./pipelines/fill-mask.js').FillMaskOutput} FillMaskOutput
- * @typedef {import('./pipelines/text-classification.js').TextClassificationSingle} TextClassificationSingle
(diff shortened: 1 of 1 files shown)


nico-martin:
@sroussey thank you so much for the input! I moved the exports to pipelines.js so that the main export does not become too large and the types remain where they should be (under pipelines).
We have `export * from './pipelines.js';` that then exposes everything to the main export.

xenova:
Thanks @sroussey for the PR! 🙏 One thing I think we should do is only export the Output pipeline types. i.e.,
```
@typedef {import('./pipelines/zero-shot-image-classification.js').ZeroShotImageClassificationOutputSingle} ZeroShotImageClassificationOutputSingle // <- DO NOT EXPORT
@typedef {import('./pipelines/zero-shot-image-classification.js').ZeroShotImageClassificationOutput} ZeroShotImageClassificationOutput // <- EXPORT
```

The `Single` objects are mainly meant to be helpers that we end up using to construct the true `Output` types (as these are what are consumed by users. What do you think? Maybe we can remove the `Single` classes altogether?

sroussey:
That would be fine, as I can derive the single from the output one.

Note that they were output before, so you might add that to a upgrade document for v3 to v4. Though so far, i have not found many issues.

sroussey:
@xenova I went ahead and removed the xxxSingle types

PR #1537, opened 2026-02-20
Add support for seedable random number generation

xenova:
Let there be order amidst the chaos.

Math.random() doesn't support seeding, and since this is very useful in downstream applications, we add our own random module implementation.

Based on Python's random module, we are able to reproduce values exactly.

## Example usage:
```js
import { random } from '@huggingface/transformers';

random.seed(42);
random.random();           // 0.6394267984578837  (matches Python)
random.gauss(0, 1);        // normal-distributed value
random.choices(['a','b'], [3, 1]);  // weighted pick

const arr = [1, 2, 3, 4, 5];
random.shuffle(arr);       // in-place Fisher-Yates shuffle
```

## JS vs PY:
JS:
```js 
Welcome to Node.js v22.14.0.
Type ".help" for more information.
> const { random } = await import("@huggingface/transformers");
undefined
> random.random()
0.4025037000791185
> random.seed(42)
undefined
> random.random()
0.6394267984578837
> random.random()
0.025010755222666936
> random.random()
0.27502931836911926
> random.seed()
undefined
> random.random()
0.012969455082208925
```

PY:
```py
Python 3.13.2 (v3.13.2:4f8bb3947cf, Feb  4 2025, 11:51:10) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Cmd click to launch VS Code Native REPL
>>> import random
>>> random.random()
0.7493327713461649
>>> random.seed(42)
>>> random.random()
0.6394267984578837
>>> random.random()
0.025010755222666936
>>> random.random()
0.27502931836911926
>>> random.seed()
>>> random.random()
0.9648080013684271
```

Adapted from my implementation at https://github.com/xenova/microgpt.js/blob/main/random.js,

### DIFF
--- a/packages/transformers/tests/utils/random.test.js
+++ b/packages/transformers/tests/utils/random.test.js
@@ -17,9 +17,6 @@ init();
  * The JS implementation must match Python's output exactly for reproducibility.
  */
 
-// Tolerance for floating-point comparisons
-const FLOAT_TOL = 1e-15;
-
 describe("random (Mersenne Twister 19937, Python-compatible)", () => {
   // Must be the first test — verifies auto-seeding at module load produces non-zero output.
   it("should return non-zero values before any explicit seed() call", () => {
@@ -34,7 +31,7 @@ describe("random (Mersenne Twister 19937, Python-compatible)", () => {
       const expected = [0.6394267984578837, 0.025010755222666936, 0.27502931836911926, 0.22321073814882275, 0.7364712141640124, 0.6766994874229113, 0.8921795677048454, 0.08693883262941615, 0.4219218196852704, 0.029797219438070344];
       random.seed(42);
       for (let i = 0; i < expected.length; i++) {
-        expect(Math.abs(random.random() - expected[i])).toBeLessThan(FLOAT_TOL);
+        expect(random.random()).toBe(expected[i]);
       }
     });
 
@@ -42,7 +39,7 @@ describe("random (Mersenne Twister 19937, Python-compatible)", () => {
       const expected = [0.8444218515250481, 0.7579544029403025, 0.420571580830845, 0.25891675029296335, 0.5112747213686085, 0.4049341374504143, 0.7837985890347726, 0.30331272607892745, 0.4765969541523558, 0.5833820394550312];
       random.seed(0);
       for (let i = 0; i < expected.length; i++) {
-        expect(Math.abs(random.random() - expected[i])).toBeLessThan(FLOAT_TOL);
+        expect(random.random()).toBe(expected[i]);
       }
     });
 
@@ -50,7 +47,7 @@ describe("random (Mersenne Twister 19937, Python-compatible)", () => {
       const expected = [0.41661987254534116, 0.010169169457068361, 0.8252065092537432, 0.2986398551995928, 0.3684116894884757, 0.19366134904507426, 0.5660081687288613, 0.1616878239293682, 0.12426688428353017, 0.4329362680099159];
       random.seed(12345);
       for (let i = 0; i < expected.length; i++) {
-        expect(Math.abs(random.random() - expected[i])).toBeLessThan(FLOAT_TOL);
+        expect(random.random()).toBe(expected[i]);
       }
     });
 
@@ -58,7 +55,7 @@ describe("random (Mersenne Twister 19937, Python-compatible)", () => {
       const expected = [0.6614192930187859, 0.24151842637090726, 0.6909905728254405, 0.11131596896699081, 0.8184932948202619, 0.4957890371902045, 0.3754628980018908, 0.6967482542555166, 0.5464037278028259, 0.4648564722045134];
       random.seed(2 ** 40 + 17);
       for (let i = 0; i < expected.length; i++) {
-        expect(Math.abs(random.random() - expected[i])).toBeLessThan(FLOAT_TOL);
+        expect(random.random()).toBe(expected[i]);
       }
     });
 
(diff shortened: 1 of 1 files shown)


packages/transformers/src/utils/random.js
nico-martin, line 1: Thats a nice feature to have! I'll definitely take a closer look at the implementation next monday.

But I already see a problem on the API surface:
The seed is created in the global context. This means that if I have two asynchronous functions and both create a seed, then function1 influences function2.

```javascript
async function generate(seed) {
  random.seed(seed); 
  await someAsyncStuff();
  return random.random();
}

// this works
// Finishes completely
await generate(42);
 // Then runs
await generate(99);

// Problem: Both calls interfere with each other, so both get corrupted sequences.
await Promise.all([
  generate(42),
  generate(99),
]); 
```

My solution would be to expose a Random class:
```javascript
async function generate(seed) {
  const generator = new Random(seed);
  await someAsyncStuff();
  return generator.random();
}

await Promise.all([
  generate(42),
  generate(99),
]); 
// now, each instance of Random has its own seed state
```

nico-martin: btw. looks like Python is also aware of that problem: https://docs.python.org/3/library/random.html#:~:text=The%20functions%20supplied%20by%20this%20module%20are%20actually%20bound%20methods%20of%20a%20hidden%20instance%20of%20the%20random.Random%20class.%20You%20can%20instantiate%20your%20own%20instances%20of%20Random%20to%20get%20generators%20that%20don%E2%80%99t%20share%20state.

xenova: Sure, we can expose `random.Random()` for users to use if they'd like. Internally, it would be easier to just have a global RNG state (like transformers/torch)... otherwise we need to find ways to pass state throughout the library... a nightmare!

nico-martin: Youre right. "passing state everywhere is a nightmare" is a valid concern. But, I think there's a middle ground we should look at.

The concern is the global mutable RNG state. Interestingly, Python's own docs acknowledge this: [they note that sequences are only stable as long as multiple threads are not running.](https://docs.python.org/3/library/random.html#notes-on-reproducibility)

In JS, async makes this more common than threading in Python, so it's worth getting right from the start.

The fix doesn't require threading state through the whole library. It just requires that call sites instantiate a `Random` object (maybe even with a configurable seed?) rather than mutating the shared one.

If I see that correctly, all existing uses (`rand()`, `randn()`, `gauss()` in tensor.js, `logits_sampler.js`, etc.) don't call `seed()` at all, They just consume from the global state. Those could stay as-is, using a module-level `Random` instance seeded at import time, giving identical behavior to today. The difference is that users and any future internal code that needs isolation can opt into a local instance.

For user-controlled seeding, we could add an `env.randomSeed` variable. Internally, every `new Random()` call would fall back to `env.randomSeed` if no explicit seed is passed. That way users have a single, familiar place to set a global seed when they need reproducibility, without us going back to shared mutable state.

```js
// Internally we could just do this:
const rng = new Random(env.randomSeed);
rng.random();
```

I don't feel strongly enough about this to block the PR. But global mutable RNG state is arguably a design flaw in Python, and definitively one in JavaScript where async is the norm. Since we're greenfield here, it feels like a good opportunity to not carry that forward.

xenova: Very solid points raised. 👍  The one concern I have with re-seeding before each generation call is that then the user would always get the same output... regardless of the time they seeded. For example, a user calls:
```js
import { random } from '@huggingface/transformers';

random.seed(42); // user wants to return the same sequence of 10
for (let i = 0; i < 10; i++) {
   console.log(await generate(...)); // but we would always return the same value, since we would re-seed internally
}
```

equivalent to
```js
import { random } from '@huggingface/transformers';

for (let i = 0; i < 10; i++) {
   random.seed(42);
   console.log(await generate(...)); // user wants to return the same value, explicitly re-seeding
}
```

vs. this which would be intended

```js
import { random } from '@huggingface/transformers';

random.seed(42);
for (let i = 0; i < 10; i++) {
   console.log(await generate(...)); // would always return the same SEQUENCE of items
}
```

xenova: > If I see that correctly, all existing uses (rand(), randn(), gauss() in tensor.js, logits_sampler.js, etc.) don't call seed() at all, They just consume from the global state.

Indeed, and this is intended: we seed once, call random calls *in sequence*, each internally updating the current PRNG state, and continuing.

The point you raise, however, about JS being async-heavy is true though. The only thing we can control is if the user calls the methods in sequence... If not (i.e., any interleaving), there could be differences. For example, let's say the user calls a text-generation pipeline with do_sample: true, while they are using TTS with supertonic-tts. At any point during generation, if they call TTS generation, we would randomly sample gaussian noise for supertonic TTS -- updating the PRNG state -- and leading to different output tokens being generated afterwards.

xenova: > I don't feel strongly enough about this to block the PR. But global mutable RNG state is arguably a design flaw in Python, and definitively one in JavaScript where async is the norm. Since we're greenfield here, it feels like a good opportunity to not carry that forward.

Haha -- I'm sure it's treated as a feature and not a bug nowadays in python 😜 I'd say this isn't a massive blocker for the PR (mainly because before there was no possibility of reproducibility -- not even Math.random is seed-able). And the origin of this feature was to debug models that use some form of RNG/sampling.

I'll add a note about reproducibility though. As long as we inform the user of this, I think it's okay.

nico-martin: That works for me.
Just so I've said it: No, this is not a feature in Python. It's just poor state management. :D

xenova: > Just so I've said it: No, this is not a feature in Python. It's just poor state management. :D

🤣 the record will show!

Issue #1182, opened 2025-02-04
There is no way to stop / abort / cancel a pipeline()


sroussey:
### Feature request

Provide a way to abort a long running operation.

### Motivation

Many things can take a long time (downloading a multi-gig model on 3G) or are not operating normally (infinite output) and the user may change their mind.

### Your contribution

I can look into how to use AbortSignals for the download issue. 

https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal

Any API suggestion?

xenova:
Good idea! I'm also open to API suggestions since this functionality isn't available in the python library. Following the examples in the docs, perhaps passing a `signal` parameter to calls to `from_pretrained` or `pipeline` could work. Similarly for calls to `generate`.

natanfudge:
What is currently blocking this feature?

sroussey:
No idea, but you can test my version by using `@sroussey/transformers` instead.

[Differences in fork (README version)](https://github.com/sroussey/transformers.js/tree/v3-publish?tab=readme-ov-file#this-is-a-fork)

[Differences in fork (code version)](https://github.com/huggingface/transformers.js/compare/main...sroussey:transformers.js:v3-publish)

natanfudge:
@sroussey Thanks, that will do.
Hopefully your version will be adopted soon!

natanfudge:
@sroussey I see there's a way to abort loading a model, but is there a way to abort text generation? I don't see something like that for `GenerationConfig`.

supermoos:
Is there a new status on adding support for aborting a pipeline generation?

jdp8:
Hi @xenova! Will the model download abort be added soon to transformers.js?

sroussey:
@xenova If there is interest, I can rebase my work.

@natanfudge and @jdp8 -- there is a way to abort generation since there is a break between words. There is an example project on hugging face, but the api is not intuitive. I was thinking of adding an abort controller to that as well to have a unified JS way of doing that across operations. But not everything is a text generation in transformers.js, and I haven't looked at the others. 

I had a moment where I was rewriting the whole lib in typescript, but I got sidetracked. If I ever do it, I will known the library well enough to do more.

jdp8:
[...]

PR #1538, opened 2026-02-20
Add customFetch to env for user override

sroussey:
Ref: https://github.com/huggingface/transformers.js/issues/1182 and https://github.com/huggingface/transformers.js/issues/1479


sroussey:
fine with me!

sroussey:
I have one change related to the type.

xenova:
thanks again!

PR #1539, opened 2026-02-20
feat: add beam search, diverse beam search, and beam sampling to generate()

justsml:
## Summary

Adds beam-based decoding to `generate()` for encoder-decoder and decoder-only models. `num_beams > 1` now maintains and ranks multiple continuations instead of silently using only the first sampled token.

- Standard beam search, diverse beam search (`num_beam_groups` / `diversity_penalty`), and beam sampling (`do_sample`).
- Completed-hypothesis scoring with `length_penalty`, `early_stopping`, and `num_return_sequences`.
- Interleaved batch expansion and decoder KV-cache reordering; encoder cache tensors are shared within each input's beams.
- Explicit errors for streaming, classifier-free guidance, and diverse beam sampling.

## Compatibility with current main

Merged `main` at `a1728dd3` (4.3.0). Beam decoding now uses the extracted `getPastKeyValues()` helper and preserves the `DynamicCache` methods required by subsequent decoding steps. Finalization keeps the reordered cache instead of restoring disposed output tensors, and releases caches that are not returned. Beam outputs continue to return `past_key_values: null` with `return_dict_in_generate`.

Also updates decoder attention masks, fixes declaration-generation errors in the sampling/tensor helpers, makes asynchronous row selection preserve caller-owned resources, and registers beam-search API documentation in the new documentation navigation.

## Validation

- Full workspace suite: **2,255 passed, 25 skipped, 28 suites passed**. Run with `NODE_OPTIONS=--max-old-space-size=8192 TESTING_REMOTELY=true pnpm test --maxWorkers=2 --workerIdleMemoryLimit=512MB --coverage=false`; the initial single-process run exhausted Node's default 4 GB heap while loading fixtures.

- `pnpm build` — both workspace packages and their TypeScript declarations pass.
- Generation suite — 33 tests pass using real tiny T5 and Llama models, including standard/diverse/sampling modes and existing greedy/streaming/cache behavior.
- New regressions verify `DynamicCache` methods after reordering and compare batched standard/diverse generation with independent inputs, including multiple returned sequences.
- `pnpm --filter @huggingface/transformers docs-generate` — passes; a second run produces no tracked changes.
- Prettier checks pass for all changed files. The repository-wide check has existing formatting failures in `tests/utils/tensor.test.js` and `tests/utils/tensor_ops.test.js`, verified against `main`.

Validation uses Node.js 24 on macOS with the CPU backend. Browser/WebGPU execution is not validated here.

### DIFF
--- a/packages/transformers/src/generation/beam_search.js
+++ b/packages/transformers/src/generation/beam_search.js
@@ -1,0 +1,231 @@
+/**
+ * @module generation/beam_search
+ */
+
+/**
+ * Stores completed beam search hypotheses for a single batch element.
+ */
+export class BeamHypotheses {
+    /**
+     * @param {number} num_beams Number of beams.
+     * @param {number} length_penalty Exponential penalty to the length.
+     * @param {boolean|"never"} early_stopping Whether to stop when enough hypotheses are finished.
+     */
+    constructor(num_beams, length_penalty = 1.0, early_stopping = false) {
+        this.num_beams = num_beams;
+        this.length_penalty = length_penalty;
+        this.early_stopping = early_stopping;
+
+        /** @type {{score: number, tokens: bigint[]}[]} */
+        this.beams = [];
+        this.worst_score = 1e9;
+    }
+
+    get length() {
+        return this.beams.length;
+    }
+
+    /**
+     * Add a new hypothesis to the list.
+     * @param {number} sum_logprobs Sum of log probabilities of the hypothesis.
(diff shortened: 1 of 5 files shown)


alex-calderwood:
What is the state of this? I am socked transformers.js doesn't support beam search. Is this useable if I use this branch?

justsml:
> What is the state of this? I am socked transformers.js doesn't support beam search. Is this useable if I use this branch?

It worked last i checked, I'll try update with main too

PR #1547, opened 2026-02-28
Memoize network calls in model registry

nico-martin:
The `ModelRegistry` has two points where calls to the model host are made: fetching the model config (`config.json`) and fetching file metadatas. When a consumer first calls `ModelRegistry.get_model_files` to inspect what a model needs (display the total size on the dowload button) and then creates the pipeline, both of those network calls are made twice. Which is not necessary.

This PR introduces a small `memoizePromise` utility that deduplicates promises by key. Whether the first call is still in-flight or already resolved, any subsequent call with the same key gets the same promise back. This is then applied to `get_file_metadata` and the new `get_config` helper in `get_model_files.js`, so the config and metadata fetches each happen at most once per unique set of arguments.

### DIFF
--- a/packages/transformers/docs/source/_toctree.yml
+++ b/packages/transformers/docs/source/_toctree.yml
@@ -72,8 +72,8 @@
           title: Core
         - local: api/utils/hub
           title: Hub
-        - local: api/utils/cache
-          title: Cache & File Management
+        - local: api/utils/model_registry
+          title: Model Registry
         - local: api/utils/image
           title: Image
         - local: api/utils/audio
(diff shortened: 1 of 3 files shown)


packages/transformers/src/utils/memoize_promise.js
xenova, line 1: Nice idea! I think -- in any case, we should add a max cache size, kind of like how we do here: https://github.com/huggingface/tokenizers.js/blob/488865017c65e9b84b84ede4d7fa89f348dc0fea/src/utils/data-structures/LRUCache.ts#L4

nico-martin: Great idea. Implemented it in https://github.com/huggingface/transformers.js/pull/1552

nico-martin:
Closing this infavor of https://github.com/huggingface/transformers.js/pull/1552 so diff is clean

PR #1552, opened 2026-03-02
Memoize network calls in model registry

nico-martin:
The `ModelRegistry` has two points where calls to the model host are made: fetching the model config (`config.json`) and fetching file metadatas. When a consumer first calls `ModelRegistry.get_model_files` to inspect what a model needs (display the total size on the dowload button) and then creates the pipeline, both of those network calls are made twice. Which is not necessary.

This PR introduces a small `memoizePromise` utility that deduplicates promises by key. Whether the first call is still in-flight or already resolved, any subsequent call with the same key gets the same promise back. This is then applied to `get_file_metadata` and the new `get_config` helper in `get_model_files.js`, so the config and metadata fetches each happen at most once per unique set of arguments.

### DIFF
--- a/packages/transformers/docs/source/_toctree.yml
+++ b/packages/transformers/docs/source/_toctree.yml
@@ -82,8 +82,6 @@
           title: Tensor
         - local: api/utils/maths
           title: Maths
-        - local: api/utils/data-structures
-          title: Data Structures
         - local: api/utils/logger
           title: Logger
         - local: api/utils/random


packages/transformers/src/utils/lru_cache.js
xenova, line 1: to avoid duplicating between this and tokenizers.js, maybe we should release a package `@huggingface/lru`? 😅

nico-martin: Yes, right next to `@huggingface/isEven` and `@huggingface/leftPad` 😄

xenova: LOL 🤣 love it.

but genuine question -- do we have a good way of deduplicating 😅 ? It's not that big, luckily, but two implementations of the same thing is unnecessary.

nico-martin: Fun fact, while digging into this, I found and removed an ancient `LRUCache` implementation that was in `data-structures.js` since when tokenizers was still part of transformers.js. :D

But to your question:

**Within the monorepo**, we could create a workspace package (e.g. `packages/shared`), reference it via `"workspace:*"`. pnpm handles the rest. Easy. So just a shared package that all other packages borrow stuff from.

**Across separate repos**, we would need to publish a shared npm package (e.g. `@huggingface/utils`). But tbh I don't think that makes sense for most of these generic utility functions. Just copying them seems fine to me.

nico-martin:
Turns out we dont need the classes in `utils/data-structures.js` at all. I removed them.

packages/transformers/src/utils/lru_cache.js
xenova: yeah good point - luckily this implementation is tiny, so we won't duplicate too much. maybe a future consideration.

PR #1555, opened 2026-03-03
Use ModelRegistry for pipeline file loading

xenova:
Now that we have the ModelRegistry API, we can use. it to figure out which pipelines need which files.

It also fixes issues/errors that were thrown when loading models like https://huggingface.co/onnx-community/Supertonic-TTS-ONNX, as we'd first try to load the preprocessor_config.json (which doesn't exist).

... and cleans up a bunch of old legacy code 😅 (like `loadItems`)

### DIFF
--- a/packages/transformers/src/pipelines.js
+++ b/packages/transformers/src/pipelines.js
@@ -205,7 +205,7 @@ export async function pipeline(
                     `None of the candidate model classes support this type.`,
             );
         }
-        modelPromise = matchedClass.from_pretrained(model, { ...pretrainedOptions, config: resolvedConfig      });
+        modelPromise = matchedClass.from_pretrained(model, { ...pretrainedOptions, config: resolvedConfig });
     } else {
         modelPromise = modelClasses.from_pretrained(model, pretrainedOptions);
     }


packages/transformers/src/pipelines.js
```diff
@@ -184,15 +188,47 @@ export async function pipeline(
...
+    }
+
+    // Wait for all promises to resolve in parallel
+    const keys = Object.keys(pending);
+    const values = await Promise.all(Object.values(pending));
+
+    const results = { task };
+    for (let i = 0; i < keys.length; ++i) {
+        results[keys[i]] = values[i];
+    }
```
nico-martin: I find the code here a little complicated to understand. It took me a moment. I would find a pattern like this clearer:


```suggestion
    // Resolve the correct model class (needs config when multiple candidates exist)
    const modelClasses = pipelineInfo.model;
    let modelPromise;
    if (Array.isArray(modelClasses)) {
        const resolvedConfig = config ?? (await AutoConfig.from_pretrained(model, pretrainedOptions));
        const { model_type } = resolvedConfig;
        const matchedClass = modelClasses.find((cls) => cls.supports(model_type));
        if (!matchedClass) {
            throw Error(
                `Unsupported model type "${model_type}" for task "${task}". ` +
                    `None of the candidate model classes support this type.`,
            );
        }
        modelPromise = matchedClass.from_pretrained(model, { ...pretrainedOptions, config: resolvedConfig      });
    } else {
        modelPromise = modelClasses.from_pretrained(model, pretrainedOptions);
    }

    // Load all components in parallel
    const [tokenizer, processor, model_loaded] = await Promise.all([
        hasTokenizer ? AutoTokenizer.from_pretrained(model, pretrainedOptions) : null,
        hasProcessor ? AutoProcessor.from_pretrained(model, pretrainedOptions) : null,
        modelPromise,
    ]);

    const results = { task, model: model_loaded };
    if (tokenizer) results.tokenizer = tokenizer;
    if (processor) results.processor = processor;
```

This way we dont need the pending-map-trick and have one plain array for the Promise.all.
But maybe thats just personal preference.

xenova: I like yours!

PR #1558, opened 2026-03-05
Fix WASM factory blob URL loading (#1527, #1532)

### LINKED ISSUE #1527: [4.0.0-next.3] CDN wasmPaths breaks local dev servers (CORS Worker construction error)
### System Info

- `@huggingface/transformers@4.0.0-next.3`
- `onnxruntime-web@1.25.0-dev`
- Vite 7.3.1
- Chrome (latest)
- macOS

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension

### Description

In v4, `backends/onnx.js` unconditionally sets `wasmPaths` to `cdn.jsdelivr.net`:

```js
onnxEnv.wasm.wasmPaths = `https://cdn.jsdelivr.net/npm/onnxruntime-web@${ORT_VERSION}/dist/`;
```

This causes `onnxruntime-web` to construct Workers from CDN URLs, which browsers block with a CORS error when the page is served from `localhost` (or any non-CDN origin):

```
Failed to construct 'Worker': Script at 'https://cdn.jsdelivr.net/npm/onnxruntime-web@1.25.0-dev/dist/ort-wasm-simd-threaded.asyncify.mjs'
cannot be accessed from origin 'http://localhost:5175'.
```

Additionally, even when serving the ORT files locally, the COEP requirement (`Cross-Origin-Embedder-Policy: require-corp`) for `SharedArrayBuffer` means ALL sub-resources need proper `Cross-Origin-Resource-Policy` headers — which CDN-served files don't have in a local dev context.

## v3 behavior (worked correctly)

v3.x did **not** set `wasmPaths`, allowing `onnxruntime-web` to resolve WASM files and Worker scripts relative to the importing script. Bundlers like Vite serve these naturally from `localhost`, so everything was same-origin — no CORS issues.

## Suggested Fix

Only default to CDN `wasmPaths` if the user hasn't already configured a custom path and the environment isn't a local dev server:

```js
// Only default to CDN if user hasn't overridden and not on localhost
if (!onnxEnv.wasm.wasmPaths) {
    const isLocalDev = typeof location !== 'undefined' &&
        (location.hostname === 'localhost' || location.hostname === '127.0.0.1');
    if (!isLocalDev) {
        onnxEnv.wasm.wasmPaths = `https://cdn.jsdelivr.net/npm/onnxruntime-web@${ORT_VERSION}/dist/`;
    }
}
```

Or, better yet: revert to v3 behavior and let `onnxruntime-web` resolve paths naturally. Only use CDN as an explicit opt-in via `env.backends.onnx.wasm.wasmPaths`.

## Current Workaround

Override `wasmPaths` after import and serve ORT dist files locally (e.g., via `vite-plugin-static-copy`):

```ts

// Point to locally-served ORT files instead of CDN
env.backends.onnx.wasm.wasmPaths = '/ort/';
```

Vite config (using `vite-plugin-static-copy`):
```ts
import { viteStaticCopy } from 'vite-plugin-static-copy';

// Also need cross-origin isolation headers for SharedArrayBuffer
function crossOriginIsolation() {
    return {
        name: 'cross-origin-isolation',
        configureServer(server) {
            server.middlewares.use((_req, res, next) => {
                res.setHeader('Cross-Origin-Opener-Policy', 'same-origin');
                res.setHeader('Cross-Origin-Embedder-Policy', 'require-corp');
                res.setHeader('Cross-Origin-Resource-Policy', 'same-origin');
                next();
            });
        },
    };
}

export default defineConfig({
    plugins: [
        crossOriginIsolation(),
        viteStaticCopy({
            targets: [{ src: '<path-to-onnxruntime-web>/dist/*', dest: 'ort' }],
        }),
    ],
});
```

### Reproduction

1. Create a Vite project
2. Install `@huggingface/transformers@4.0.0-next.3`
3. In a Web Worker, run:
```ts
import { AutoModelForVision2Seq, env } from '@huggingface/transformers';

env.allowLocalModels = false;
env.useBrowserCache = true;

const model = await AutoModelForVision2Seq.from_pretrained(
    'HuggingFaceTB/SmolVLM-500M-Instruct',
    { dtype: 'q4f16', device: 'webgpu' }
);
```
4. Browser console shows: `Failed to construct 'Worker'` CORS error

### LINKED ISSUE #1532: [v4.0.0-next.3] Incorrect implementation of `loadWasmFactory`
The string replacement in function `loadWasmFactory` is incorrect implemented.

original code: https://cdn.jsdelivr.net/npm/onnxruntime-web@1.25.0-dev.20260212-1a71a5f46e/dist/ort-wasm-simd-threaded.asyncify.mjs
```js
var a=new Worker(new URL(import.meta.url),{type:"module",workerData:"em-pthread",name:"em-pthread"});
```

The actual code being executed in the browser:

```js
var a = new Worker(new URL("https://cdn.jsdelivr.net/npm/onnxruntime-web@1.25.0-dev.20260212-1a71a5f46e/dist"),{
            type: "module",
            workerData: "em-pthread",
            name: "em-pthread"
        });
```

nico-martin:
## Problem

Two related bugs in `loadWasmFactory` (introduced in v4.0.0-next.3):

- Closes #1532: The `import.meta.url` string replacement was broken — `new URL(fileName, import.meta.url)` inside the blob URL factory resolved against `blob:...` instead of the original CDN URL, producing malformed Worker URLs.
- Closes #1527: CDN-served `wasmPaths` caused a CORS error when constructing Workers from a different origin (e.g. `localhost` dev server).

## Root cause

The factory `.mjs` file was cached and re-executed as a blob URL, but `import.meta.url` inside it became `blob:...`. The attempted fix (string-replacing `import.meta.url`) was fragile and broke edgecases.

## Fix

Relies on [onnxruntime PR #27411](https://github.com/microsoft/onnxruntime/pull/27411) (merged Feb 26 2026): when `wasmBinary` is pre-set in the ORT config, ORT sets `locateFile = (f) => f`, bypassing the `new URL(fileName, import.meta.url)` call entirely.

Changes:
- Removesthe `import.meta.url` string replacements from `loadWasmFactory`
- Pre-fetch the `.wasm` binary and set `ONNX_ENV.wasm.wasmBinary` before the factory executes; the `.wasm` and `.mjs` fetches run in parallel, with a revert guard: if `wasmBinary` fails, `wasmPaths.mjs` is reverted so ORT never sees a blob URL without a corresponding binary
- Added `canUseBlobURLs()` exclude service workers and Chrome extensions from using the blob "hack"
- `useWasmCache` now controls both `.wasm` binary caching and `.mjs` factory blob URL caching (they are coupled: a blob URL factory is only safe when `wasmBinary` is set)
- Deno support: patch `globalThis.process?.versions?.node → false` in the factory blob to prevent Deno's web runtime from entering the Node.js branch; throw a clear error if `useWasmCache=false` is set in Deno

### DIFF
--- a/packages/transformers/src/backends/onnx.js
+++ b/packages/transformers/src/backends/onnx.js
@@ -211,11 +211,10 @@ async function ensureWasmLoaded() {
         ONNX_ENV?.wasm?.wasmPaths?.mjs;
 
     if (!shouldUseWasmCache) {
-        // In Deno's web runtime, the WASM factory must be loaded via blob URL so that
-        // Node.js detection can be patched out (see loadWasmFactory). Without caching,
-        // the factory is imported directly from its URL and Deno would crash trying to
-        // use Node.js APIs. useWasmCache defaults to true in this environment, so this
-        // only happens if the user explicitly disables it.
+        // In Deno's web runtime, the WASM factory must be loaded via blob URL so that Node.js detection
+        // can be patched out (see loadWasmFactory). Without caching, the factory is imported directly
+        // from its URL and Deno would crash trying to use Node.js APIs. useWasmCache defaults to true
+        // in this environment, so this only happens if the user explicitly disables it.
         if (apis.IS_DENO_WEB_RUNTIME) {
             throw new Error(
                 "env.useWasmCache=false is not supported in Deno's web runtime. Remove the useWasmCache override.",


xenova (approved):
Very nice PR! 🙏 

I just made some minor updates: moving env detection to env.js and simplifying `loadWasmFactory`. also updated some comments.

PR #1559, opened 2026-03-05
feat: add fast boolean `is_cached` / `is_pipeline_cached`, closes #1554

### LINKED ISSUE #1554: ModelRegistry.is_pipeline_cached does not need network calls
### Question

Assuming the https://github.com/huggingface/transformers.js/pull/1548 is accepted, then if the right directory exists in cache, then we know it is a complete copy, so we don't need to check all the files (or you could if you wanted to verify cache, but that is a different operation really). So if the directory exists, it is cached. 

You could add is_pipeline_cached_and_verified if you want. That could check sizes and everything to try and see that there is no corruption.

But is_pipeline_cached should be near instantaneous, just doing local cache access. 

By the way, i love these additions.

nico-martin:
Splits `is_cached` and `is_pipeline_cached` into two function pairs:

| Function | Returns |
|---|---|
| `is_cached(modelId, options)` | `Promise<boolean>` |
| `is_cached_files(modelId, options)` | `Promise<CacheCheckResult>` |
| `is_pipeline_cached(task, modelId, options)` | `Promise<boolean>` |
| `is_pipeline_cached_files(task, modelId, options)` | `Promise<CacheCheckResult>` |

The boolean variants short-circuit on `config.json`: if that file is not cached, `false` is returned immediately without enumerating all model files.

## Why two functions instead of an option flag

The alternative (`is_cached(modelId, { detail: true })`) was rejected because I dont like the idea of a function changing its return type based on a parameter value.
I know we do this at some places and yes, this can also be reflected in typesctipt types, but I still think having separate functions is a cleaner approach. This way its clear that `_files` returns the details for each file while the other just returns the boolean.
Separate functions keep types simple and make intent explicit at the call site.

### DIFF
--- a/packages/transformers/src/utils/model_registry/is_cached.js
+++ b/packages/transformers/src/utils/model_registry/is_cached.js
@@ -44,7 +44,51 @@ async function check_files_cache(modelId, files, options = {}) {
 }
 
 /**
- * Checks if all files for a given model are already cached.
+ * Internal helper to check whether a single file is cached.
+ * @private
+ * @param {string} modelId - The model id
+ * @param {string} filename - The file path to check
+ * @param {Object} options - Options including cache_dir
+ * @returns {Promise<boolean>}
+ */
+async function is_file_cached(modelId, filename, options = {}) {
+    const cache = await getCache(options?.cache_dir);
+    if (!cache) return false;
+    const { localPath, proposedCacheKey } = buildResourcePaths(modelId, filename, options, cache);
+    return !!(await checkCachedResource(cache, localPath, proposedCacheKey));
+}
+
+/**
+ * Quickly checks if a model is cached by verifying that `config.json` is present,
+ * then confirming all required files are cached.
+ * Returns a plain boolean — use `is_cached_files` if you need per-file detail.
+ *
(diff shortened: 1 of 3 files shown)


packages/transformers/src/utils/model_registry/is_cached.js
```diff
@@ -44,7 +44,51 @@ async function check_files_cache(modelId, files, options = {}) {
...
+ * @param {string} filename - The file path to check
+ * @param {Object} options - Options including cache_dir
+ * @returns {Promise<boolean>}
+ */
+async function is_file_cached(modelId, filename, options = {}) {
+    const cache = await getCache(options?.cache_dir);
+    if (!cache) return false;
+    const { localPath, proposedCacheKey } = buildResourcePaths(modelId, filename, options, cache);
+    return !!(await checkCachedResource(cache, localPath, proposedCacheKey));
+}
```
xenova, line 59: Maybe in future we could make this more efficient by supporting multiple file checks. for same built resource paths.

Would be useful for checking whether all the files needed for a pipeline (for example) are cached.

Luckily, not too much of an issue, and we can consider that in a a future PR (since this is how it works currently).

PR #1569, opened 2026-03-07
Voxtral Realtime, LFM2-VL, Granite Speech, and modeling type refactoring

xenova:
This PR does the following
- Adds support for LFM2-VL, Voxtral Realtime, and Granite Speech
- Allows multimodal models (e.g., qwen3_5, gemma3n gemma3n) to be loaded with the text-generation pipeline
- move model-type -> config data to packages/transformers/src/models/modeling_utils.js, to be a single source of truth. For example, we defined the components twice (once for model registry and once for model type mapping)
- Like the python library, past_key_values/cache should be of type `DynamicCache`... with helper methods to get sequence length
- Add support for `num_logits_to_keep`, significantly reducing memory overhead for generation models
- added more unit tests

### DIFF
--- a/packages/transformers/src/cache_utils.js
+++ b/packages/transformers/src/cache_utils.js
@@ -9,8 +9,16 @@ class _DynamicCache {
      * @param {Record<string, Tensor>} [entries] Initial name→Tensor mappings.
      */
     constructor(entries) {
-        if (entries) {
-            Object.assign(this, entries);
+        if (!entries) return;
+        for (const key in entries) {
+            if (key in this) {
+                throw new TypeError(`Key "${key}" conflicts with an existing property on DynamicCache`);
+            }
+            const value = entries[key];
+            if (!(value instanceof Tensor)) {
+                throw new TypeError(`Expected a Tensor for key "${key}", got ${typeof value}`);
+            }
+            this[key] = value;
         }
     }
 


packages/transformers/src/models/lfm2_vl/image_processing_lfm2_vl.js
```diff
@@ -0,0 +1,337 @@
...
+ * while keeping pixel count within the allowed range.
+ * @param {number} height
+ * @param {number} width
+ * @param {number} downsample_factor
+ * @param {number} min_image_tokens
+ * @param {number} max_image_tokens
+ * @param {number} encoder_patch_size
+ * @returns {[number, number]} [width, height]
+ */
+function smart_resize(height, width, downsample_factor, min_image_tokens, max_image_tokens, encoder_patch_size) {
```
nico-martin: We already have a `smart_resize` in packages/transformers/src/image_processors_utils.js. I think we should reuse this.

nico-martin: We have another `smart_resize` here: https://github.com/huggingface/transformers.js/blob/main/packages/transformers/src/models/qwen2_vl/image_processing_qwen2_vl.js#L19

packages/transformers/src/cache_utils.js
nico-martin, line 1: I like this feature a lot!
The naming could conflict a bit with the networ/caching we already have. I would suggest something like `TensorCache`.
I also think we should move it to packages/transformers/src/utils/TensorCache.js

packages/transformers/src/cache_utils.js
```diff
@@ -0,0 +1,54 @@
...
+     */
+    async dispose() {
+        const promises = [];
+        for (const t of /** @type {Tensor[]} */ (Object.values(this))) {
+            if (t.location === 'gpu-buffer') {
+                promises.push(t.dispose());
+            }
+        }
+        await Promise.all(promises);
+    }
```
nico-martin, line 53: I think this should also delete the reference:

```suggestion
    async dispose() {
        const promises = [];
        /** @type {Record<string, Tensor>} */
        const self = /** @type {any} */ (this);
        
        for (const key in self) {
            const t = self[key];
            if (t.location === 'gpu-buffer') {
                promises.push(t.dispose());
            }
            delete self[key];
        }
        await Promise.all(promises);
    }
```

packages/transformers/src/cache_utils.js
```diff
@@ -0,0 +1,54 @@
...
+ * A cache class that stores past key values as named tensors.
+ */
+class _DynamicCache {
+    /**
+     * Create a DynamicCache, optionally pre-populated with entries.
+     * @param {Record<string, Tensor>} [entries] Initial name→Tensor mappings.
+     */
+    constructor(entries) {
+        if (entries) {
+            Object.assign(this, entries);
```
nico-martin: Maybe add input validation?
```suggestion
            for (const [key, value] of Object.entries(entries)) {
                if (!(value instanceof Tensor)) {
                    throw new TypeError(`Cache entry '${key}' must be a Tensor, got ${typeof value}`);
                }
            } 
            Object.assign(this, entries);
```

packages/transformers/src/cache_utils.js
xenova: Indeed, the naming is slightly confusing, but this aligns with and is exactly the same as the python transformers library: https://github.com/huggingface/transformers/blob/f8786de5e0cde36cc591e963e4332bf70460f752/src/transformers/cache_utils.py#L889

so, we may need to rename some other files (custom to transformers.js)... but in a separate PR

packages/transformers/src/models/lfm2_vl/image_processing_lfm2_vl.js
xenova: let me see if we can reuse a lot of this functionality

packages/transformers/src/cache_utils.js
xenova: hmm, an interesting idea... there's not a direct equivalent in the python library (since they don't need to dispose manually)... Let me think about this.

packages/transformers/src/cache_utils.js
xenova: yeah that'll work! We could even just update items in the loop too... no need for a final `Object.assign` at the end.

packages/transformers/src/cache_utils.js
nico-martin: Perhaps it's not really important to delete the references in Python because you don't have any long-running processes anyway. In JS, it's more common. I also don't think it has a big impact on performance. But imo, it's just cleaner that way.

packages/transformers/src/cache_utils.js
nico-martin: I do not find the name ideal in the Python context either, even though it is certainly less confusing there. This would be a point where I would deviate from the Python library because having a more descriptive name makes sense and has no impact on the public-facing API.

packages/transformers/src/cache_utils.js
xenova: Yeah, dispose basically releases the tensor's memory and "marks" the tensors as disposed. And attempting to use it again will result in a "Tensor has been disposed" error. If we delete the keys, a problem could occur where `past_key_values` is passed (not null or undefined, it is `{}`), but the keys itself are missing, leading to an error which says past key values are missing. Also, currently, the disposal only occurs for webgpu devices. Not sure if other devices need it.

xenova: maybe @fs-eire or @guschmue have insights here (re: disposal of tensors; is deleting the key a good idea?).

packages/transformers/src/models/lfm2_vl/image_processing_lfm2_vl.js
xenova: good spot - I was able to reduce duplication quite a lot (https://github.com/huggingface/transformers.js/pull/1569/commits/ca544ba21f4bd9d016629de1ad9a2bb4ecc1f45e)

PR #1570, opened 2026-03-09
Add Voxtral Realtime 4B speech-to-text model support

shreyaskarnik:
Adds full inference pipeline for VoxtralRealtime (Voxtral-Mini-4B-Realtime-2602):
- Custom model class with per-step audio embedding addition
- Mel spectrogram feature extraction with streaming audio padding
- Processor with streaming prefix token format [BOS, PAD*38]
- Force-EOS when audio positions exhausted
- SuppressTokensLogitsProcessor for generation

Architecture: audio_encoder (ONNX) -> audio_embeds + text_embeds at every decoder step, not just during prefill like standard multimodal models.

Usage:
```js
import {
  VoxtralRealtimeForConditionalGeneration,
  VoxtralRealtimeProcessor,

const model_id = "onnx-community/Voxtral-Mini-4B-Realtime-2602-ONNX";
const processor = await VoxtralRealtimeProcessor.from_pretrained(model_id);
const model = await VoxtralRealtimeForConditionalGeneration.from_pretrained(
  model_id,
  { dtype: { audio_encoder: "q4", embed_tokens: "q4", decoder_model_merged: "q4" }, device: "webgpu" },
);

// audio: Float32Array of 16kHz PCM samples
const inputs = await processor(null, audio);
const outputIds = await model.generate({ ...inputs, max_new_tokens: 512 });

const promptLen = inputs.input_ids.dims[1];
const genIds = Array.from(outputIds.data).slice(promptLen).map(Number);
const text = processor.tokenizer.decode(
  genIds.filter((id) => id !== 32 && id !== 2),
  { skip_special_tokens: true },
);
```

### DIFF
--- a/packages/transformers/src/configs.js
+++ b/packages/transformers/src/configs.js
@@ -70,6 +70,7 @@ function getNormalizedConfig(config) {
         case 'idefics3':
         case 'ultravox':
         case 'voxtral':
+        case 'voxtral_realtime':
         case 'smolvlm':
         case 'gemma3n':
         case 'chatterbox':
@@ -131,6 +132,7 @@ function getNormalizedConfig(config) {
         case 'cohere':
         case 'cohere2':
         case 'mistral':
+        case 'voxtral_realtime_text':
         case 'starcoder2':
         case 'qwen2':
         case 'qwen2_moe':
--- a/packages/transformers/src/generation/logits_process.js
+++ b/packages/transformers/src/generation/logits_process.js
@@ -198,6 +198,34 @@ export class ForcedEOSTokenLogitsProcessor extends LogitsProcessor {
 }
 
 /**
+ * A LogitsProcessor that suppresses a list of tokens at every generation step.
+ */
+export class SuppressTokensLogitsProcessor extends LogitsProcessor {
+    /**
+     * @param {number[]} suppress_tokens The IDs of the tokens to suppress.
+     */
+    constructor(suppress_tokens) {
+        super();
+        this.suppress_tokens = suppress_tokens;
+    }
+
+    /**
+     * @param {bigint[][]} input_ids
+     * @param {Tensor} logits
+     * @returns {Tensor}
+     */
+    _call(input_ids, logits) {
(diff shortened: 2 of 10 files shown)


xenova:
Hi @shreyaskarnik! Thanks for the PR 😇 Coincidentally, we actually are in the process of adding support for this model in this PR: https://github.com/huggingface/transformers.js/pull/1569

Importantly, that model adds support for streamed audio/input features, which is a key feature of the architecture. We'll be releasing sampled code and a demo for it shortly. Perhaps you can look at that PR and compare it with yours, seeing if there are any additions/modifications you would make... or maybe something we missed in the other PR? In a quick scan, perhaps `SuppressTokensLogitsProcessor` could be useful on its own?

shreyaskarnik:
Thanks @xenova! Great timing 😄 I'll review #1569 and compare approaches.

A few things I ran into that might be useful:

- **STREAMING_PAD suppression causes hallucination**: Suppressing token 32 via `suppress_tokens` in logits forces the model to pick "Thank you for watching..." during the silence/delay period. The fix is to let the model output STREAMING_PAD freely and filter at decode time instead.
- **Force-EOS at audio boundary**: When all audio positions are consumed, the model needs to stop — otherwise it hallucinates beyond the audio range. Your `AudioExhaustedCriteria` handles this cleanly 👍

Happy to close this PR in favor of #1569 — your streaming architecture with async iterable chunks and `DynamicCache` is the right approach for the library.

shreyaskarnik:
Sorry for duplication of work @xenova I scanned the open PRs just based on title and did not pick up #1569. Closing this PR and looking forward to the sample code and demo for Voxtral Realtime!

xenova:
No worries! 🤗 really appreciate your enthusiasm!

> STREAMING_PAD suppression causes hallucination: Suppressing token 32 via suppress_tokens in logits forces the model to pick "Thank you for watching..." during the silence/delay period. The fix is to let the model output STREAMING_PAD freely and filter at decode time instead.

Will check this suggestion out 👍 thanks!

PR #1572, opened 2026-03-11
Include `content-length` header in COS `Response`

tomayac:
@nico-martin Quick follow-up to #1549 to navigate around a performance cliff. Without this header, we need to dynamically grow a buffer, which is really slow:

Screenshot 2026-03-11 at 14 00 43

With the header, the buffer is sized correctly from the start. 🚀

### DIFF
--- a/packages/transformers/src/utils/cache/CrossOriginStorageCache.js
+++ b/packages/transformers/src/utils/cache/CrossOriginStorageCache.js
@@ -61,7 +61,7 @@ export class CrossOriginStorage {
             const blob = await handle.getFile();
             return new Response(blob, {
                 headers: {
-                    'content-length': String(blob.size),
+                    'Content-Length': String(blob.size),
                 },
             });
         } catch {


packages/transformers/src/utils/cache/CrossOriginStorageCache.js
```diff
@@ -59,7 +59,11 @@ export class CrossOriginStorage {
         try {
             const [handle] = await navigator.crossOriginStorage.requestFileHandles([makeHashDescriptor(hashValue)]);
             const blob = await handle.getFile();
-            return new Response(blob);
+            return new Response(blob, {
+                headers: {
+                    'content-length': blob.size,
```
xenova:
```suggestion
                    'content-length': String(blob.size),
```

tomayac:
Making TypeScript happier is definitely a good idea; also made a cosmetic change to the header name. PTAL.

xenova:
thanks! lgtm ✅

PR #1589, opened 2026-03-17
docs: add Svelte tutorial

### LINKED ISSUE #171: [Doc request] Add an example guide of how to use it in Svelte (and deploy to HF Spaces)
Similar to the cool React guide, would be awesome to showcase how to use transformers.js from Svelte (and how to deploy the resulting app to Spaces)

No need to do a SvelteKit version IMO, Svelte would be sufficient

Maybe a good first issue for the community?

mvanhorn:
## Summary
- Adds a step-by-step tutorial for building a multilingual translation app with Svelte + Vite + Transformers.js
- Mirrors the structure of the existing React tutorial: project init, Web Worker for ML inference, Svelte components, CSS styling, and HF Spaces deployment
- Adds the Svelte tutorial entry to `_toctree.yml` (placed between React and Next.js)

Closes #171

## Test plan
- [ ] Verify the tutorial renders correctly in the HuggingFace docs preview
- [ ] Follow the tutorial steps to confirm a working Svelte app can be created
- [ ] Confirm the `_toctree.yml` ordering is correct and the page appears in the sidebar

### DIFF
--- a/packages/transformers/docs/source/tutorials/svelte.md
+++ b/packages/transformers/docs/source/tutorials/svelte.md
@@ -118,7 +118,7 @@ First, let's create some child components. Create a folder called `lib` in the `
    </script>
 
    <div class="progress-container">
-     <div class="progress-bar" style="width: {percentage}%">
+     <div class="progress-bar" style:width="{percentage}%">
        {text} ({percentage.toFixed(2)}%)
      </div>
    </div>
--- a/packages/transformers/docs/source/tutorials/svelte.md
+++ b/packages/transformers/docs/source/tutorials/svelte.md
@@ -1,0 +1,483 @@
+# Building a Svelte application
+
+In this tutorial, we'll be building a simple [Svelte](https://svelte.dev/) application that performs multilingual translation using Transformers.js! The final product will look something like this:
+
+![Demo](https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/react-translator-demo.gif)
+
+Useful links:
+
+- [Source code](https://github.com/huggingface/transformers.js/tree/main/examples/svelte-translator)
+
+## Prerequisites
+
+- [Node.js](https://nodejs.org/en/) version 18+
+- [npm](https://www.npmjs.com/) version 9+
+
+## Step 1: Initialise the project
+
+For this tutorial, we will use [Vite](https://vitejs.dev/) to initialise our project. Vite is a build tool that allows us to quickly set up a Svelte application with minimal configuration. Run the following command in your terminal:
+
+```bash
+npm create vite@latest svelte-translator -- --template svelte
+```
+
+If prompted to install `create-vite`, type <kbd>y</kbd> and press <kbd>Enter</kbd>.
+
+Next, enter the project directory and install the necessary development dependencies:
+
+```bash
+cd svelte-translator
+npm install
+```
+
+To test that our application is working, we can run the following command:
+
+```bash
+npm run dev
+```
+
+Visiting the URL shown in the terminal (e.g., [http://localhost:5173/](http://localhost:5173/)) should show the default "Svelte + Vite" landing page.
+You can stop the development server by pressing <kbd>Ctrl</kbd> + <kbd>C</kbd> in the terminal.
+
+## Step 2: Install and configure Transformers.js
+
+Now we get to the fun part: adding machine learning to our application! First, install Transformers.js from [NPM](https://www.npmjs.com/package/@huggingface/transformers) with the following command:
+
+```bash
+npm install @huggingface/transformers
+```
+
+For this application, we will use the [Xenova/nllb-200-distilled-600M](https://huggingface.co/Xenova/nllb-200-distilled-600M) model, which can perform multilingual translation among 200 languages. Before we start, there are 2 things we need to take note of:
+
+1. ML inference can be quite computationally intensive, so it's better to load and run the models in a separate thread from the main (UI) thread.
+2. Since the model is quite large (>1 GB), we don't want to download it until the user clicks the "Translate" button.
+
+We can achieve both of these goals by using a [Web Worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers).
+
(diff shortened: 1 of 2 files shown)


packages/transformers/docs/source/tutorials/svelte.md
```diff
@@ -0,0 +1,483 @@
...
+2. `Progress.svelte`: This component will display the progress for downloading each model file.
+
+   ```svelte
+   <script>
+     export let text = '';
+     export let percentage = 0;
+   </script>
+
+   <div class="progress-container">
+     <div class="progress-bar" style="width: {percentage}%">
```
Jerboas86: I would avoid inline style, it could conflict with strict CSP

mvanhorn: Fixed in dd33f16: switched to Svelte's `style:width` directive, which avoids inline style attributes entirely.

packages/transformers/docs/source/tutorials/svelte.md
```diff
@@ -0,0 +1,483 @@
+# Building a Svelte application
+
+In this tutorial, we'll be building a simple [Svelte](https://svelte.dev/) application that performs multilingual translation using Transformers.js! The final product will look something like this:
+
+![Demo](https://huggingface.co/datasets/Xenova/transformers.js-docs/resolve/main/react-translator-demo.gif)
+
+Useful links:
+
+- [Source code](https://github.com/huggingface/transformers.js/tree/main/examples/svelte-translator)
```
nico-martin: Does not exist. Do you want to add the source here?
https://github.com/huggingface/transformers.js-examples/

packages/transformers/docs/source/tutorials/svelte.md
```diff
@@ -0,0 +1,483 @@
...
+<Tip>
+
+We recommend starting the development server again with `npm run dev`
+(if not already running) so that you can see your changes in real-time.
+
+</Tip>
+
+First, let's create some child components. Create a folder called `lib` in the `src` directory, and create the following files:
+
+1. `LanguageSelector.svelte`: This component will allow the user to select the input and output languages. Check out the full list of languages [here](https://github.com/huggingface/transformers.js-examples/tree/main/react-translator/src/components/LanguageSelector.jsx).
```
nico-martin: I think this should then also point to the svelte example.

packages/transformers/docs/source/tutorials/svelte.md
```diff
@@ -0,0 +1,483 @@
...
+    </div>
+  </div>
+
+  <button disabled={disabled} on:click={translate}>Translate</button>
+
+  <div class="progress-bars-container">
+    {#if ready === false}
+      <label>Loading models... (only run once)</label>
+    {/if}
+    {#each progressItems as data (data.file)}
```
nico-martin: Since version 4.0 we have a new `progress_total` item in the `progress_callback`:

```
const pipe = await pipeline(
  "feature-extraction",
  "onnx-community/all-MiniLM-L6-v2-ONNX",
  {
    progress_callback: (i) => i.status === "progress_total" && console.log(i.progress),
  }
);
```
I would suggest to only pass the total progress doen to the UI and show that here because in the end thats what is important for the users.

packages/transformers/docs/source/tutorials/svelte.md
```diff
@@ -0,0 +1,483 @@
...
+
+    worker.addEventListener('message', onMessageReceived);
+  });
+
+  onDestroy(() => {
+    worker?.removeEventListener('message', onMessageReceived);
+  });
+
+  function onMessageReceived(e) {
+    switch (e.data.status) {
```
nico-martin, line 370: see comment on line 177

nico-martin (changes requested):
Hi @mvanhorn, Thank you very much for the PR. I love to see more examples for different integrations!

There are a couple of small this I mentioned in the files. But otherwise the Transformers.js integration looks solid.

packages/transformers/docs/source/tutorials/svelte.md
```diff
@@ -0,0 +1,483 @@
...
+
+- [Node.js](https://nodejs.org/en/) version 18+
+- [npm](https://www.npmjs.com/) version 9+
+
+## Step 1: Initialise the project
+
+For this tutorial, we will use [Vite](https://vitejs.dev/) to initialise our project. Vite is a build tool that allows us to quickly set up a Svelte application with minimal configuration. Run the following command in your terminal:
+
+```bash
+npm create vite@latest svelte-translator -- --template svelte
```
pngwn: The canonical way to create a svelte app is with the [svelte CLI](https://svelte.dev/docs/cli/overview) `sv create`.

pngwn (changes requested):
Thanks for the PR @mvanhorn! I left an inline comment about using the `sv` CLI instead of create-vite. Other than that, the main issue with the PR is that it is using the older Svelte 3/4 syntax. The syntax changed quite a bit in version 5 which has been out for some time, so I think it makes sense to use that syntax here. 

The main change  is migrating the reactivity system to use runes.

mvanhorn:
Done in 307482e - you were right that sv is the way. Since sv create scaffolds SvelteKit projects (there's no plain-SPA template), I migrated the whole tutorial to the SvelteKit minimal template rather than just swapping the scaffold command: src/routes/+page.svelte instead of App.svelte, components under src/lib, the worker created in onMount (with a note on why, since Kit SSRs by default), and the deploy step now points at adapter-static for the Static Space.

qgallouedec:
This contribution comes from a profile that opens a massive number of pull requests generated purely by AI; @pngwn I think you can close it.

mvanhorn:
Fair call on volume, @qgallouedec. Noting for @pngwn: this one did go through two review rounds here, and the last push migrated the tutorial to sv create + SvelteKit with runes per the review comments. If the team would rather not take it I'll close it, no hard feelings either way.

mvanhorn:
Updated.

The tutorial is Svelte 5 now: `$props()`, `$state()`, `onclick`/`onchange`, and `onMount` cleanup instead of `onDestroy`. Scaffolding was already `sv create`.

Progress uses the v4 `progress_total` callback only, so the bar is total progress, not per-file.

I don't have push to `transformers.js-examples`, so the source link now goes to the Xenova svelte-translator Space instead of a missing repo path. Language list points at the FLORES-200 codes. `style:width` is still there for CSP.

PR #1600, opened 2026-03-21
fix(onnx): pass externalData to session_options when using customCache in Node.js

### LINKED ISSUE #1408: qwen3 embedding with customCache throw error
### System Info

transformers.js 3.7.2
node 22.17.0
windows 10

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension

### Description

if run on node env but not use `useFSCache`(use customCache) it will throw error

then I found useFSCache, will return filepath, customCache return uint8Array.Then will throw error

https://github.com/xenova/transformers.js/blob/main/src/models.js#L265

Error: Exception during initialization: file_size: ϵͳ�Ҳ���ָ�����ļ���: "model.onnx_data"

### Reproduction

run code as flow.
save model on `path.join(process.cwd(), 'models')`
```ts
import { pipeline, matmul, env } from '@huggingface/transformers';
import path from 'node:path';
import fs from 'fs';
(async () => {
  const CONTENT_TYPE_MAP = {
    txt: 'text/plain',
    html: 'text/html',
    css: 'text/css',
    js: 'text/javascript',
    json: 'application/json',
    png: 'image/png',
    jpg: 'image/jpeg',
    jpeg: 'image/jpeg',
    gif: 'image/gif',
  } as any;
  class FileResponse {
    filePath;
    headers;
    exists;
    status;
    statusText;
    body;
    constructor(filePath: string) {
      this.filePath = filePath;
      this.headers = new Headers();

      this.exists = fs.existsSync(filePath);
      if (this.exists) {
        this.status = 200;
        this.statusText = 'OK';

        let stats = fs.statSync(filePath);
        this.headers.set('content-length', stats.size.toString());

        this.updateContentType();

        const stream = fs.createReadStream(filePath);
        this.body = new ReadableStream({
          start(controller) {
            stream.on('data', (chunk) => controller.enqueue(chunk));
            stream.on('end', () => controller.close());
            stream.on('error', (err) => controller.error(err));
          },
          cancel() {
            stream.destroy();
          },
        });
      } else {
        this.status = 404;
        this.statusText = 'Not Found';
        this.body = null;
      }
    }


    updateContentType() {
      // Set content-type header based on file extension
      const extension = this.filePath.toString().split('.').pop()!.toLowerCase();
      this.headers.set('content-type', CONTENT_TYPE_MAP[extension] ?? 'application/octet-stream');
    }


    clone() {
      let response = new FileResponse(this.filePath);
      response.exists = this.exists;
      response.status = this.status;
      response.statusText = this.statusText;
      response.headers = new Headers(this.headers);
      return response;
    }


    async arrayBuffer() {
      const data = await fs.promises.readFile(this.filePath);
      return /** @type {ArrayBuffer} */ data.buffer;
    }


    async blob() {
      const data = await fs.promises.readFile(this.filePath);
      return new Blob([data as any], { type: this.headers.get('content-type')! });
    }

 
    async text() {
      const data = await fs.promises.readFile(this.filePath, 'utf8');
      return data;
    }

 
    async json() {
      return JSON.parse(await this.text());
    }
  }
  class OnlyCopyFileCache {
    path;
    constructor(path: string) {
      this.path = path;
    }

    async match(request: string) {
      let filePath = path.resolve(this.path, request);
      let file = new FileResponse(filePath);

      if (file.exists) {
        return file;
      } else {
        return undefined;
      }
    }

    async put(request: string, response: Response, progress_callback?: any) {
      throw new Error('not use');
    }
  }
  // throw error
  env.useFS = false;
  env.localModelPath = path.join(process.cwd(), 'models');
  env.allowLocalModels = false;
  env.allowRemoteModels = true;
  env.cacheDir = path.join(process.cwd(), 'models');
  env.customCache = new OnlyCopyFileCache(path.join(process.cwd(), 'models'));
  env.useBrowserCache = false;
  env.useFSCache = false;
  env.useCustomCache = true;
  // work
  // env.cacheDir = path.join(process.cwd(), 'models');
  const extractor = await pipeline('feature-extraction', 'onnx-community/Qwen3-Embedding-0.6B-ONNX', {
    dtype: 'fp32',
  });
  const output = await extractor(['1'], {
    pooling: 'last_token',
    normalize: true,
  });
  console.log('run?', output);
})();

```

s-zx:
## Problem

When a model uses external data (e.g. `model.onnx_data`) and the user provides a `customCache` (which returns a `Uint8Array`/`Response` rather than a filesystem path), loading fails in Node.js with:

```
file_size: system cannot find the specified path: "model.onnx_data"
```

## Root Cause

`session.js` only sets `session_options.externalData` when `!IS_NODE_ENV`.

In Node.js with `useFSCache` this is intentional — the ONNX runtime locates `.onnx_data` files automatically relative to the on-disk model path.

But when `customCache` is used the data arrives as an in-memory `Uint8Array` (not a file path), so the runtime has no way to discover the sidecar files on disk.

## Fix

Check whether any resolved external-data entry is a `{path, data}` object (i.e. in-memory bytes from a custom cache). If so, pass `externalData` to `session_options` even in Node.js:

```javascript
// Before
if (externalData.length > 0 && !apis.IS_NODE_ENV) {
    session_options.externalData = externalData;
}

// After
if (externalData.length > 0) {
    if (!apis.IS_NODE_ENV || externalData.some(item => typeof item !== 'string')) {
        session_options.externalData = externalData;
    }
}
```

Fixes #1408

### DIFF
--- a/packages/transformers/src/models/session.js
+++ b/packages/transformers/src/models/session.js
@@ -122,7 +122,7 @@ async function getSession(
         // However, when using a custom cache (which returns Uint8Array instead of a
         // file path), the ONNX runtime cannot find the files on disk, so we must
         // supply the data in-memory via session_options.externalData.
-        if (!apis.IS_NODE_ENV || externalData.some(item => typeof item !== 'string')) {
+        if (!apis.IS_NODE_ENV || externalData.some(item => item instanceof Uint8Array)) {
             session_options.externalData = externalData;
         }
     }


xenova:
after a bit more digging, perhaps the `typeof item !== 'string'` is better because the external data format we use is
```
Promise<Array<string | {
    path: string;
    data: Uint8Array;
}>
```

PR #1609, opened 2026-03-27
Emit progress_total events from PreTrainedModel.from_pretrained()

### LINKED ISSUE #1052: Provide file size information in advance to enable single file download progress bar
### Feature request

Return file size information prior to model file download to enable better UI when downloading multiple model files.

After loading the config files, we fire a single 'session_info' (name subject to change) which contains an object like this:
```
{
// [file name] => [info]
  encoder_model: {...},
  decoder_model_merged: {...},
}
```
where each info element contains info like file name, file size, url, etc.

Clients can then use this to do different style of downloading UI (i.e. total percentage downloaded of all model files)

### Motivation

As we look to integrate transformers.js with more UI we may want to have a more predictable experience downloading models. Right now the callbacks don't have enough information for a single progress bar downloading the model.

### Your contribution

I may be able to help code this.

JacobiusMakes:
## Summary

Closes #1052.

When users call `PreTrainedModel.from_pretrained()` directly (rather than through `pipeline()`), they currently have no way to know the total size of all model files before downloads begin. This makes it impossible to render a single aggregate progress bar.

This PR adds the same `progress_total` event infrastructure that already exists in `pipeline()` to `PreTrainedModel.from_pretrained()`:

- After resolving the model config but **before** starting downloads, gathers file metadata (sizes) for all expected model files via lightweight Range/HEAD requests using the existing `get_model_files()` and `get_file_metadata()` utilities
- Wraps the user's `progress_callback` to emit `progress_total` events on each `progress` update, containing aggregate `loaded`/`total` byte counts plus a per-file breakdown in `files`
- Follows the exact same pattern already used in `pipelines.js` (lines 139-178)
- Fails gracefully: if metadata fetching fails (e.g., local-only models), downloads proceed normally without total progress tracking

### Example usage

```js
const model = await AutoModel.from_pretrained('model-id', {
  progress_callback: (info) => {
    if (info.status === 'progress_total') {
      // info.progress  — 0..100 across all files
      // info.loaded    — total bytes loaded so far
      // info.total     — total bytes expected
      // info.files     — per-file { loaded, total } map
      updateProgressBar(info.progress);
    }
  },
});
```

## Test plan

- [x] Verified `prettier` formatting passes
- [x] Ran `tests/utils/hub.test.js` — all pass
- [x] Ran `tests/utils/model_registry.test.js` — all pass
- [x] Syntax check (`node -c`) passes
- [ ] Manual test with a multi-file model to verify `progress_total` events fire correctly

## Changes

**`packages/transformers/src/models/modeling_utils.js`** — Added ~60 lines to `PreTrainedModel.from_pretrained()` that gather file metadata upfront and wrap the progress callback to emit `progress_total` events.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/transformers/src/models/modeling_utils.js
+++ b/packages/transformers/src/models/modeling_utils.js
@@ -39,6 +39,7 @@ import { logger } from '../utils/logger.js';
 import { DynamicCache } from '../cache_utils.js';
 import { get_model_files } from '../utils/model_registry/get_model_files.js';
 import { get_file_metadata } from '../utils/model_registry/get_file_metadata.js';
+import { MODEL_TYPES } from './session_config.js';
 
 /**
  * Converts an array or Tensor of integers to an int64 Tensor.
@@ -84,25 +85,7 @@ export function boolTensor(value) {
     return new Tensor('bool', [value], [1]);
 }
 
-export const MODEL_TYPES = {
-    EncoderOnly: 0,
-    EncoderDecoder: 1,
-    Seq2Seq: 2,
-    Vision2Seq: 3,
-    DecoderOnly: 4,
-    DecoderOnlyWithoutHead: 5,
-    MaskGeneration: 6,
-    ImageTextToText: 7,
-    Musicgen: 8,
-    MultiModality: 9,
-    Phi3V: 10,
-    AudioTextToText: 11,
-    AutoEncoder: 12,
-    ImageAudioTextToText: 13,
-    Supertonic: 14,
-    Chatterbox: 15,
-    VoxtralRealtime: 16,
-};
+export { getSessionsConfig, MODEL_TYPES } from './session_config.js';
 
 const MODEL_TYPE_CONFIG = {
     [MODEL_TYPES.DecoderOnly]: {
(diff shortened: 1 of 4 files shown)


nico-martin:
I ran into the exact same ptoblem too yesterday :D
Thank you so much for making a PR. Will check it out asap!

JacobiusMakes:
Thanks for the thorough review! Both issues fixed in e28ce2c:

**1) config.json progress stuck below 100%** — config.json is now pre-marked as fully loaded in `files_loading` since `AutoConfig.from_pretrained()` fetches it before the tracker is set up.

**2) Duplicate progress_total with pipeline()** — Wrapped callbacks now carry a `_progress_total_wrapped` flag. `from_pretrained()` checks for it and skips wrapping if pipeline() already set it up. pipeline()'s wrapper also sets this flag for consistency.

nico-martin:
There is still a problem with the prod build:

```
tsc --build

Error: src/models/modeling_utils.js(430,53): error TS2339: Property '_progress_total_wrapped' does not exist on type 'ProgressCallback'.
Error: src/pipelines.js(180,33): error TS2339: Property '_progress_total_wrapped' does not exist on type '(info: ProgressInfo) => void'.
 ELIFECYCLE  Command failed with exit code 1.
/home/runner/work/transformers.js/transformers.js/packages/transformers:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @huggingface/transformers@4.0.0-next.10 build: `node scripts/build.mjs && pnpm typegen`
Exit status 1
 ELIFECYCLE  Command failed with exit code 1.
Error: Process completed with exit code 1.
```

https://github.com/huggingface/transformers.js/actions/runs/23668078336/job/69010757045?pr=1609

xenova:
thanks for the PR! I think a better way to handle this is with a wrapper function class instance which can detect if we need to wrap the object using an `instanceof` test. this means we don't need to attach hacky properties like `_progress_total_wrapped` to it.

xenova:
merged via https://github.com/huggingface/transformers.js/pull/1615

Issue #264, opened 2023-08-28
[Question] TypeScript rewrite


Lantianyou:
Hi Joshua. I found your idea is extremely exciting.
I am a frontend developer who has worked on TypeScript professionally for three years. Would you mind me doing a TypeScript re-write, so this npm package can have a better DX.

I just forked this repo. https://github.com/Lantianyou/transformers.js

Lantianyou:
I do not think TS is always the best solution either, however, it would be nice to have it for this project. For example, this project use BigInt, it assumes ES 2020 or later. I think TypeScript transpiler can be a great use case here.

quantuminformation:
TS will die if js ever gets types, its at proposal stage atm

kungfooman:
> I do not think TS is always the best solution either, however, it would be nice to have it for this project.

No, it would be as ugly as it gets, it just makes everything 200 times more complicated while not actually improving anything (actually decreasing the readability) and prevents everyone from quickly testing and making PR's because you have to fumble with a build process in different scenarious, even though you just want to test a few lines of new code.

quantuminformation:
yup, why do you think I dumped ts for my flagship patterns:

https://github.com/quantuminformation/vanillajs-patterns

Lantianyou:
I believe type proposal for ES is in a very far future.

sigma-andex:
I give my 👍 to the _rewrite this in typescript_ proposal. While I agree that TS config can be bit annoying - not to say a pain the a** - I do think that a simple setup is possible that works for most people checking out this repo (at least if you are not using something abstruse like Nix). I think the benefits ( type-safety, better editor auto-completion, using the language that most modern web libraries are written in, ...) far outweigh the config hiccups.

Lantianyou:
Thank you @sigma-andex for your support.

kungfooman:
> I think the benefits ( type-safety, better editor auto-completion, using the language that most modern web libraries are written in, ...) far outweigh the config hiccups.

You are just exposing that you don't know how current technology works. JSDoc and TypeScript is using the *same* language server for IntelliSense.

You gain nothing but have a PITA now. *you are sold*

The amount of TypeScript fanboys that are convinced how much better TS is while not knowing anything about the underlying technology is really getting worrisome...

quantuminformation:
TypeScript is great if you like low signal/noise ratios

marek-hanzal:
[...]

PR #1613, opened 2026-03-29
transformers.ts

sroussey:


sroussey:
I did in steps so one step was renaming only, the next renaming test files, and the third is actually changing the types.

nico-martin:
@sroussey, Thanks for the PR!

@xenova and I discussed this topic at length a few weeks ago and came to the conclusion that we don't really see any advantages to TS over our current setup.
JSDoc (and its IDE integrations) offers us all the benefits when it comes to type checking (and the type system in general), while the compilation step remains optional. This means that the library’s source files can be executed directly in any JavaScript environment. In our opinion, this advantage outweighs the few benefits that TS would provide (mainly the syntax imo).

Also see https://github.com/huggingface/transformers.js/issues/264

sroussey:
The compilation step is needed for an old JS env that wants to run the source, though i have no idea who on Earth would do that. Users of the library don't use source since you bundle.

But i can just use the source `bun src/transformers.ts` directly, though I use canonical file endings for TS, so in node you have to `node --import ./scripts/register-ts-resolve.mjs ./src/transformers.ts`. There is no compile step unless you have TS enums (and there is a setting that will disallow anything that node would error on). I think that is the only typescript thing node does not support. This codebase seems to work just fine.

You could, of course, not bundle and just use source, in which case, yes old JS engines would not be happy. And TS has some stuff where it does not complain about anything in .d.ts files generally, but if module resolution was digging into sources files, you had better have the tightest of rules.

Anyhow, as an exercise you can look through the commit that actually changes types. It shows some interesting stuff, particularly where the jsdoc types are wrong.

I'll just close it here and use it to propose changes to the types in jsdocs later.

nico-martin:
I see what you mean. I think it makes sense to discuss the subject again.

kungfooman:
> The compilation step is needed for an old JS env that wants to run the source, though i have no idea who on Earth would do that. Users of the library don't use source since you bundle.

You don't seem to understand modern ESM with importmaps. I check out a branch I want to test and press Ctrl+R in any modern browser - done, I'm testing the latest changes.

Extra bundle steps are not needed, no one needs to spawn shells if they don't want to, no one needs to run this or that command to finally build something.

And ESM and TS literally share the same language server - if you find errors in TS, you would see the same errors in JSDoc. Instead of rewriting everything, you could just fix the code as it is.

But yea, I know you just love TS "with warts an all" and that is the entire motivation. Not technical improvement, but it "must be TS for whatever invalid reasons referring to old JS env's or whatever comes to mind to make some irrelevant points".

sroussey:
> ESM with importmaps

Indeed, not something I use, so i did go learn about it. Thank you. 

However, I did not see how it helps when transformers.js only exports "./dist/transformers.web.js" which is part of the dist directory which is bundled/built at the time of packaging for npm.

Even pnpm link (alone) does not help here.

I am not a fan of the whole, edit/watch-build path myself, so i have a script that alters package.json files once when i start developing so that the exports point to the source, and then i bun link (could be npm or pnpm link as well) the packages. Then i can edit transformers.js which is used my by workglow lib which is used by my workflow builder site, and i don't have build anywhere.

Issue #1630, opened 2026-04-02
is_pipeline_cached("feature-extraction") incorrect


sroussey:
### System Info

transformers@4

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension
- [ ] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

ModelRegistry.is_pipeline_cached wrong for "feature-extraction"

### Reproduction

```ts
const modelId = "onnx-community/Qwen3-Embedding-0.6B-ONNX";
const task = "feature-extraction";
const options = {device:"webgpu", dtype:"q4f16"};
await pipeline(task, modelId, options);
await ModelRegistry.is_pipeline_cached(task, modelId, options);
```

```js
allCached:false,
files:[{file: 'config.json', cached: true}, {file: 'onnx/model_quantized.onnx', cached: true},{file: 'generation_config.json', cached: **false**},{file: 'tokenizer.json', cached: true},{file: 'tokenizer_config.json', cached: true}
```

apparently it expects generation_config.json to be loaded.

xenova:
Indeed, generation_config should not be needed. @nico-martin any ideas?

nico-martin:
Solved with a config change. See https://github.com/huggingface/transformers.js/pull/1631

PR #1631, opened 2026-04-02
fix: is_pipeline_cached for some models

sroussey:
Enhance ModelRegistry file retrieval by adding model_type_override parameter to get_files and get_model_files functions. This allows for more flexible model type resolution when loading files for pipelines.

Problem

`is_pipeline_cached("feature-extraction", "onnx-community/Qwen3-Embedding-0.6B-ONNX", ...)` incorrectly required
`generation_config.json` to be cached, even though the pipeline never fetches it.

The mismatch: get_pipeline_files used resolve_model_type(config) which reads config.architectures[0] = "Qwen3ForCausalLM" → DecoderOnly → has optional_configs: { generation_config }. But pipeline("feature-extraction", ...) uses AutoModel, which iterates its MODEL_CLASS_MAPPINGS in order and finds "qwen3" in MODEL_MAPPING_NAMES_DECODER_ONLY first → loads Qwen3Model → DecoderOnlyWithoutHead → no optional_configs. So generation_config.json is never fetched, but is_pipeline_cached thought it
  was required.

  Fix

Added resolveEffectiveModelType(taskConfig, config) in get_pipeline_files.js. Instead of using config.architectures to infer the model type, it simulates what the pipeline's auto class actually does: iterates autoClass.MODEL_CLASS_MAPPINGS using config.model_type to find the first matching class (e.g. Qwen3Model), then looks that up in MODEL_TYPE_MAPPING (→  DecoderOnlyWithoutHead). This resolved type is passed as model_type_override down through get_files → get_model_files, replacing the resolve_model_type call there.

Cross-architecture detection (e.g. Qwen3VLForCausalLM loading a ForConditionalGeneration model for text-generation) is also replicated to avoid regression.

### DIFF
--- a/packages/transformers/src/utils/model_registry/get_files.js
+++ b/packages/transformers/src/utils/model_registry/get_files.js
@@ -14,6 +14,7 @@ import { get_processor_files } from './get_processor_files.js';
  * @param {string|null} [options.model_file_name=null|null] Override the model file name (excluding .onnx suffix)
  * @param {boolean} [options.include_tokenizer=true] Whether to check for tokenizer files (set to false for vision-only models)
  * @param {boolean} [options.include_processor=true] Whether to check for processor files
+ * @param {number|null} [options.model_type_override=null] Override the resolved model type (one of MODEL_TYPES). Passed through to get_model_files.
  * @returns {Promise<string[]>} Array of file paths that will be loaded
  */
 export async function get_files(
@@ -25,9 +26,10 @@ export async function get_files(
         model_file_name = null,
         include_tokenizer = true,
         include_processor = true,
+        model_type_override = null,
     } = {},
 ) {
-    const files = await get_model_files(modelId, { config, dtype, device, model_file_name });
+    const files = await get_model_files(modelId, { config, dtype, device, model_file_name, model_type_override });
 
     if (include_tokenizer) {
         const tokenizerFiles = await get_tokenizer_files(modelId);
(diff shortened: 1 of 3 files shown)


sroussey:
To fix #1630 

@nico-martin

xenova:
I don't necessarily think the solution proposed is the best; it seems overcomplicated. Our model type mapping should be setting this model correctly. The issue is that the model has model_type: qwen3, and architectures: [Qwen3ForCausalLM], so it thinks it's a text-generation model (understandably).

However, the sample python code converts it to `Qwen3Model` (which it should be):
```
>>> from transformers import AutoModel
>>> model = AutoModel.from_pretrained('onnx-internal-testing/tiny-random-Qwen3ForCausalLM')
Loading weights: 100%|█████████████████████████████████████████| 24/24 [00:00<00:00, 8427.23it/s]
>>> type(x)
<class 'transformers.models.qwen3.modeling_qwen3.Qwen3Model'>
```

I updated the architecture to `Qwen3Model` at https://huggingface.co/onnx-community/Qwen3-Embedding-0.6B-ONNX/commit/c25a394dd583836952667c12f008335071b3f43d

So, it should work now (just refresh your cache)

nico-martin:
I can confirm that #1630 does work now [after the config.json change](https://huggingface.co/onnx-community/Qwen3-Embedding-0.6B-ONNX/commit/c25a394dd583836952667c12f008335071b3f43d).

```javascript
const modelId = "onnx-community/Qwen3-Embedding-0.6B-ONNX";
const task = "feature-extraction";
const options = {
  device: "webgpu",
  dtype: "q4f16",
};
await pipeline(task, modelId, options);
const cached = await ModelRegistry.is_pipeline_cached(task, modelId, options);
const files = await ModelRegistry.get_pipeline_files(task, modelId, options);
console.log("cached", cached); // true
console.log("files", files); // ["config.json", "onnx/model_q4f16.onnx", "tokenizer.json", "tokenizer_config.json" ]
```

sroussey:
Great, thanks!

PR #1633, opened 2026-04-03
Align default pipeline generation parameters with python transformers

### LINKED ISSUE #1632: max_new_tokens?
### Question

In the docs there is this example:

```ts

  "text-generation",
  "onnx-community/Qwen2.5-Coder-0.5B-Instruct",
  { dtype: "q4" },
);

const messages = [
  { role: "user", content: "Write a quick sort algorithm." },
];

// Create text streamer
const streamer = new TextStreamer(generator.tokenizer, {
  skip_prompt: true,
  // Optionally, do something with the text (e.g., write to a textbox)
  // callback_function: (text) => { /* Do something with text */ },
});

// Generate a response
const result = await generator(messages, {
  max_new_tokens: 512,
  do_sample: false,
  streamer,
});
```

The default value for max_new_tokens = null. Is that supposed to mean zero? Or unlimited? In reality, it means one. If you comment it out, you just get one word:

```
Here
```

Also, what and where is printing this out??? There is no console.log, and yet...

sroussey:
Enhance pipeline configurations by setting default parameters for max_new_tokens, num_beams, do_sample, and temperature in Automatic Speech Recognition, Document Question Answering, Text Generation, and Text2Text Generation pipelines.

Closes https://github.com/huggingface/transformers.js/issues/1632

### DIFF
--- a/packages/transformers/src/pipelines/automatic-speech-recognition.js
+++ b/packages/transformers/src/pipelines/automatic-speech-recognition.js
@@ -141,6 +141,11 @@ export class AutomaticSpeechRecognitionPipeline
     )
 {
     async _call(audio, kwargs = {}) {
+        kwargs = {
+            max_new_tokens: 256,
+            num_beams: 5, 
+            ...kwargs
+        };
         switch (this.model.config.model_type) {
             case 'whisper':
             case 'lite-whisper':
--- a/packages/transformers/src/pipelines/document-question-answering.js
+++ b/packages/transformers/src/pipelines/document-question-answering.js
@@ -44,6 +44,10 @@ export class DocumentQuestionAnsweringPipeline
     )
 {
     async _call(image, question, generate_kwargs = {}) {
+        generate_kwargs = {
+            max_new_tokens: 256,
+            ...generate_kwargs
+        };
         if (Array.isArray(image)) {
             if (image.length !== 1) {
                 throw Error('Document Question Answering pipeline currently only supports a batch size of 1.');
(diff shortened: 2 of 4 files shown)


nico-martin:
There are a couple of failing tests. But besides that, changing the default behaviour (especially for the `do_sample` and the `temperature`) is definitely a breaking change. Although I really like this change, @xenova, are we sure we want that for a minor- or even patch-release?

xenova:
Hmm, yeah I didn't think about the do_sample: true default. this could be a bit confusing. the max_new_tokens default I think it pretty safe, because the current approach (inherited from python version) is to use `max_length: 20`, which doesn't care about current sequence length. it could only generate one token, then end because the prompt is longer than 20 tokens.

xenova:
if anything, this should have been part of v4 😅 (and I forgot/delayed)

many models these days have their own generation config (sometimes using do_sample: true), so the behaviour should only be considered "fully-specified" if the user is the one doing the property setting

xenova (changes requested):
thanks for the PR! I think the main fix we need is `max_new_tokens`, and although `do_sample` and `temperature` are less important, we can still include them for consistency. Marking as v4.1 is okay, as this probably should have been done in v4.0

to better align with the python library (see [PR](https://github.com/huggingface/transformers/pull/38129))

maybe we should introduce `_default_generation_config` on the class itself and use when we construct the generation config to actually use.

xenova:
merged via https://github.com/huggingface/transformers.js/pull/1649
image

PR #1638, opened 2026-04-10
Cached generation improvements (+ past_key_values via pipeline function)

xenova:
This PR now allows past_key_values to be used inside the pipeline function, leading to significantly faster generation for long conversations (no unnecessary recomputation)! 🙌 

Example usage:
```js
import {
  pipeline,
  DynamicCache,

  "text-generation",
  "onnx-community/LFM2.5-350M-ONNX",
  { dtype: "q4", device: "webgpu" },
);

// NEW: Create a shared cache for multi-turn conversation
const past_key_values = new DynamicCache();

const messages = [{ role: "user", content: "Tell me a very long story about cheese." }];

// === Turn 1 ===
let start = performance.now();
const output1 = await generator(messages, {
  max_new_tokens: 1024,
  do_sample: false,
  past_key_values,
});
const reply1 = output1[0].generated_text.at(-1);
console.log(`Turn 1: ${(performance.now() - start).toFixed(2)}ms | Cache seq length: ${past_key_values.get_seq_length()}`);

// === Turn 2 ===
messages.push(reply1);
messages.push({ role: "user", content: "Answer in one word: What is 1 + 1?" });

start = performance.now();
const output2 = await generator(messages, {
  max_new_tokens: 128,
  do_sample: false,
  past_key_values,
});
const reply2 = output2[0].generated_text.at(-1);
console.log(`Turn 2: ${(performance.now() - start).toFixed(2)}ms | Cache seq length: ${past_key_values.get_seq_length()}`);

// === Turn 3 ===
messages.push(reply2);
messages.push({ role: "user", content: "Answer in one word: What is 2 + 2?" });

start = performance.now();
await generator(messages, {
  max_new_tokens: 128,
  do_sample: false,
  past_key_values,
});
console.log(`Turn 3: ${(performance.now() - start).toFixed(2)}ms | Cache seq length: ${past_key_values.get_seq_length()}`);

// Clean up
```

outputs
```
Turn 1: 4235.52ms | Cache seq length: 996
Turn 2: 117.24ms | Cache seq length: 1025
Turn 3: 72.07ms | Cache seq length: 1049
```

without setting past_key_values, you get
```
Turn 1: 4355.55ms | Cache seq length: 0
Turn 2: 293.84ms | Cache seq length: 0
Turn 3: 248.94ms | Cache seq length: 0
```

🔥 Improvements will be even more noticeable for multimodal models, since we don't need to compute image features.

### DIFF
--- a/packages/transformers/src/models/modeling_utils.js
+++ b/packages/transformers/src/models/modeling_utils.js
@@ -204,6 +204,7 @@ export class PreTrainedModel extends Callable {
     forward_params = ['input_ids', 'attention_mask'];
 
     _return_dict_in_generate_keys = null;
+
     /**
      * Creates a new instance of the `PreTrainedModel` class.
      * @param {import('../configs.js').PretrainedConfig} config The model configuration.
@@ -664,7 +665,7 @@ export class PreTrainedModel extends Callable {
      */
     _update_model_kwargs_for_generation({ generated_input_ids, outputs, model_inputs, is_encoder_decoder }) {
         // update past_key_values
-        model_inputs['past_key_values'] = this.getPastKeyValues(outputs, model_inputs.past_key_values);
+        model_inputs['past_key_values'] = getPastKeyValues(outputs, model_inputs.past_key_values);
 
         // update inputs for next run
         model_inputs['input_ids'] = new Tensor('int64', generated_input_ids.flat(), [generated_input_ids.length, 1]);
@@ -966,7 +967,7 @@ export class PreTrainedModel extends Callable {
             if (generation_config.return_dict_in_generate) {
                 if (generation_config.output_attentions) {
                     // Get attentions if they are present
-                    const token_attentions = this.getAttentions(outputs);
+                    const token_attentions = getAttentions(outputs);
                     for (const key in token_attentions) {
                         if (!(key in attentions)) {
                             attentions[key] = [];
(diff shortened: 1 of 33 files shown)


xenova:
tested and it works with gemma4 too. pretty much zero latency between calls (generation still takes some time, of course)
```
Turn 1: 4771.26ms | Cache seq length: 765
Turn 2: 1124.67ms | Cache seq length: 822
Turn 3: 1414.80ms | Cache seq length: 1135 (new image input)
```

packages/transformers/src/models/modeling_utils.js
```diff
@@ -1030,12 +1031,26 @@ export class PreTrainedModel extends Callable {
...
-        // Retrieve and dispose all final past key values (including encoder attentions)
-        const past_key_values = this.getPastKeyValues(outputs, model_inputs.past_key_values, true);
-
         // TODO: ensure all_input_ids is padded correctly...
         const sequences = new Tensor('int64', all_input_ids.flat(), [all_input_ids.length, all_input_ids[0].length]);
 
+        // Update past key values from the final forward pass
+        const past_key_values = getPastKeyValues(outputs, model_inputs.past_key_values);
+
+        // Dispose output tensors not held by the cache
```
nico-martin, line 1040: In generate(), the disposal loop disposes GPU tensors from outputs unless they are in past_key_values.
Tensors saved for return (attentions and _return_dict_in_generate_keys items) are not excluded from disposal.
This means when return_dict_in_generate (especially with output_attentions) is enabled, returned tensors can already be disposed.

packages/transformers/src/transformers.js
nico-martin, line 1: One more general comment so I'll just put it here.

I notices that this does not throw a type error:

```
const past_key_values = "hello";
const resp = await pipe(messages, {
  max_new_tokens: 1024,
  streamer,
  temperature,
  past_key_values,
});
```

Would it be possible to improve the types?
```
past_key_values?: DynamicCache | null
```
Also if provided and not DynamicCache, throw TypeError("past_key_values must be a DynamicCache").

nico-martin (commented):
I think the API is great!

```
  "text-generation",
  "onnx-community/LFM2.5-350M-ONNX",
  { dtype: "q4", device: "webgpu" },
);

const past_key_values = new DynamicCache();

const messages = [{ role: "user", content: "Tell me a very long story about cheese." }];

await generator(messages, {
  max_new_tokens: 1024,
  do_sample: false,
  past_key_values,
});
```
It's extremely easy to use and handles all the caching behind the scenes!

packages/transformers/src/models/modeling_utils.js
xenova: Indeed, this section of code is a bit redundant, but it matches the previous implementation. Basically, all of those will already be on CPU since the only tensors with location on the GPU are the KV cache elements (set via preferredOutputLocation).

So, this isn't technically an issue, as the current disposal loop doesn't do anything right now. I think I originally put it there as a fail-safe in case the user set a custom output mapping.

packages/transformers/src/transformers.js
xenova: sounds good! I added a type test for this too in [Add test case for past_key_values type](https://github.com/huggingface/transformers.js/pull/1638/commits/9960f76bbb49c9b608046cd4f078393fdf32ca0b)

xenova: & improved error message locations
image

xenova: thanks to claude, error messages are much better:

image

xenova:
decided to also put the typescript improvements in this PR (Claude and I were on a roll)... so, a bit more to review (@nico-martin 🙏 ) but totally worth it!

image

nico-martin (approved):
lgtm!

PR #1641, opened 2026-04-13
Improve tokenizer types based on input function parameters

### LINKED ISSUE #1624: types in example
### Question

In this model [README](https://huggingface.co/onnx-community/functiongemma-270m-it-ONNX), there is an example, but TS frowns:

```ts
import { AutoModelForCausalLM, AutoTokenizer } from "@huggingface/transformers";

// Load the model and tokenizer
const model_id = "onnx-community/functiongemma-270m-it-ONNX";
const tokenizer = await AutoTokenizer.from_pretrained(model_id);
const model = await AutoModelForCausalLM.from_pretrained(model_id);

const weather_function_schema = {
  type: "function",
  function: {
    name: "get_current_temperature",
    description: "Gets the current temperature for a given location.",
    parameters: {
      type: "object",
      properties: {
        location: {
          type: "string",
          description: "The city name, e.g. San Francisco",
        },
      },
      required: ["location"],
    },
  },
};

const messages = [
  {
    role: "developer",
    content: "You are a model that can do function calling with the following functions",
  },
  {
    role: "user",
    content: "What's the temperature in London?",
  },
];

const inputs = tokenizer.apply_chat_template(messages, {
  tools: [weather_function_schema],
  tokenize: true,
  add_generation_prompt: true,
  return_dict: true,
});

const output = await model.generate({ ...inputs, max_new_tokens: 512 });
const decoded = tokenizer.decode(output.slice(0, [inputs.input_ids.dims[1], null]), { skip_special_tokens: false });
console.log(decoded);
```

On the `...inputs` part the error is:

```
Spread types may only be created from object types.ts(2698)
const inputs: string | number[] | number[][] | Tensor | { input_ids: number[] | number[][] | Tensor; attention_mask: number[] | number[][] | Tensor; token_type_ids?: number[] | number[][] | Tensor; }
⚠ Error (TS2698) 
```

and on `output.slice(`
```
Property 'slice' does not exist on type 'ModelOutput | Tensor'.
  Property 'slice' does not exist on type 'ModelOutput'.ts(2339)
```

and on `inputs.input_ids`
```
Property 'input_ids' does not exist on type 'string | number[] | number[][] | Tensor | { input_ids: number[] | number[][] | Tensor; attention_mask: number[] | number[][] | Tensor; token_type_ids?: number[] | number[][] | Tensor | undefined; }'.
  Property 'input_ids' does not exist on type 'string'.ts(2339)
```

also on that second to last line, there is an error with null:
```
Type 'null' is not assignable to type 'number'.ts(2322)
```

Not sure why null is there, I am pretty sure it is a bug, but if we just remove it and assume that was the intent, I end up with:

```ts
const inputs = tokenizer.apply_chat_template(messages, {
  tools: [weather_function_schema],
  tokenize: true,
  add_generation_prompt: true,
  return_dict: true,
});
const inputsAsSomething = inputs as {
  input_ids: Tensor;
  attention_mask: number[] | number[][] | Tensor;
  token_type_ids?: number[] | number[][] | Tensor;
};
const output = await model.generate({ ...inputsAsSomething, max_new_tokens: 512 });
const outputAsTensor = output as Tensor;
const decoded = tokenizer.decode(
  outputAsTensor.slice(0, [inputsAsSomething.input_ids.dims[1]]),
  {
    skip_special_tokens: false,
  }
);
console.log(decoded);
```

So the question is... how am i supposed to know which of these types will be the result? 

Like the inputsAsSomething is the right result from apply_chat_template? 

And that inputsAsSomething.input_ids is a Tensor that has dims not an array of numbers that does not? 

And why is output a Tensor and not ModelOutput? 

(BTW: what is BaseModelOutput used for? seems to be ignored)

xenova:
Inspired by recent updates in https://github.com/huggingface/transformers.js/pull/1638#issuecomment-4232359558, we now can have better types for tokenization input.

```js
const a = tokenizer('text', { return_tensor: true });
a.input_ids; // Tensor
a.attention_mask; // Tensor

const b = tokenizer(['text'], { return_tensor: true });
b.input_ids; // Tensor
b.attention_mask; // Tensor

const c = tokenizer('text', { return_tensor: false });
c.input_ids; // number[]
c.attention_mask; // number[]

const d = tokenizer(['text'], { return_tensor: false });
d.input_ids; // number[][]
d.attention_mask; // number[][]
```

same for apply_chat_template
```js

const a = tokenizer.apply_chat_template(conversation);
a.input_ids // Tensor
a.attention_mask // Tensor

const b = tokenizer.apply_chat_template(conversation, { return_tensor: false, return_dict: true });
b.input_ids // number[]
b.attention_mask // number[]

const c = tokenizer.apply_chat_template(conversation, { return_tensor: true, return_dict: false });
c // Tensor

const d = tokenizer.apply_chat_template(conversation, { return_tensor: false, return_dict: false });
d // number[]

const e = tokenizer.apply_chat_template(conversation, { tokenize: false });
e // string
```

Closes https://github.com/huggingface/transformers.js/issues/1624

### DIFF
--- a/packages/transformers/src/pipelines/text-generation.js
+++ b/packages/transformers/src/pipelines/text-generation.js
@@ -124,11 +124,15 @@ export class TextGenerationPipeline
             // If the input is a chat, we need to apply the chat template
             inputs = /** @type {string[]} */ (
                 /** @type {Chat[]} */ (texts).map((x) =>
-                    this.tokenizer.apply_chat_template(x, {
-                        tokenize: false,
-                        add_generation_prompt: true,
-                        ...tokenizer_kwargs,
-                    }),
+                    /** @type {string} */ (
+                        /** @type {unknown} */ (
+                            this.tokenizer.apply_chat_template(x, {
+                                tokenize: false,
+                                add_generation_prompt: true,
+                                ...tokenizer_kwargs,
+                            })
+                        )
+                    ),
                 )
             );
             // Chat template handles these already
--- a/packages/transformers/docs/plugins/preprocess.js
+++ b/packages/transformers/docs/plugins/preprocess.js
@@ -25,12 +25,47 @@ function stripLeadingGenericParams(expr) {
   return depth === 0 ? expr.slice(i) : expr;
 }
 
+function hasTopLevelConditional(expr) {
+  let angle = 0,
+    paren = 0,
+    brace = 0,
+    bracket = 0;
+
+  for (let i = 0; i < expr.length; ++i) {
+    const ch = expr[i];
+    if (ch === "<") angle++;
+    else if (ch === ">") angle = Math.max(0, angle - 1);
+    else if (ch === "(") paren++;
+    else if (ch === ")") paren = Math.max(0, paren - 1);
+    else if (ch === "{") brace++;
+    else if (ch === "}") brace = Math.max(0, brace - 1);
+    else if (ch === "[") bracket++;
+    else if (ch === "]") bracket = Math.max(0, bracket - 1);
+    else if (ch === "?" && angle === 0 && paren === 0 && brace === 0 && bracket === 0) {
+      return true;
+    }
+  }
+  return false;
+}
+
 function transformType(expr) {
   let result = expr,
     prev = "";
   while (result !== prev) {
     prev = result;
     result = stripLeadingGenericParams(result); // <T extends X>(...) -> (...)
+    if (/^Promise<function\s*\(/.test(result)) {
+      result = "Promise.<Function>";
+      continue;
+    }
+    if (/^function\s*\(/.test(result)) {
+      result = "Function";
+      continue;
+    }
+    if (hasTopLevelConditional(result)) {
+      result = "any";
+      continue;
+    }
     result = result
       .replace(/import\([^)]+\)((?:\.\w+)*)/g, (_, p) => p?.slice(1) || "any") // import() -> Type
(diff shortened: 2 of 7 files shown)


packages/transformers/src/pipelines/text-generation.js
```diff
@@ -124,11 +124,15 @@ export class TextGenerationPipeline
...
-                    }),
+                    /** @type {string} */ (
+                        /** @type {unknown} */ (
+                            this.tokenizer.apply_chat_template(x, {
+                                tokenize: false,
+                                add_generation_prompt: true,
+                                ...tokenizer_kwargs,
+                            })
+                        )
+                    ),
```
xenova, line 135: even putting `tokenize: false` after, it doesn't seem to pick up, so we need to unknown -> string it

nico-martin: I think thats fine since its just a local workaround and does not affect the API 👍

nico-martin (approved):
love it! Those are some really solid type improvements :)

packages/transformers/src/pipelines/text-generation.js
xenova: yeah i played with it a bit more and indeed the `...tokenizer_kwargs,` causes typescript to not know that `tokenize: false`... even if I place if afterwards 🤷 
luckily, not a big deal.

Issue #1642, opened 2026-04-14
Auto device on Linux x64 fails hard when CUDA shared library is unavailable instead of falling back


alexewerlof:
### System Info

# Minimal repro

Environment
* OS: Linux x64
* CPU: AMD Ryzen (no NVIDIA CUDA runtime installed)
* Node: LTS 24.14.0
* @huggingface/transformers: 4.0.1 (or exact resolved version)

### Environment/Platform

- [ ] Website/web-app
- [ ] Browser extension
- [x] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

# Summary

When running in Node on Linux x64 with device set to `'auto'`, Transformers.js currently prioritizes `'cuda'` and fails with a hard error if CUDA shared libraries are not available. This prevents fallback to other providers (like `'webgpu'` or `'cpu'`).

# Observed error

```
Error: OrtSessionOptionsAppendExecutionProvider_Cuda: Failed to load shared library
      at new OnnxruntimeSessionHandler (node_modules/onnxruntime-node/lib/backend)
      at Immediate.<anonymous> (node_modules/onnxruntime-node/lib/backend.ts:147:)
      at process.processImmediate (node:internal/timers:504:21)
```

On Linux x64, CUDA is added to supported devices first in `onnx.js`:

```js
    switch (process.platform) {
        case 'win32': // Windows x64 and Windows arm64
            supportedDevices.push('dml');
            break;
        case 'linux': // Linux x64 and Linux arm64
            if (process.arch === 'x64') {
                supportedDevices.push('cuda');
            }
            break;
        case 'darwin': // MacOS x64 and MacOS arm64
            supportedDevices.push('coreml');
            break;
    }

    supportedDevices.push('webgpu');
    supportedDevices.push('cpu');
    defaultDevices = ['cpu'];
```

# Proposed fix

In Node, when `device` is `'auto'` and session creation fails specifically due to CUDA provider shared-library load error, retry once with executionProviders filtered to remove cuda.

Even better: `'auto'` should retry the list from the best performance preferred first to safe fallback option ('cpu').

`'auto'` returns that list which is passed directly to session options in session.js.

### Reproduction

# Run this snippet on an AMD APU

```js
import { pipeline } from '@huggingface/transformers'

await pipeline('text-generation', 'onnx-community/LFM2-1.2B-ONNX', {
  device: 'auto',
[...]

PR #1646, opened 2026-04-15
feat: export deviceToExecutionProviders and getSupportedDevices

### LINKED ISSUE #1645: Export supportedDevices / deviceToExecutionProviders from public API
### Feature request

**`deviceToExecutionProviders()`** and the **`supportedDevices`** array in `src/backends/onnx.js` are currently internal. Exporting them from the public API would let downstream libraries implement their own device fallback logic without hardcoding platform-specific device lists.

### Why this is needed

When `device="auto"` is used, transformers.js passes the full `supportedDevices` array (e.g. `['cuda', 'webgpu', 'cpu']` on Linux x64) to `InferenceSession.create()`. If a provider fails hard — such as CUDA when `libcudnn.so` is missing — onnxruntime-node crashes instead of falling through to the next provider (see #1642).

To work around this, libraries like [wandler](https://github.com/runpod-labs/wandler) need to try each device individually and catch failures. But there's no way to get the platform-specific device list from transformers.js — `deviceToExecutionProviders` and `supportedDevices` aren't exported. This forces us to hardcode our own copy of the platform detection logic, which drifts when transformers.js adds new providers or changes the ordering.

### Proposed change

Export `deviceToExecutionProviders` (and optionally `supportedDevices`) from the package's public API so consumers can call:

```js
import { deviceToExecutionProviders } from '@huggingface/transformers';

const devices = deviceToExecutionProviders('auto');
// e.g. ['cuda', 'webgpu', 'cpu'] on Linux x64
// e.g. ['coreml', 'webgpu', 'cpu'] on macOS
```

This would allow downstream code to implement per-device retry/fallback without duplicating platform detection logic.

### Relationship to #1642

Issue #1642 tracks the root cause (onnxruntime crashing instead of falling back). Even if that gets fixed inside transformers.js or onnxruntime, exporting the device list is independently useful — it lets consumers make informed decisions about device selection, logging, and error handling.

TimPietrusky:
## Summary

Export two device utility functions from the public API so downstream libraries can implement their own device fallback logic:

- **`deviceToExecutionProviders(device)`** — maps a device string (e.g. `"auto"`, `"cuda"`, `"cpu"`) to the corresponding ONNX execution providers list. Already existed internally, just not re-exported.
- **`getSupportedDevices()`** — returns a copy of the platform-specific `supportedDevices` array in priority order (e.g. `["cuda", "webgpu", "cpu"]` on Linux x64, `["coreml", "webgpu", "cpu"]` on macOS).

## Why

When `device="auto"` and a provider crashes hard (e.g. CUDA when `libcudnn.so` is missing — see #1642), downstream code needs to retry providers individually. Without access to the device list, libraries have to hardcode their own copy of the platform detection logic, which drifts when transformers.js updates.

### Example usage

```js
import { getSupportedDevices } from '@huggingface/transformers';

// Walk devices one at a time, catching provider crashes
for (const device of getSupportedDevices()) {
  try {
    model = await AutoModel.from_pretrained(modelId, { device });
    break;
  } catch (err) {
    console.warn(`device=${device} failed: ${err.message}`);
  }
}
```

## Changes

- `src/backends/onnx.js`: added `getSupportedDevices()` function
- `src/transformers.js`: re-exported `deviceToExecutionProviders` and `getSupportedDevices`

Closes #1645

### DIFF
--- a/packages/transformers/src/backends/onnx.js
+++ b/packages/transformers/src/backends/onnx.js
@@ -178,6 +178,20 @@ export function deviceToExecutionProviders(device = null) {
 }
 
 /**
+ * Get the list of supported devices for the current platform, sorted by
+ * priority/performance (best first, CPU/WASM last).
+ *
+ * This is the same ordering used internally when `device` is set to `"auto"`.
+ * Returns a shallow copy so callers can mutate or iterate without affecting
+ * library internals.
+ *
+ * @returns {import("../utils/devices.js").DeviceType[]} The supported devices.
+ */
+export function getSupportedDevices() {
+    return /** @type {import("../utils/devices.js").DeviceType[]} */ ([...supportedDevices]);
+}
+
+/**
  * Currently, Transformers.js doesn't support simultaneous loading of sessions in WASM/WebGPU.
  * For this reason, we need to chain the loading calls.
  * @type {Promise<any>}
--- a/packages/transformers/src/transformers.js
+++ b/packages/transformers/src/transformers.js
@@ -55,6 +55,9 @@ export { random } from './utils/random.js';
 
 export { DynamicCache } from './cache_utils.js';
 
+// Device utilities
+export { deviceToExecutionProviders, getSupportedDevices } from './backends/onnx.js';
+
 // Cache and file management
 export { ModelRegistry } from './utils/model_registry/ModelRegistry.js';
 


packages/transformers/src/transformers.js
```diff
@@ -55,6 +55,9 @@ export { random } from './utils/random.js';
 
 export { DynamicCache } from './cache_utils.js';
 
+// Device utilities
+export { deviceToExecutionProviders, getSupportedDevices } from './backends/onnx.js';
```
nico-martin, line 59: This adds runtime exports from the package entry point, but the checked-in generated declaration files are not updated. `packages/transformers/types/transformers.d.ts` still does not export `deviceToExecutionProviders` or `getSupportedDevices`, and `packages/transformers/types/backends/onnx.d.ts` still does not contain `getSupportedDevices`.

TypeScript consumers importing the new public API from `@huggingface/transformers` will therefore get a type error even though the runtime export exists. Please run the type generation step or add the generated declaration updates to this PR.

packages/transformers/src/backends/onnx.js
```diff
@@ -177,6 +177,20 @@ export function deviceToExecutionProviders(device = null) {
...
+ * priority/performance (best first, CPU/WASM last).
+ *
+ * This is the same ordering used internally when `device` is set to `"auto"`.
+ * Returns a shallow copy so callers can mutate or iterate without affecting
+ * library internals.
+ *
+ * @returns {import("../utils/devices.js").DeviceType[]} The supported devices.
+ */
+export function getSupportedDevices() {
+    return /** @type {import("../utils/devices.js").DeviceType[]} */ ([...supportedDevices]);
```
nico-martin, line 191: Now that `deviceToExecutionProviders` is exported publicly, returning `supportedDevices` directly for `device === "auto"` exposes mutable internal state:

```js
case 'auto':
    return supportedDevices;
```

A caller can accidentally or intentionally mutate the returned array, for example `deviceToExecutionProviders('auto').length = 0`, which changes the library's internal device order for later calls. `getSupportedDevices()` already avoids this by returning `[...supportedDevices]`; this path should do the same or otherwise return an immutable copy.

nico-martin (changes requested):
Hi @TimPietrusky 

Thanks for looking itno this. It's agreat freature, but I have two small issues that should be adressed before merging.

nico-martin:
closing this since there is no response from the author

PR #1655, opened 2026-04-17
Added `tools` to `TextGenerationPipeline`

nico-martin:
pipe(messages, { tools, ... }) is now supported directly in text-generation to align with @huggingface/transfomers:
https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/text_generation.py#L143

Top-level chat-template args are now handled in the pipeline:
- tools
- documents
- chat_template
- add_generation_prompt
- tokenizer_encode_kwargs still works and has precedence if both are provided

### DIFF
--- a/packages/transformers/src/pipelines/text-generation.js
+++ b/packages/transformers/src/pipelines/text-generation.js
@@ -27,6 +27,10 @@ function isChat(x) {
  * @typedef {Object} TextGenerationSpecificParams Parameters specific to text-generation pipelines.
  * @property {boolean} [add_special_tokens] Whether or not to add special tokens when tokenizing the sequences.
  * @property {boolean} [return_full_text=true] If set to `false` only added text is returned, otherwise the full text is returned.
+ * @property {boolean} [add_generation_prompt=true] Whether to append the assistant generation prompt when applying a chat template.
+ * @property {Object[]|null} [tools=null] A list of tools to expose to chat templates that support tool use.
+ * @property {Record<string, string>[]|null} [documents=null] A list of documents to expose to chat templates that support RAG.
+ * @property {string|null} [chat_template=null] A specific chat template (or template name) to apply.
  * @property {Object} [tokenizer_encode_kwargs] Additional keyword arguments to pass along to the encoding step of the tokenizer.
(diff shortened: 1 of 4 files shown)


packages/transformers/src/pipelines/text-generation.js
```diff
@@ -128,17 +141,29 @@ export class TextGenerationPipeline
...
+                tokenize: false,
+                add_generation_prompt: add_generation_prompt ?? true,
+                ...tokenizer_kwargs,
+            };
+
+            if (tools !== undefined && chat_template_kwargs.tools === undefined) {
+                chat_template_kwargs.tools = tools;
+            }
+
+            if (documents !== undefined && chat_template_kwargs.documents === undefined) {
```
xenova: Can maybe use our `pick` helper function

packages/transformers/src/pipelines/text-generation.js
```diff
@@ -27,6 +27,10 @@ function isChat(x) {
  * @typedef {Object} TextGenerationSpecificParams Parameters specific to text-generation pipelines.
  * @property {boolean} [add_special_tokens] Whether or not to add special tokens when tokenizing the sequences.
  * @property {boolean} [return_full_text=true] If set to `false` only added text is returned, otherwise the full text is returned.
+ * @property {boolean} [add_generation_prompt=true] Whether to append the assistant generation prompt when applying a chat template.
```
xenova: not needed at the pipeline API-level, imo. this is very specific to chat templating, as we'll always set this to true for the pipeline API.

packages/transformers/src/pipelines/text-generation.js
```diff
@@ -128,17 +141,29 @@ export class TextGenerationPipeline
...
+                chat_template_kwargs.tools = tools;
+            }
+
+            if (documents !== undefined && chat_template_kwargs.documents === undefined) {
+                chat_template_kwargs.documents = documents;
+            }
+
+            if (chat_template !== undefined && chat_template_kwargs.chat_template === undefined) {
+                chat_template_kwargs.chat_template = chat_template;
+            }
```
xenova: as mentioned, the pick function will be useful here.

xenova (commented):
thanks! can you include sample code of how this also works with the python pipeline API?

xenova (approved):
thanks!

Issue #1660, opened 2026-04-23
Missing chat template for Gemma4


thomas-0816:
### System Info

using "@huggingface/transformers": "^4.2.0"

Example code:

```
            import { pipeline, env, TextStreamer } from "/node_modules/@huggingface/transformers/dist/transformers.js";
            env.localModelPath = '/models/'
            env.backends.onnx.wasm.wasmPaths = '/node_modules/onnxruntime-web/dist/';
            env.allowRemoteModels = false;
            env.allowLocalModels = true;
            env.useBrowserCache = false;
            env.useWasmCache = false;
            const output = document.getElementById('output');
            output.innerText += `${new Date().toLocaleTimeString()} gemma-4-E2B-it-ONNX ...\n`;
            const model = await pipeline("text-generation", "gemma-4-E2B-it-ONNX", { dtype: "q4f16", device: "webgpu" });

            const messages = [
                { role: "system", content: "You are a helpful assistant." },
                { role: "user", content: "Write a quick sort algorithm in php 8, only the code." },
            ];

            output.innerText += `${new Date().toLocaleTimeString()} loaded.\n`;
            const result = await model(messages, {
                temperature: 1.0, top_k: 64, top_p: 0.95, min_p: 0.0,
                max_new_tokens: 4096,
                do_sample: true,
                streamer: new TextStreamer(model.tokenizer, {
                    skip_prompt: false, skip_special_tokens: true,
                    callback_function: (text) => output.innerText += text
                }),
            });
            output.innerText += `\n${new Date().toLocaleTimeString()} done.`;
            model.dispose();
            console.log(result);
```

gives:

```
transformers.js:15313 Uncaught Error: Cannot use apply_chat_template() because tokenizer.chat_template is not set and no template argument was passed! For information about writing templates and setting the tokenizer.chat_template attribute, please see the documentation at https://huggingface.co/docs/transformers/main/en/chat_templating
    at transformers_gemma.html:27:34
```

Qwen models have "chat_template" defined in tokenizer_config.json, but I can't find it in https://huggingface.co/onnx-community/gemma-4-E2B-it-ONNX/blob/main/tokenizer_config.json

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension
- [ ] Server-side (e.g., Node.js, Deno, Bun)
- [ ] Desktop app (e.g., Electron)
- [ ] Other (e.g., VSCode extension)

### Description

Error on text generation

### Reproduction

[...]

PR #1662, opened 2026-04-24
added fallback to chat_template_jinja if chat_template == null

nico-martin:
I had a problem with Gemma 4 E2B:
```Uncaught Error: Cannot use apply_chat_template() because tokenizer.chat_template is not set and no template argument was passed! For information about writing templates and setting the tokenizer.chat_template attribute, please see the documentation at ...```

The problem is that `chat_template` is missing in https://huggingface.co/onnx-community/gemma-4-E2B-it-ONNX/blob/main/tokenizer_config.json

Same issue is also described here: https://github.com/huggingface/transformers.js/issues/1660

This fix adds a fallback to `loadTokenizer`  to resolve a missing chat_template during from_pretrained with this fallback order:

1. existing tokenizer_config.chat_template (unchanged behavior)
2. tokenizer_config.chat_template_jinja
3. options.config.chat_template_jinja (covers pipeline-loaded model config)
4. chat_template.jinja fetched from the repo via getModelText(...)

If a non-empty fallback template is found, it is assigned to tokenizerConfig.chat_template before tokenizer construction, so apply_chat_template(...) stays synchronous and unchanged.

### DIFF
--- a/packages/transformers/src/tokenization_utils.js
+++ b/packages/transformers/src/tokenization_utils.js
@@ -14,6 +14,7 @@ import { max } from './utils/maths.js';
 import { Tensor } from './utils/tensor.js';
 import { logger } from './utils/logger.js';
 import { CHAT_TEMPLATE_NAME } from './utils/constants.js';
+import { get_file_metadata } from './utils/model_registry/get_file_metadata.js';
 import { get_tokenizer_files } from './utils/model_registry/get_tokenizer_files.js';
 
 /**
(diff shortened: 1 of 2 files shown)


nico-martin:
Solved with adding chat_template to tokenizer_config again: https://huggingface.co/onnx-community/gemma-4-E2B-it-ONNX/commit/9f4bef82ea6e296bc69f8a2f5939f73af81b07a6

PR #1669, opened 2026-05-01
Fix num_logits_to_keep default in decoder_forward and add get_available_devices()

### LINKED ISSUE #1643: get_available_devices()
### Feature request

Add a function to get all the possible valid values of `device` in a given environment (e.g. node, electron, browser).

### Motivation

We have `ModelRegistery.get_available_dtypes()` which is a huge help in user onboarding:

- Instead of guessing the best dtype for a given machine, we can let the users choose.
- We can try different values and see which one works (fallback mechanism).

But when it comes to `device`, we're left with guessing.

- `auto` has its own problems (#1624)
- Without a list, the code for trying fallbacks relies on hard coded strings

The value is available as `suportedDevices` in [onnx.js](https://github.com/huggingface/transformers.js/blob/main/packages/transformers/src/backends/onnx.js) but not exposed externally.

### Your contribution

I could implement a PR if needed.

### LINKED ISSUE #1666: Gemma 4 generation passes num_logits_to_keep=0 in decoder_forward, causing full-prompt logits memory blowup
### Bug

Gemma 4 generation can pass `num_logits_to_keep=0` into the ONNX decoder session, which causes full-prompt logits to be computed during generation. For long prompts this wastes a large amount of memory and can contribute to OOMs.

### Evidence from current `main`

In `packages/transformers/src/models/modeling_utils.js`, `decoder_prepare_inputs_for_generation()` correctly sets `num_logits_to_keep=1` for generation:

```js
// During generation, only the last token's logits are needed. Setting num_logits_to_keep=1
// avoids computing logits for the entire sequence, significantly reducing memory usage.
const session = self.sessions['decoder_model_merged'] ?? self.sessions['model'];
if (session?.inputNames.includes('num_logits_to_keep') && !model_inputs.num_logits_to_keep) {
    model_inputs.num_logits_to_keep = new Tensor('int64', [1n], []);
}
```

But `decoder_forward()` has the same comment and then inserts `0n` when the input is missing:

```js
if (session.inputNames.includes('num_logits_to_keep') && !new_model_inputs.num_logits_to_keep) {
    // ... During generation, the default is 1 ...
    new_model_inputs.num_logits_to_keep = new Tensor('int64', [0n], []);
}
```

For Gemma 4/Gemma3n conditional generation, `Gemma3nForConditionalGeneration.forward()` calls `decoder_forward(this, ..., true)` directly. In practice the ONNX feed reaches `decoder_model_merged` with `num_logits_to_keep=0`.

### Reproduction context

We hit this in [Wandler](https://wandler.ai), an OpenAI-compatible local inference server powered by Transformers.js, while running `onnx-community/gemma-4-E4B-it-ONNX` with `@huggingface/transformers@4.2.0`.

Wandler tracking issue for removing our temporary workaround once this is fixed upstream:

- https://github.com/runpod-labs/wandler/issues/40

The same code path is relevant to the promoted Gemma 4 browser assistant pattern using:

```js
pipeline("text-generation", "onnx-community/gemma-4-E2B-it-ONNX", {
  dtype: "q4f16",
  device: "webgpu",
})
```

### Measured impact

For a ~20k token Gemma 4 prompt:

- vocab size: 262144
- full-prompt logits estimate: ~20 GB
- after forcing `num_logits_to_keep=1` at the ONNX feed, that logits waste disappears

We also separately found a full-attention prefill allocation issue for long prompts, but this issue is only about the logits side: generation should not request full prompt logits.

### Expected behavior

During generation, Gemma 4 decoder sessions with `num_logits_to_keep` should receive `1`, not `0`, unless the user explicitly requested full-sequence logits.

### Possible fix

`decoder_forward()` should not default `num_logits_to_keep` to `0n` for generation paths. It likely needs to preserve an explicit user value but use `1n` when called from generation. At minimum, the current comment and behavior conflict.

### Version checked

- `@huggingface/transformers`: 4.2.0
- upstream `main` inspected at commit `f7487c737aa8cafbc106c9adf69dc9578c8f3fe0`

Suh0161:
Closes #1666, closes #1643.

### Fix: `num_logits_to_keep` defaults to `0n` instead of `1n` (#1666)

`decoder_forward()` had a comment correctly explaining that `num_logits_to_keep=1` reduces memory during generation, but the code set `0n` — causing the ONNX model to compute logits for the **entire prompt** instead of just the last token. For Gemma 4 with a 20k token prompt and 262k vocabulary this wastes ~20 GB of memory and can cause OOM crashes.

**Fix:** Change `[0n]` → `[1n]` to match both the comment and the behavior of `decoder_prepare_inputs_for_generation()`.

### Feature: `get_available_devices()` (#1643)

The `supportedDevices` list already existed inside the ONNX backend but was never exposed to users. This adds a public `get_available_devices()` function:

```js
import { get_available_devices } from '@huggingface/transformers';

const devices = get_available_devices();
// Node.js (Windows):   ['dml', 'webgpu', 'cpu']
// Node.js (Linux x64): ['cuda', 'webgpu', 'cpu']
// Browser (WebGPU):    ['webgpu', 'wasm']
// Browser (no WebGPU): ['wasm']


---

Also note: **#1182 already has open PR #1190**, so I skipped that one to avoid duplicating work.

### DIFF
--- a/packages/transformers/src/backends/onnx.js
+++ b/packages/transformers/src/backends/onnx.js
@@ -154,6 +154,25 @@ if (ORT_SYMBOL in globalThis) {
 const InferenceSession = ONNX.InferenceSession;
 
 /**
+ * Returns the list of devices available in the current environment, sorted by priority/performance.
+ *
+ * **Example:** Check available devices before loading a model.
+ * ```javascript
+ * import { get_available_devices } from '@huggingface/transformers';
+ *
+ * const devices = get_available_devices();
+ * // Node.js (Windows):  ['dml', 'webgpu', 'cpu']
+ * // Node.js (Linux x64): ['cuda', 'webgpu', 'cpu']
+ * // Browser (WebGPU):   ['webgpu', 'wasm']
+ * // Browser (no WebGPU): ['wasm']
+ * ```
+ * @returns {import("../utils/devices.js").DeviceType[]} The list of available devices.
+ */
+export function get_available_devices() {
+    return [...supportedDevices];
+}
+
+/**
  * Map a device to the execution providers to use for the given device.
  * @param {import("../utils/devices.js").DeviceType|"auto"|null} [device=null] (Optional) The device to run the inference on.
  * @returns {ONNXExecutionProviders[]} The execution providers to use for the given device.
--- a/packages/transformers/src/models/modeling_utils.js
+++ b/packages/transformers/src/models/modeling_utils.js
@@ -1325,7 +1325,7 @@ export async function decoder_forward(self, model_inputs, is_encoder_decoder = f
         // logits will be calculated. During generation, the default is 1 because only the logits of the last
         // prompt token are needed for generation. For long sequences, the logits for the entire sequence may
         // use a lot of memory so, setting `num_logits_to_keep=1` will reduce memory footprint significantly.
-        new_model_inputs.num_logits_to_keep = new Tensor('int64', [0n], []);
+        new_model_inputs.num_logits_to_keep = new Tensor('int64', [1n], []);
     }
 
     // Unpack the `past_key_values` object into model inputs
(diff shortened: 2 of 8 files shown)


Suh0161:
Closing in favor of #1670, which contains this work in the same commit stack.

PR #1668, opened 2026-05-01
Add start/end character offsets to token classification and return_offsets_mapping to tokenizer

Suh0161:
**Closes #425, closes #633.**

Two related feature-parity gaps with the Python `transformers` library, fixed in one PR since they share the same underlying mechanism (character-offset computation).

1. `return_offsets_mapping` on the tokenizer (#425)

`PreTrainedTokenizer` now accepts `return_offsets_mapping: true`:

```js
const encoding = tokenizer('My name is Sarah', {
  return_tensor: false,
  return_offsets_mapping: true,
});
// encoding.offset_mapping → [[0,0],[0,2],[3,7],[8,10],[11,16],[0,0]]

PR #1670, opened 2026-05-01
Auto device fallback on provider failure and tokenizer_options passthrough in Text2TextGenerationPipeline

### LINKED ISSUE #425: [Feature request] Return offset mapping using tokenizer
**Return offset mapping using tokenizer**

Hi guys, awesome work with this project, it is helping us a lot! Is it possible to add the parameter "return_offset_mapping" to the Tokenizer class, just like in the Transformers python library?

**Reason for request**

The returned offsets are crucial for our project in Named Entity Recognition, as we need to associate the predicted tags with the original sentence (prior to tokenization). Currently we can only associate the predicted tags with the list of subtokens, as they are a 1-to-1 association, and we know that the list of subtokens ignore some elements, such as whitespaces, which are present in the original sentence.

**Additional context**

Reference for "return_offset_mapping" on Transformers python library: https://huggingface.co/docs/transformers/v4.35.2/en/main_classes/tokenizer#transformers.PreTrainedTokenizerFast.__call__.return_offsets_mapping

### LINKED ISSUE #633: Is 'aggregation_strategy' parameter available for token classification pipeline?
### Question

Hi. I have question.

From HuggingFace Transformers documentation, they have **'aggregation_strategy'** parameter in token classification pipeline. [Link](https://huggingface.co/docs/transformers/en/main_classes/pipelines#transformers.TokenClassificationPipeline.aggregation_strategy)
Need to know in this library provide this parameter?

Thanks.

### LINKED ISSUE #1096: Allow configuring the tokenizer in `Text2TextGenerationPipeline`
### Feature request

The current behavior has hardcoded options for the tokenizer https://github.com/huggingface/transformers.js/blob/main/src/pipelines.js#L761-L775

### Motivation

It would be nice to stick with the generic `pipeline` API and be able to pass through custom tokenizer options that are overwriting `tokenizer.json`

### Your contribution

If the idea sounds good I can contribute the patch

### LINKED ISSUE #1642: Auto device on Linux x64 fails hard when CUDA shared library is unavailable instead of falling back

### LINKED ISSUE #1643: get_available_devices()

### LINKED ISSUE #1666: Gemma 4 generation passes num_logits_to_keep=0 in decoder_forward, causing full-prompt logits memory blowup

Suh0161:
Closes #425, closes #633, closes #1643, closes #1666, closes #1642, closes #1096.

Supersedes #1668 and #1669 (same changes, single PR).

### Summary

- **#425 / #633** — `return_offsets_mapping` on the tokenizer; token-classification outputs can include `start` / `end` character spans; GPT-2 ByteLevel `offset_mapping` includes the leading space where appropriate.
- **#1643 / #1666** — Adds `get_available_devices()`; fixes default `num_logits_to_keep` in `decoder_forward` so it matches generation behavior and avoids computing full-prompt logits (memory / OOM risk on large vocab + long prompts).
- **#1642 / #1096** — When `device: 'auto'`, tries execution providers sequentially instead of crashing if a higher-priority provider (e.g. CUDA) fails to load; `tokenizer_options` in `Text2TextGenerationPipeline` is forwarded to the tokenizer and stripped before `generate()`.

### Fix: Auto device falls back gracefully when a provider fails to load (#1642)

When `device: 'auto'` is used on a machine where a high-priority provider (e.g. CUDA on Linux without `libcuda.so`) is unavailable, the session previously crashed hard. `constructSessions` in `session.js` now tries each execution provider individually, logs a warning for each failure, and continues until one succeeds or all are exhausted. `createInferenceSession` is unchanged.

```js
await pipeline('text-generation', 'onnx-community/LFM2-1.2B-ONNX', { device: 'auto' });
```

### Feature: tokenizer_options passthrough in Text2TextGenerationPipeline (#1096)

```js
const output = await generator('Translate to French: Hello world', {
  max_new_tokens: 100,
  tokenizer_options: { max_length: 64, truncation: true },
});
```

### DIFF
--- a/packages/transformers/src/backends/onnx.js
+++ b/packages/transformers/src/backends/onnx.js
@@ -297,15 +297,6 @@ async function ensureWasmLoaded() {
 
 /**
  * Create an ONNX inference session.
- *
- * When `device: 'auto'` is used, multiple execution providers may be listed in
- * priority order (e.g. `['cuda', 'webgpu', 'cpu']`). Some providers (like CUDA)
- * can fail to load even when the hardware is present because the required shared
- * libraries are missing. In that case ONNX Runtime throws instead of falling back
- * to the next provider. This function retries session creation, dropping the
- * first provider on each failure, so the user always gets the best available
- * device rather than a hard crash.
- *
  * @param {Uint8Array|string} buffer_or_path The ONNX model buffer or path.
  * @param {import('onnxruntime-common').InferenceSession.SessionOptions} session_options ONNX inference session options.
  * @param {Object} session_config ONNX inference session configuration.
@@ -314,37 +305,15 @@ async function ensureWasmLoaded() {
 export async function createInferenceSession(buffer_or_path, session_options, session_config) {
     await ensureWasmLoaded();
     const logSeverityLevel = getOnnxLogSeverityLevel(env.logLevel ?? LogLevel.WARNING);
-
-    const providers = session_options.executionProviders
-        ? [...session_options.executionProviders]
-        : undefined;
-
-    let lastError;
-    // Try each provider prefix in turn. On failure, drop the first provider and retry.
-    // This handles cases like CUDA shared libraries being absent on a Linux x64 machine.
-    const attempts = providers ? providers.length : 1;
-    for (let i = 0; i < attempts; ++i) {
-        const opts = {
+    const load = () =>
+        InferenceSession.create(buffer_or_path, {
+            // Set default log severity level, but allow overriding through session options
             logSeverityLevel,
             ...session_options,
-            ...(providers ? { executionProviders: providers.slice(i) } : {}),
-        };
-        if (i > 0) {
-            logger.warn(
-                `Failed to create session with "${providers[i - 1]}" provider: ${lastError?.message ?? lastError}. ` +
-                `Retrying with: [${providers.slice(i).join(', ')}].`,
-            );
-        }
-        try {
-            const load = () => InferenceSession.create(buffer_or_path, opts);
-            const session = await (apis.IS_WEB_ENV ? (webInitChain = webInitChain.then(load)) : load());
-            session.config = session_config;
-            return session;
-        } catch (err) {
-            lastError = err;
-        }
-    }
-    throw lastError;
+        });
+    const session = await (apis.IS_WEB_ENV ? (webInitChain = webInitChain.then(load)) : load());
+    session.config = session_config;
+    return session;
 }
 
 /**
--- a/packages/transformers/src/backends/onnx.js
+++ b/packages/transformers/src/backends/onnx.js
@@ -154,6 +154,25 @@ if (ORT_SYMBOL in globalThis) {
 const InferenceSession = ONNX.InferenceSession;
 
 /**
+ * Returns the list of devices available in the current environment, sorted by priority/performance.
+ *
+ * **Example:** Check available devices before loading a model.
+ * ```javascript
+ * import { get_available_devices } from '@huggingface/transformers';
+ *
+ * const devices = get_available_devices();
+ * // Node.js (Windows):  ['dml', 'webgpu', 'cpu']
+ * // Node.js (Linux x64): ['cuda', 'webgpu', 'cpu']
+ * // Browser (WebGPU):   ['webgpu', 'wasm']
+ * // Browser (no WebGPU): ['wasm']
+ * ```
+ * @returns {import("../utils/devices.js").DeviceType[]} The list of available devices.
+ */
+export function get_available_devices() {
+    return [...supportedDevices];
+}
+
+/**
  * Map a device to the execution providers to use for the given device.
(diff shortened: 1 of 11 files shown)


xenova:
Hi 👋 thanks for the PR. Could you maybe make separate PRs for each issue you are trying to fix? This makes it a lot easier to review.

Thanks

Suh0161:
> Hi 👋 thanks for the PR. Could you maybe make separate PRs for each issue you are trying to fix? This makes it a lot easier to review.
> 
> Thanks

Hi, thanks for the feedback! I've split everything into separate PRs:

- #1671 — Fix num_logits_to_keep default and add get_available_devices() (closes #1666, #1643)
- #1672 — Fix auto device falling back when a provider fails to load (closes #1642)
- #1673 — Allow callers to pass tokenizer_options in Text2TextGenerationPipeline (closes #1096)

The offsets PR is still open separately as well. Closing this one.

PR #1674, opened 2026-05-04
Fix auto device falling back when a provider fails to load

### LINKED ISSUE #1642: Auto device on Linux x64 fails hard when CUDA shared library is unavailable instead of falling back

Suh0161:
Closes #1642

### Fix: auto device falls back gracefully when a provider fails to load

When `device: 'auto'` is used and a high-priority execution provider (e.g. CUDA on Linux without `libcuda.so` installed) is unavailable, `createInferenceSession` would throw instead of trying the next provider in the list.

The fallback now lives in `constructSessions` in `session.js`, where the device selection context is available. Each provider is tried individually — on failure a warning is logged and the next one is attempted. `createInferenceSession` itself is unchanged.

```js
// Before: hard crash on Linux x64 without libcuda.so
// After: warns and automatically falls back to the next available provider
await pipeline('text-generation', 'onnx-community/LFM2-1.2B-ONNX', { device: 'auto' });

### DIFF
--- a/packages/transformers/src/models/session.js
+++ b/packages/transformers/src/models/session.js
@@ -7,7 +7,7 @@ import {
 } from '../backends/onnx.js';
 import { getCacheNames } from '../configs.js';
 import { DATA_TYPES, DEFAULT_DTYPE_SUFFIX_MAPPING, isWebGpuFp16Supported, selectDtype } from '../utils/dtypes.js';
-import { selectDevice } from '../utils/devices.js';
+import { selectDevice, DEVICE_TYPES } from '../utils/devices.js';
 import { apis } from '../env.js';
 import { getCoreModelFile, getModelDataFiles } from '../utils/model-loader.js';
 import { Tensor } from '../utils/tensor.js';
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/transformers/src/models/session.js [hunk 1]
```diff
@@ -154,6 +154,36 @@ export async function constructSessions(pretrained_model_name_or_path, names, op
                     cache_config,
                     name,
                 );
+
+                // When device='auto', try each execution provider individually so that a failing
+                // accelerator (e.g. CUDA not installed) falls back cleanly to the next one.
+                // For any explicit device the caller already knows what they want — no fallback.
+                const isAuto =
```
nico-martin, line 161: The `isAuto` check re-reads the raw option/config value instead of using the resolved `selectedDevice` already computed in `getSession` and returned in `session_config.device`:

Please base this on the resolved device, for example `session_config.device === DEVICE_TYPES.auto`, or return `selectedDevice` explicitly from `getSession` and use that.

nico-martin (changes requested):
Hi @Suh0161

Thanks fo looking into this. I think it would be a great improvement for `device: "auto"`.

This matches the current `device: "auto"` behavior, because the candidates are already ordered by preference. But that order is static, and session creation only proves that a provider works, not that it is fastest for the user's hardware or model.

Should `auto` mean "first preferred provider that starts", or should it try to choose the best provider for the actual hardware/model? I would not block this fix on that, but the intended behavior should be documented.

This PR also overlaps conceptually with PR #1646, which proposes exporting `deviceToExecutionProviders` and `getSupportedDevices` so downstream callers can implement their own fallback logic. If both PRs move forward, it would be useful to align the semantics: `getSupportedDevices()` / `deviceToExecutionProviders("auto")` should describe the same priority order that this internal fallback uses, and documentation should make clear whether callers should rely on that order as a performance preference or only as a capability fallback order.

nico-martin:
closing this because there is now response from the author

PR #1675, opened 2026-05-04
Allow callers to pass tokenizer_options in Text2TextGenerationPipeline

### LINKED ISSUE #1096: Allow configuring the tokenizer in `Text2TextGenerationPipeline`

Suh0161:
Closes #1096

### Feature: `tokenizer_options` passthrough in `Text2TextGenerationPipeline`

The tokenizer call inside `_call()` previously used hardcoded options with no way to override them. Callers can now pass `tokenizer_options` alongside generation kwargs to control things like `max_length` and `truncation_side`.

`tokenizer_options` is extracted from `generate_kwargs` before `model.generate()` is called so it never leaks into `GenerationFunctionParameters`. Pipeline defaults (`padding: true`, `truncation: true`) still apply and caller values take precedence.

```js
const generator = await pipeline('text2text-generation', 'Xenova/LaMini-Flan-T5-783M');
const output = await generator('Translate to French: Hello world', {
  max_new_tokens: 100,
  tokenizer_options: { max_length: 64, truncation: true },
});

### DIFF
--- a/packages/transformers/src/pipelines/text2text-generation.js
+++ b/packages/transformers/src/pipelines/text2text-generation.js
@@ -72,16 +72,23 @@ export class Text2TextGenerationPipeline
         }
 
         const tokenizer = this.tokenizer;
+
+        // Callers may pass tokenizer_options as a top-level key inside generate_kwargs to
+        // control tokenization (e.g. max_length, truncation side) without touching generation
+        // parameters.  We extract it here so it never leaks into model.generate().
+        const { tokenizer_options: caller_tokenizer_options, ...rest_generate_kwargs } = generate_kwargs;
         const tokenizer_options = {
             padding: true,
             truncation: true,
+            ...caller_tokenizer_options,
         };
+
         let inputs;
         if (this.task === 'translation' && '_build_translation_inputs' in tokenizer) {
             // TODO: move to Translation pipeline?
             // Currently put here to avoid code duplication
             // @ts-ignore
-            inputs = tokenizer._build_translation_inputs(texts, tokenizer_options, generate_kwargs);
+            inputs = tokenizer._build_translation_inputs(texts, tokenizer_options, rest_generate_kwargs);
         } else {
             inputs = tokenizer(texts, tokenizer_options);
         }
@@ -89,7 +96,7 @@ export class Text2TextGenerationPipeline
         const outputTokenIds = await this.model.generate({
             ...inputs,
             ...this._default_generation_config,
-            ...generate_kwargs,
+            ...rest_generate_kwargs,
         });
         return tokenizer
             .batch_decode(/** @type {Tensor} */ (outputTokenIds), {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/transformers/src/pipelines/text2text-generation.js [hunk 1]
```diff
@@ -72,24 +72,31 @@ export class Text2TextGenerationPipeline
         }
 
         const tokenizer = this.tokenizer;
+
+        // Callers may pass tokenizer_options as a top-level key inside generate_kwargs to
+        // control tokenization (e.g. max_length, truncation side) without touching generation
```
nico-martin, line 77: The comment says callers can control `truncation side`, and the PR description also mentions `truncation_side`. That key would be passed inside `tokenizer_options`, but the tokenizer implementation does not appear to use it. `PreTrainedTokenizer._call()` only reads `truncation` and `max_length`, and `truncateHelper(...)` always truncates from the right by shortening the array.

So `tokenizer_options: { truncation_side: 'left' }` would be silently ignored.

packages/transformers/src/pipelines/text2text-generation.js
nico-martin, line 1: The public callback type still describes `options` only as generation parameters:

```js
@param {Partial<import('../generation/parameters.js').GenerationFunctionParameters>} [options]
```

If `tokenizer_options` remains part of the pipeline API, it should get its own typedef, similar to how `TextGenerationPipeline` defines text-generation-specific params.

packages/transformers/tests/pipelines/test_pipelines_text2text_generation.js
```diff
@@ -31,6 +31,24 @@ export default () => {
...
+
+      it(
+        "tokenizer_options are forwarded to the tokenizer",
+        async () => {
+          const text = "This is a test.";
+          // max_length=3 forces aggressive truncation via tokenizer_options;
+          // the model should still run without error and return a string result.
+          // tokenizer_options is extracted before generate() is called so it
+          // never leaks into GenerationFunctionParameters.
+          const output = await pipe(text, {
```
nico-martin, line 43: This test would still pass if `tokenizer_options` were ignored, because it only checks that generation returns an array with a string. To protect this feature, please assert a visible effect of the tokenizer option, or spy/mock the tokenizer call and verify it receives `{ truncation: true, max_length: 3 }` while `model.generate()` does not receive `tokenizer_options`.

nico-martin (changes requested):
HI @Suh0161, thanks for looking into this!

I actually think the change is useful. But there is also an API-design question: if we want to stay close to Python Transformers, explicit preprocess options like `truncation` are a better fit than a broad `tokenizer_options` bag.

@xenova, what do you think?

nico-martin:
closing this because there is now response from the author

PR #1676, opened 2026-05-04
Fix transformers.js crashing in Bun --compile single binaries

### LINKED ISSUE #1672: transformers.js can't be used in a Bun `--compile` single binary without patches
## Summary

`@huggingface/transformers@4.x` can't be used inside a [Bun](https://bun.sh) `--compile` single binary on Linux (and arguably on any platform) without patching the package, because of three independent issues:

1. **`onnxruntime-node` is statically imported** at `src/backends/onnx.js:23` and `dist/transformers.node.mjs:7545`. Bun's `--compile` extracts the native `.node` addon at runtime, but it doesn't bundle the accompanying `libonnxruntime.so.1` / `libonnxruntime.1.24.3.dylib` shared library that the addon `dlopen`s. The binary crashes on first import:

   ```
   Error [ERR_DLOPEN_FAILED]: libonnxruntime.so.1: cannot open shared object file: No such file or directory
   ```

2. **`sharp` is statically imported** at `src/utils/image.js:1` / `dist/transformers.node.mjs:17733`. Same root cause — sharp's native binding can't be bundled into a single binary, and even text-only pipelines (e.g. `feature-extraction`) crash at module load:

   ```
   Error: Could not load the "sharp" module using the darwin-arm64 runtime
   ```

3. **Even after routing to onnxruntime-web, `getCoreModelFile`/`getModelDataFiles` return file *paths* instead of buffers when `apis.IS_NODE_ENV`** (`dist/transformers.node.mjs:22330,22338`). The bundled `ort.webgpu.bundle.min.mjs` has its `node:fs` read path eliminated by tree-shaking (`if (false)` dead-code path) and tries to `fetch()` the bare path string, failing with `TypeError [ERR_INVALID_URL]: fetch() URL is invalid`.

## Repro

Minimal reproduction in any Bun project:

```ts
// src/cli.ts
const extractor = await pipeline("feature-extraction", "Xenova/all-MiniLM-L6-v2");
console.log((await extractor("hello world", { pooling: "mean", normalize: true })).data.length);
```

```bash
bun build --compile --minify ./src/cli.ts --outfile dist/app
./dist/app
# ERR_DLOPEN_FAILED on Linux, or sharp error on macOS
```

Real-world project hitting this: [evantahler/mcpx](https://github.com/evantahler/mcpx). The patches we ended up shipping are at [`patches/@huggingface%2Ftransformers@4.2.0.patch`](https://github.com/evantahler/mcpx/blob/main/patches/%40huggingface%252Ftransformers%404.2.0.patch).

## What we patched (summary)

| Change | Reason |
|--------|--------|
| Replace `import * as ONNX_NODE from "onnxruntime-node"` with `var ONNX_NODE = void 0` | Skip native binding load |
| In `IS_NODE_ENV` branch: set `ONNX = ort_webgpu_bundle_min_exports`, `supportedDevices = ["wasm"]`, `defaultDevices = ["wasm"]` | Use the bundled web ort instead of native |
| Replace `import sharp from "sharp"` with a stub function that throws lazily | Skip native binding load; image processing fails only if actually invoked |
| In `getCoreModelFile`/`getModelDataFiles`: pass `return_path = false` instead of `apis.IS_NODE_ENV` | Return buffer so the bundled ort-web doesn't try to `fetch()` a bare path |

## Suggested upstream fixes

1. **Make `onnxruntime-node` an optional / dynamic import** gated behind `apis.IS_NODE_ENV && !apis.IS_BUN_COMPILE_ENV` (or just behind a successful try). This is the single biggest unblocker — once the static import is gone, end users can override the backend via `globalThis[Symbol.for("onnxruntime")]` (which is already supported and documented) without patching.

2. **Make `sharp` a dynamic import** in `src/utils/image.js`, evaluated only when `loadImageFunction` is actually called. Text-only pipelines never need it.

3. **Ship a built target that doesn't statically import either native package.** Today the `exports` field has `node` → `transformers.node.mjs` and `default` → `transformers.web.js`. A third entry — `"bun-bundle": "./dist/transformers.bundle.mjs"` or a runtime feature-detect — would let Bun-compiled binaries opt into the WASM-only path without monkey-patching.

4. **Fix the `if (false)` dead-code in the bundled ort-web** so the `node:fs` read path stays alive. That would let users feed file paths through transformers in node-like environments without our `return_path = false` workaround. (This may be an upstream `onnxruntime-web` bundling issue rather than a transformers.js one.)

## Why the existing escape hatch isn't enough

The `globalThis[Symbol.for("onnxruntime")]` override (lines 11551-11553 of `dist/transformers.node.mjs`) is a great hook — but it runs *after* the top-level `import * as ONNX_NODE from "onnxruntime-node"` has already executed and crashed. The override never gets a chance to fire. Same for sharp.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Suh0161:
Closes #1672

Addresses all three root causes described in the issue.

### 1. `onnxruntime-node` static import crash

The static top-level `import * as ONNX_NODE from 'onnxruntime-node'` crashed with `ERR_DLOPEN_FAILED` before the `globalThis[Symbol.for('onnxruntime')]` override hook could fire. Now loaded via `await import()` wrapped in try/catch. On failure the library falls back to `onnxruntime-web` (WASM) automatically. Normal Node.js usage is completely unaffected.

### 2. `sharp` static import crash

Same issue as above. `import sharp from 'sharp'` crashed at module load time even for text-only pipelines. `sharp` is now loaded dynamically and only attempted when image processing is actually invoked. Text-only pipelines (`feature-extraction`, `text-generation`, etc.) no longer crash on import regardless of whether `sharp` is installed.

### 3. Model file paths instead of buffers

`getCoreModelFile` and `getModelDataFiles` returned file path strings (efficient for native `onnxruntime-node`) even when the WASM backend was active. The bundled `ort-web` has its `node:fs` read path eliminated by tree-shaking and tried to `fetch()` the bare path, failing with `ERR_INVALID_URL`. Both functions now return buffers when the WASM backend is active, using a `_nodeOnnxActive` flag exported from `onnx.js`.

### DIFF
--- a/packages/transformers/src/backends/onnx.js
+++ b/packages/transformers/src/backends/onnx.js
@@ -18,9 +18,22 @@
 
 import { env, apis, LogLevel } from '../env.js';
 
-// NOTE: Import order matters here. We need to import `onnxruntime-node` before `onnxruntime-web`.
-// In either case, we select the default export if it exists, otherwise we use the named export.
-import * as ONNX_NODE from 'onnxruntime-node';
+// onnxruntime-node is loaded dynamically so that environments where its native
+// bindings are unavailable (e.g. Bun --compile single binaries) do not crash
+// at module load time. When the load fails we fall back to onnxruntime-web
+// (WASM backend) so the library still works without patching.
+//
+// Note: we previously used a static import here, but that caused Bun --compile
+// binaries to crash before the globalThis[Symbol.for('onnxruntime')] override
+// hook could even fire. See: https://github.com/huggingface/transformers.js/issues/1672
+let ONNX_NODE = null;
+if (apis.IS_NODE_ENV) {
+    try {
+        ONNX_NODE = await import('onnxruntime-node');
+    } catch {
+        // native bindings not available; will fall back to the WASM backend below
+    }
+}
 import * as ONNX_WEB from 'onnxruntime-web/webgpu';
 import { loadWasmBinary, loadWasmFactory } from './utils/cacheWasm.js';
 import { isBlobURL, toAbsoluteURL } from '../utils/hub/utils.js';
@@ -102,11 +115,17 @@ let defaultDevices;
 let ONNX;
 const ORT_SYMBOL = Symbol.for('onnxruntime');
 
+// Tracks whether the native onnxruntime-node backend is active.
+// model-loader uses this to decide whether to pass file paths (efficient for
+// the native backend) or load buffers (required by the WASM backend).
+export let _nodeOnnxActive = false;
+
 if (ORT_SYMBOL in globalThis) {
     // If the JS runtime exposes their own ONNX runtime, use it
     ONNX = globalThis[ORT_SYMBOL];
-} else if (apis.IS_NODE_ENV) {
+} else if (apis.IS_NODE_ENV && ONNX_NODE) {
     ONNX = ONNX_NODE;
+    _nodeOnnxActive = true;
 
     // Updated as of ONNX Runtime 1.23.0-dev.20250612-70f14d7670
     // The following table lists the supported versions of ONNX Runtime Node.js binding provided with pre-built binaries.
--- a/packages/transformers/src/utils/image.js
+++ b/packages/transformers/src/utils/image.js
@@ -13,9 +13,19 @@ import { apis } from '../env.js';
 import { Tensor } from './tensor.js';
 import { saveBlob } from './io.js';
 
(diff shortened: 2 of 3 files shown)


xenova:
dynamic imports as they currently stand will break the build process we currently have. so, I unfortunately can't accept this PR.

Perhaps we should look upstream to why bun is crashing?

PR #1688, opened 2026-05-20
fix: bump onnxruntime-node dependency

omribz156:
Summary:
- bump `onnxruntime-node` from `1.24.3` to `^1.26.0`
- refresh `pnpm-lock.yaml` so it resolves `onnxruntime-node@1.26.0` and `global-agent@4.1.3`
- remove the deprecated `boolean@3.2.0` dependency chain from the lockfile

Verification:
- `npm view onnxruntime-node@1.24.3 dependencies version dist.unpackedSize --json`
- `npm view onnxruntime-node@1.26.0 dependencies version dist.unpackedSize --json`
- `corepack pnpm install --lockfile-only --ignore-scripts`
- `corepack pnpm install --lockfile-only --ignore-scripts --frozen-lockfile`
- `git diff --check -- packages/transformers/package.json pnpm-lock.yaml`
- confirmed `boolean@3.2.0`, `global-agent@3.0.0`, `onnxruntime-node@1.24.3`, and `onnxruntime-common@1.24.3` no longer appear in the package/lockfile

I did not run a full install/test pass because this is a dependency metadata-only change and `onnxruntime-node` is a large native package; the lockfile was updated without scripts or package downloads.

Implemented with Codex assistance, with the patch kept to the dependency and lockfile update.

### DIFF
--- a/packages/transformers/package.json
+++ b/packages/transformers/package.json
@@ -57,7 +57,7 @@
   "dependencies": {
     "@huggingface/jinja": "^0.5.6",
     "@huggingface/tokenizers": "^0.1.3",
-    "onnxruntime-node": "1.24.3",
+    "onnxruntime-node": "^1.26.0",
     "onnxruntime-web": "1.26.0-dev.20260416-b7804b056c",
     "sharp": "^0.34.5"
   },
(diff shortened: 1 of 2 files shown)


nico-martin:
A fix that includes this update is already on its way with the latest version of Transformers.js

PR #1696, opened 2026-05-29
Fix missing global exports

xenova:


Jofthomas:
Thanks for taking a look !

PR #1701, opened 2026-06-06
Added knip

nico-martin:
At React Norway I listened to a talk by @TkDodo about [knip](https://knip.dev/).

With this I added a knip command to npm that checks fo unused dependencies and unused code in Transformers.js. It already found a couple of lines. @xenova, do you think it makes sense to keep it in the project?

## Summary

- Add Knip to the monorepo with a workspace-aware configuration for root scripts and `packages/transformers`.
- Add a dedicated `knip` CI job in the unit test workflow using Node.js 24.10.0.
- Fix dependency issues surfaced by Knip by declaring direct imports and removing unused Jest environment dependency.
- Resolve unused and duplicate export findings by making internal helpers local, removing dead helpers/constants, and preserving compatibility exports without duplicate aliases.
- Fix a stale JSDoc import path in tokenizer tests.


package.json
```diff
@@ -23,6 +24,8 @@
   },
   "homepage": "https://github.com/huggingface/transformers.js#readme",
   "devDependencies": {
+    "@types/node": "^24.10.9",
+    "knip": "^6.15.0",
```
xenova:
```suggestion
    "@types/node": "24.10.9",
    "knip": "6.15.0",
```

imo we can pin exact versions for dev dependencies

packages/transformers/package.json
```diff
@@ -57,17 +57,18 @@
...
   },
   "devDependencies": {
+    "@jest/globals": "30.2.0",
     "@types/jest": "^30.0.0",
     "@types/node": "^24.1.0",
     "@webgpu/types": "^0.1.69",
     "esbuild": "^0.27.2",
     "jest": "^30.2.0",
-    "jest-environment-node": "^30.2.0",
     "jsdoc-to-markdown": "^9.1.3",
```
xenova, line 72: can you explain the jest-environment-node -> jest/globals package change?

packages/transformers/package.json
```diff
@@ -57,17 +57,18 @@
   "dependencies": {
     "@huggingface/jinja": "^0.5.6",
     "@huggingface/tokenizers": "^0.1.3",
+    "onnxruntime-common": "1.24.3",
```
xenova, line 60: I remember someone mentioned this before, but since onnxruntime-node and onnxruntime-web depend on it, I don't think it's necessary to define here.

but I suppose if there is a mismatch one day between e.g., `Tensor` from onnxruntime-node and onnxruntime-web, it may cause some issues.

Lmk what you think!

.github/workflows/tests.yml
```diff
@@ -18,6 +18,22 @@ env:
   ONNXRUNTIME_NODE_INSTALL: skip
 
 jobs:
+  knip:
```
xenova, line 21: while we're at it, probably a good idea to add a prettier check here which runs a test to enforce formatting has been done

packages/transformers/src/utils/core.js
xenova, line 1: nice... left-over code from the tokenizers migration

xenova (commented):
Very cool! It definitely caught some old/dead code and other code smells (unused exports, etc.)

I think I need to start using `knip` for a currently-unreleased library I'm working on 🤫 

My one minor gripe with the library is that it introduces quite a few dev dependencies, but since it's quite a well-used library across the ecosystem, I don't see this as a blocker. We also so happen to be removing many other dev dependencies in https://github.com/huggingface/transformers.js/pull/1665/, so we can consider it a fair swap haha.

PS: Big fan of @TkDodo ❤️ -- would love to see the talk if it's available online?

knip.json
xenova, line 1: also, if we can find a way to not require *another* config file, that would be great.

packages/transformers/package.json
nico-martin: This was one of the suggestions from knip since we don't use it.

knip.json
nico-martin: apparently it is possible to add the knip.json to package.json :)

packages/transformers/package.json
nico-martin: I think we should keep this as a direct dependency because `@huggingface/transformers` imports `onnxruntime-common` directly. And in my opinion as soon as we have a direct import, we should have it as a direct dependency and not rely on being provided by sub-dependencies.

The version mismatch concern is valid, but declaring it directly makes the resolved version explicit instead of depending on hoisting. If ORT node/web diverge further, we should probably revisit the ORT versions together rather than rely on a transitive copy.

nico-martin:
@xenova, btw, the talk is online 😊:
https://www.youtube.com/live/Ge0DbLf-2R0?si=omYvDPwNaD9_k5DV&t=10975

packages/transformers/package.json
nico-martin: Oh and about the "@jest/globals": knip found that we import @jest/globals in two places in the project but we don't have it in the dependencies. It never caused an issue because I think its still somewhere in the dependency tree, but as I said in the "onnxruntime-common", whenever we import a package directly, we should also have it in our dependencies.

nico-martin:
Closed in favor of https://github.com/huggingface/transformers.js/pull/1763 and https://github.com/huggingface/transformers.js/pull/1764

PR #1706, opened 2026-06-10
fix(onnx): fall back when an execution provider's shared lib fails to load

### LINKED ISSUE #1642: Auto device on Linux x64 fails hard when CUDA shared library is unavailable instead of falling back

tsushanth:
Closes #1642.

## Summary

On Linux x64 the auto-fallback chain pushes `cuda` first. ONNX Runtime fails the whole session when **any** provider in `executionProviders` fails to register, so a box with no CUDA install throws

```
OrtSessionOptionsAppendExecutionProvider_Cuda: Failed to load shared library
```

and never reaches `webgpu` / `cpu` — `device: 'auto'` degenerates into a hard error instead of falling back. Same shape for any other optional provider whose shared library isn't on the host.

## Fix

`createInferenceSession` in `backends/onnx.js`:

- Tracks providers that have failed to register in a module-level `Set` so subsequent sessions skip them up front rather than re-probing on every call.
- On `InferenceSession.create` failure, parses the failing provider name out of the ORT error (`OrtSessionOptionsAppendExecutionProvider_<Name>`), drops it from the list, logs which provider got disabled, and retries with the remaining providers — but only when there's something left to fall back to.
- Leaves caller-supplied non-array `executionProviders` and unrelated failures untouched: the retry is gated on a regex match of the ORT-specific load-failure message **and** the presence of more than one provider, so non-provider-load errors (model corruption, etc.) still surface immediately.

## Notes

I didn't add a unit test — the existing test suite is integration-style (loads real models from HF Hub) and there's no module-level mock pattern for `onnxruntime-node`'s `InferenceSession.create`. The fix is auditable as a self-contained change in `onnx.js` and was structured to be a no-op in the happy path (no caller-supplied providers, single-provider chains, non-Linux platforms).

### DIFF
--- a/packages/transformers/src/backends/onnx.js
+++ b/packages/transformers/src/backends/onnx.js
@@ -277,6 +277,43 @@ async function ensureWasmLoaded() {
 }
 
 /**
+ * Tracks providers we've seen fail to initialize so subsequent sessions don't
+ * pay the same cost twice (e.g. on Linux x64 boxes without a CUDA runtime,
+ * `cuda` is on the auto-fallback chain but throws on first use).
+ *
+ * @type {Set<string>}
+ */
+const failedExecutionProviders = new Set();
+
+/**
+ * Extracts the provider name from an ONNX Runtime "failed to append provider"
+ * style error. Returns the lowercase provider name (matching the names used in
+ * `DEVICE_TO_EXECUTION_PROVIDER_MAPPING`) or `null` if the error doesn't look
+ * like a provider-load failure.
+ *
+ * @param {unknown} error
+ * @returns {string|null}
+ */
+function detectFailedExecutionProvider(error) {
+    // @ts-ignore — we only read .message if present.
+    const message = String(error?.message ?? error ?? '');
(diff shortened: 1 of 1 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #1713, opened 2026-06-30
Respect HF_ENDPOINT for env.remoteHost

slandau3:
## Background

Some enterprise deployments route Hugging Face traffic through an internal proxy or self-hosted Hub, for example to allow only approved models. Today, Transformers.js initializes `env.remoteHost` to `https://huggingface.co/`, so Node.js users targeting those deployments need to override `env.remoteHost` manually in code.

## Change

This changes the default to read `process.env.HF_ENDPOINT` when present, and fall back to `https://huggingface.co/` otherwise. This follows the existing environment variable pattern used for Hub auth tokens while preserving the current default behavior.

## Tests

Adds a focused `env.test.js` covering both the fallback and `HF_ENDPOINT` cases.

### DIFF
--- a/packages/transformers/src/env.js
+++ b/packages/transformers/src/env.js
@@ -257,7 +257,10 @@ export const env = {
     },
     /////////////////// Model settings ///////////////////
     allowRemoteModels: true,
-    remoteHost: 'https://huggingface.co/',
+    remoteHost:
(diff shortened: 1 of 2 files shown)


nico-martin:
Close as superseded: incorporated into the broader #1717 implementation

PR #1714, opened 2026-07-01
fix: respect TypedArray byteOffset and byteLength in RawAudio.toBlob()

### LINKED ISSUE #1704: RawAudio.toBlob() encodes the entire underlying ArrayBuffer, ignoring TypedArray byteOffset/length
### System Info

  `@huggingface/transformers` 4.2.0 (web, also reproduces in Node). Regression from v3.x — 3.5.1 encodes the same input correctly.

  ### Description
  
  `encodeWAV` builds the output blob from each sample array's whole backing buffer:

  ```js
  return new Blob([buffer, ...chunks.map((chunk) => chunk.buffer)], { type: "audio/wav" });
  ```

  The WAV header's data-chunk size is computed from the **view's** length, but `chunk.buffer` contributes the **entire** underlying `ArrayBuffer`. When the audio
  is a `subarray()` view (byteOffset ≠ 0, or length < buffer length), header and payload disagree. Decoders honor the header, so they play the first N samples of
  the buffer — the wrong window of audio.
  
  ### Reproduction

  ```js
  import { RawAudio } from "@huggingface/transformers";

  const full = new Float32Array(48000);                                  // 2s @ 24kHz
  for (let i = 24000; i < 48000; i++) full[i] = Math.sin(i / 20) * 0.5;  // tone in 2nd half

  const view = full.subarray(24000);                                     // keep only the tone
  const blob = new RawAudio(view, 24000).toBlob();

  console.log(blob.size);
  // actual:   192044  (44-byte header + the whole 2s buffer)
  // expected:  96044  (44-byte header + the 1s view)
  ```

  Decoding the blob (e.g. `decodeAudioData`) plays the first second — silence — instead of the tone.
  
  ### Real-world impact

  In a streaming TTS pipeline that trims leading silence with `subarray()` before calling `toBlob()`, every start-trim silently becomes an end-truncation: the
  chunk keeps its leading silence and loses real speech off the end.

  ### Suggested fix

  Respect the view's window when building the blob:
  
  ```js
  return new Blob(
    [buffer, ...chunks.map((c) => new Uint8Array(c.buffer, c.byteOffset, c.byteLength))],
    { type: "audio/wav" }
  );
  ```
  
  (Passing the typed-array views directly also works — `Blob` parts honor `ArrayBufferView` offset/length.)

jahanzaib-iqbal-dev:
## Summary

Fixes `RawAudio.toBlob()` producing malformed WAV files when audio chunks are TypedArray subarrays with non-zero `byteOffset`.

## Root Cause

The `encodeWAV` method uses `chunk.buffer` to get the raw binary data from each audio chunk. However, for TypedArray subarrays (created via `.subarray()` or `.slice()` in streaming contexts), `.buffer` returns the **entire backing `ArrayBuffer`** regardless of the view's offset and length. This causes the WAV blob to contain extra bytes from outside the actual audio data, creating a length mismatch with the WAV header.

## Fix

Replace `chunk.buffer` with `new Uint8Array(c.buffer, c.byteOffset, c.byteLength)` to correctly extract only the relevant portion of the backing buffer, respecting the TypedArray view's offset and length.

**Before:**
\`\`\`js
return new Blob([buffer, ...chunks.map((chunk) => chunk.buffer)], { type: 'audio/wav' });
\`\`\`

**After:**
\`\`\`js
return new Blob([buffer, ...chunks.map((c) => new Uint8Array(c.buffer, c.byteOffset, c.byteLength))], { type: 'audio/wav' });
\`\`\`

## Impact

Fixes silent audio corruption in TTS pipelines that use streaming output, where audio chunks are subarrays of a larger buffer. The WAV header length fields are computed correctly from `chunk.length` (number of elements), but the old `chunk.buffer` could include padding bytes outside the view's range, causing the actual data payload to mismatch the declared lengths.

Fixes #1704

### DIFF
--- a/packages/transformers/src/utils/audio.js
+++ b/packages/transformers/src/utils/audio.js
@@ -813,7 +813,7 @@ function encodeWAV(chunks, rate) {
     /* data chunk length */
     view.setUint32(40, totalLength * 4, true);
 
-    return new Blob([buffer, ...chunks.map((chunk) => /** @type {ArrayBuffer} */ (chunk.buffer))], {
+    return new Blob([buffer, ...chunks.map((c) => new Uint8Array(c.buffer, c.byteOffset, c.byteLength))], {
         type: 'audio/wav',
     });
 }


nico-martin:
Close as duplicate of #1712

PR #1715, opened 2026-07-03
feat(hub): resume interrupted model downloads via HTTP Range (Node)

### LINKED ISSUE #1220: Support resuming model downloads from where they were interrupted.
### Feature request

Use OPFS to replace caches, Support resuming model downloads from where they were interrupted.

### Motivation

Since models are typically large in size, they are more prone to interruptions. Having to re-download the entire model each time can easily lead to wasted time and bandwidth. 

To address this, it's important to implement a mechanism that supports **resuming downloads from where they were interrupted**, allowing users to continue downloading without starting over. This approach not only improves efficiency but also reduces unnecessary resource consumption.

### Your contribution

I can provide a PR (Pull Request) to contribute this functionality

renezander030:
Follow-up to the discussion in #1220 (thanks @emojiiii for handing it over, and @sroussey for the Xet/CDC pointer).

This is **step 1: HTTP Range-resume for the Node file-cache path**, which is where large model weights are streamed to disk. It ships resume for how downloads work today, with no new transport dependency. Xet-native chunk caching (via the huggingface.js download path) is the natural follow-up and is intentionally out of scope here.

### What changes

- **`FileCache`** streams large downloads to a deterministic `<key>.incomplete` file plus a small sidecar (`{ etag, total }`), and **keeps the partial on failure** instead of deleting it.
  - `getResumeInfo(request)` reports `{ size, etag, total }` when a consistent partial exists.
  - `put()` branches on the response status: **206** appends to the partial (offset/total parsed from `Content-Range`); **200** restarts cleanly (truncates any stale partial). A truncated body (`loaded < Content-Length`) is rejected and the partial retained for the next attempt.
  - A per-key **lock** preserves the previous concurrent-writer guarantee: if the partial is already held by another writer, `put()` falls back to the original unique-temp path (correct, non-resumable, never corrupts the shared partial). Stale locks are stolen after a TTL.
- **`hub.js`** only sends `Range`/`If-Range` on the streaming path (`IS_NODE_ENV && return_path`), since a 206 on the buffered path would yield only the trailing bytes. It now accepts `206` alongside `200` and caches both. `If-Range: <etag>` makes the server fall back to a full 200 if the file changed upstream.
- **`CacheInterface`** gains an optional `getResumeInfo` so custom caches can opt in.

### Scope / notes

- Node-only by design. The browser Cache API cannot store partials; OPFS/Xet is the follow-up (step 2).
- Incremental sha256 integrity verification is deliberately left for a follow-up so this PR stays focused on the resume mechanics.

### Tests

New `tests/utils/file_cache.test.js` covers: no-partial, full 200 (no leftovers), truncated body (partial + sidecar retained, `getResumeInfo` correct), 206 append to completion, and 200 restart over a stale partial. Existing `cache`/`custom_cache` suites still pass; typecheck, prettier, and build are green.

Closes #1220

Opening as a draft to get maintainer direction on the Xet question in #1220 before polishing. Happy to adjust.

### DIFF
--- a/packages/transformers/src/utils/cache/FileCache.js
+++ b/packages/transformers/src/utils/cache/FileCache.js
@@ -28,20 +28,29 @@ const LOCK_ABANDONED_MS = 60 * 60 * 1000;
  * Destroy a write stream and wait until its file descriptor is actually
  * released. Used on the error paths, where the stream is still open over a
  * partial or temp file that is about to be unlinked or handed to another
- * writer. A no-op when the stream is absent or already closed.
+ * writer.
+ *
+ * The early-out tests `closed`, not `destroyed`: `destroyed` is set
+ * synchronously when teardown *begins*, so a stream already destroyed by an
+ * auto-destroy on error can still be holding its descriptor. `closed` flips
+ * with the `close` event, which is emitted once the descriptor is gone.
  *
  * @param {import('node:fs').WriteStream | undefined} stream
  * @returns {Promise<void>}
  */
 function destroyStream(stream) {
-    if (!stream || stream.destroyed) {
+    if (!stream || stream.closed) {
         return Promise.resolve();
     }
     return new Promise((resolve) => {
-        // `close` fires once the fd is gone, whether the teardown itself
-        // succeeded or not — either way there is nothing further to wait for.
+        // `close` fires whether the teardown itself succeeded or not — either
+        // way there is nothing further to wait for.
         stream.once('close', resolve);
-        stream.destroy();
+        // An already-destroyed stream is mid-teardown; destroying it again is
+        // pointless, but its pending `close` is still worth waiting for.
+        if (!stream.destroyed) {
+            stream.destroy();
+        }
     });
 }
 
@@ -586,6 +595,12 @@ export class FileCache {
      * request resumes onto — reviving bytes the caller asked to delete — and a
      * lock sitting in the way of the next writer.
      *
+     * An in-flight download by another writer is left alone: its lock, partial
+     * and sidecar are that writer's working set, and removing them would let a
+     * third writer take the same key and write concurrently. The completed
+     * file is still removed in that case — it is the entry the caller asked to
+     * delete, and it is not what the other writer is holding.
+     *
      * @param {string} request
      * @returns {Promise<boolean>} A Promise that resolves to `true` if a cache
      * entry (the completed file or a partial download) was deleted, `false`
@@ -595,10 +610,12 @@ export class FileCache {
     async delete(request) {
         const filePath = path.join(this.path, request);
         const incompletePath = filePath + '.incomplete';
+        const lockPath = incompletePath + '.lock';
 
         // Drop our own reservation first, so the lock is removed by the writer
         // that owns it and its heartbeat stops — rather than unlinking a path a
-        // live timer would keep touching.
+        // live timer would keep touching. This also means the check below only
+        // ever considers a lock belonging to somebody else.
         await this.releaseResume(request);
 
         const unlink = (target) =>
@@ -607,11 +624,18 @@ export class FileCache {
                 () => false,
             );
 
+        // Absent, stale-with-a-dead-owner, or stamped by this process: all
+        // report abandoned, and are ours to clear. Anything else is a writer
+        // actively downloading right now.
+        if (!(await this.#isLockAbandoned(lockPath))) {
+            return unlink(filePath);
+        }
+
         const [fileDeleted, partialDeleted] = await Promise.all([
             unlink(filePath),
             unlink(incompletePath),
             unlink(incompletePath + '.json'),
-            unlink(incompletePath + '.lock'),
+            unlink(lockPath),
         ]);
 
         return fileDeleted || partialDeleted;
--- a/packages/transformers/src/utils/cache/FileCache.js
+++ b/packages/transformers/src/utils/cache/FileCache.js
@@ -8,9 +8,19 @@ import { apis } from '../../env.js';
 // Create a dedicated random instance for generating unique temporary file names
 const rng = new Random();
 
+// How long a `.incomplete.lock` may sit before another writer treats it as
+// stale (a previous run crashed without releasing it). Well above any realistic
+// single-chunk write, but low enough that a resume is not blocked forever.
+const LOCK_STALE_MS = 10 * 60 * 1000;
+
 /**
  * File system cache implementation that implements the CacheInterface.
  * Provides `match` and `put` methods compatible with the Web Cache API.
+ *
+ * Large downloads are streamed to a deterministic `<key>.incomplete` file with
+ * a small sidecar recording the server `ETag` and expected size. If a download
+ * is interrupted the partial is kept, so a later attempt can send a `Range`
+ * request (see `getResumeInfo`) and continue instead of starting from byte 0.
  */
 export class FileCache {
     /**
@@ -38,6 +48,36 @@ export class FileCache {
     }
 
     /**
+     * Report whether a resumable partial download exists for `request`.
+     * Returns `{ size, etag, total }` when a `.incomplete` file and its sidecar
+     * are present and internally consistent, otherwise `undefined`. Callers use
+     * this to send a `Range`/`If-Range` request and continue the download.
+     * @param {string} request
+     * @returns {Promise<{size: number, etag: string|null, total: number} | undefined>}
+     */
+    async getResumeInfo(request) {
+        const incompletePath = path.join(this.path, request) + '.incomplete';
+        const sidecarPath = incompletePath + '.json';
+        try {
+            const [stat, sidecarRaw] = await Promise.all([
+                fs.promises.stat(incompletePath),
+                fs.promises.readFile(sidecarPath, 'utf-8'),
+            ]);
+            const sidecar = JSON.parse(sidecarRaw);
+            const total = typeof sidecar.total === 'number' ? sidecar.total : 0;
+            // Only resume when there is something to resume and we have not
+            // somehow written past the expected size. Anything inconsistent
+            // falls through to a clean restart.
+            if (stat.size > 0 && (total === 0 || stat.size < total)) {
+                return { size: stat.size, etag: sidecar.etag ?? null, total };
+            }
+        } catch {
+            // No partial, no sidecar, or unreadable — nothing to resume.
+        }
+        return undefined;
+    }
+
+    /**
      * Adds the given response to the cache.
      * @param {string} request
      * @param {Response} response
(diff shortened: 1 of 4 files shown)


packages/transformers/src/utils/cache/FileCache.js
```diff
@@ -47,20 +87,104 @@ export class FileCache {
...
+        const lockPath = incompletePath + '.lock';
+
+        // Claim the deterministic partial for this key. If another writer (this
+        // or another process) already holds it, fall back to the legacy
+        // random-suffix temp path: correct and non-resumable, but it never
+        // corrupts the shared partial — preserving the previous concurrency
+        // guarantee for parallel loads of the same file.
+        const locked = await this.#acquireLock(lockPath);
+        if (!locked) {
+            return this.#putUniqueTemp(filePath, response, progress_callback);
```
nico-martin, line 222: `response` may be a suffix-only 206. Passing it to the unique-temp full-download path can publish only those trailing bytes as the complete model. Please coordinate before issuing Range, wait for the active writer, or discard/refetch as 200.

packages/transformers/src/utils/cache/FileCache.js
```diff
@@ -47,20 +87,104 @@ export class FileCache {
...
+        }
+
+        // A 206 (Partial Content) means the server honored our Range request and
+        // we append to the existing partial. Anything else (200) is a fresh
+        // download: discard any stale partial and start over.
+        const resuming = response.status === 206;
```
nico-martin: Please validate the full `Content-Range`: start must equal the partial size, end must agree with body length, and total must remain consistent. The current total-only check can promote overlapping or gapped data.

packages/transformers/src/utils/cache/FileCache.js
```diff
@@ -90,19 +208,64 @@ export class FileCache {
...
+    }
+
+    /**
+     * Acquire the per-key write lock via an exclusive create (`wx`). If a lock
+     * already exists but is older than `LOCK_STALE_MS`, it is treated as
+     * abandoned by a crashed writer and stolen.
+     * @param {string} lockPath
+     * @returns {Promise<boolean>}
+     */
+    async #acquireLock(lockPath) {
```
nico-martin: If another process requests the same file after ten minutes, it may treat the lock from a still-active slow download as stale and start writing concurrently. Is concurrent use of this cache supported? If so, could we refresh the lock while downloading or verify that its owner is no longer alive before removing it?

packages/transformers/src/utils/hub.js
```diff
@@ -330,10 +336,28 @@ export async function loadResourceFile(
...
+            // partial for this key. This only applies to the streaming path
+            // (`IS_NODE_ENV && return_path`), where the body is written straight to
+            // disk; on the buffered path a 206 would yield only the trailing bytes.
+            // `Range` asks the CDN to continue from the last byte we have; `If-Range`
+            // makes the server fall back to a full 200 if the file changed upstream.
+            let resumeHeaders;
+            if (apis.IS_NODE_ENV && return_path && cache && typeof cache.getResumeInfo === 'function') {
+                const resume = await cache.getResumeInfo(proposedCacheKey);
+                if (resume && resume.size > 0) {
+                    resumeHeaders = { Range: `bytes=${resume.size}-` };
```
nico-martin: Please do not resume without a validator. A Range request without `If-Range` can combine two remote versions; restart with a full request when no safe validator exists.

nico-martin (changes requested):
Hi @renezander030, thank you so much for looking into this! Resumable downloads would be really useful here. Before this is safe to merge, we need to fix the 206 lock-contention path, avoid stealing locks from slow active downloads, and fully validate ranges and validators. Could you also add concurrency and malformed-range tests? I'm holding off on merging because the current edge cases can corrupt a cached model.

renezander030:
Thanks @nico-martin — all four are fair, and the 206 one is a genuine corruption bug rather than a sharp edge. You're right that a ranged response can reach the unique-temp path and get renamed over the final file.

To answer the concurrency question directly: **yes, concurrent use is supported.** The random-suffix temp path predates this PR and exists for exactly that. My fallback preserved it for 200 responses and quietly broke it for 206.

Rather than patch the fallback, I moved the coordination ahead of the request, which I think removes the class of bug instead of the instance:

- **Reserve before requesting.** `getResumeInfo` is now `reserveResume` — it takes the per-key lock and only then reports the partial. A `Range` is therefore only ever sent by the writer that owns the bytes it continues. A writer that loses the race gets `undefined` and does a plain full download, i.e. today's behaviour. `put()` refuses an unreserved 206 outright, and the fallback path refuses any non-200, so neither can publish a fragment as a whole file. `releaseResume` covers callers that abandon a download — including the case where another writer finished the file first, where `put()` is skipped entirely.

- **Lock ownership and liveness, not a blind TTL.** A held lock records its owner and is refreshed every 30s while the download runs, so an untouched lock means an *absent* writer rather than a slow one — that was the missing piece in the original TTL. A lock is only taken over when it is stale **and** its owner is provably gone (`process.kill(pid, 0)` → `ESRCH`). Since a pid on a shared cache directory can belong to an unrelated process on another host, an unverifiable owner is left alone until a one-hour ceiling, which prevents a permanent deadlock without ever interrupting a live slow download.

- **Full `Content-Range` validation.** Start must equal the partial size (no gap, no overlap), the span must agree with `Content-Length`, and the total must match the sidecar. `bytes */N` and suffix `bytes -N/M` are rejected outright — neither can be appended safely. On a mismatch the partial is discarded so the next attempt restarts clean rather than retrying onto data from a different revision forever.

- **No resume without a validator.** If there is no usable ETag to pin via `If-Range`, no `Range` is sent at all and the download restarts from byte 0.

Tests added as requested: concurrent writers (both finish, file complete exactly once, no temp/lock/sidecar left behind), a failing writer not stranding the key, unreserved and contended 206s, ten malformed `Content-Range` forms, lock stealing vs. not-stealing across four ownership scenarios, and the no-validator case. 29 tests in `file_cache.test.js`; the existing `cache`/`custom_cache`/`hub` suites, typecheck, prettier, and build are green.

One open item from the original description that I don't think got picked up: is the step-1 (`Range`) / step-2 (Xet-native) split the scope you want here, or would you rather this land closer to the huggingface.js download path? Happy to go either way — I'll take it out of draft now since the correctness gate is addressed.

renezander030:
@nico-martin one small thing when you get a chance — the workflow runs on this branch are sitting in `action_required`, so CI hasn't actually executed against the new commit. Could you approve them?

The full gate is green locally (jest including the new concurrency and malformed-range tests, `tsc --build`, prettier, and the esbuild bundles), but I'd rather have CI confirm that on your matrix than take my machine's word for it.

renezander030:
Thanks for approving the runs. Results on `76e1bc8c`: **Unit tests passed on Node 18, 20 and 22**; Build PR Documentation failed, and that one was on me — now fixed in `677362ac`.

The failure was in the workflow's `pre_command`, at `docs-api`. `jsdoc` cannot parse an object literal nested inside a generic, so the `CacheInterface` entry

```js
@property {(request: string) => Promise<{size: number, etag: string|null, total: number}|undefined>}
```

aborted the run with `Invalid type expression`. Extracted into a named `ResumeInfo` typedef, referenced from both the interface and `FileCache.reserveResume`.

Worth flagging that this was not introduced by the review fixes — it has been present since the first commit on this branch under the old `getResumeInfo` name. I confirmed it by running `docs-api` against `main` (passes), against the original commit `5101df2c` (fails with the same error), and against the fix (passes). It went unnoticed because the workflow runs on this branch were awaiting approval and had never actually executed. So the `Range`-resume work itself was never doc-clean; it just looked that way.

Local gate on the new commit: `docs-api`, the four cache/hub jest suites, `tsc --build`, prettier and the esbuild bundles are all green. Could you approve the runs once more so CI can confirm?

nico-martin:
@renezander030 Thanks, the redesign addresses the four blocking issues from my first review, and rest looks good too.

I found two cleanup issues remaining:

- FileCache.delete() should remove the associated .incomplete, sidecar, and lock artifacts, not only the completed file.
- Please ensure the output stream is destroyed/closed before releasing the reservation on reader, writer, or callback errors. While touching this path, completion should require loaded === total, not only reject loaded < total, so an oversized ranged body cannot be promoted.

Once those are covered with failure/deletion tests, I’m happy to approve.

renezander030:
@nico-martin Both covered in `8b3f681d`.

**`delete()` artifacts.** It now removes the `.incomplete`, its sidecar and the lock alongside the completed file. If this instance holds a reservation for the key it is released first, so the lock is dropped by its owner and its heartbeat stops, rather than being unlinked out from under a live timer.

One deliberate choice on the return value: it still reports `true` only for a real entry — the completed file or a partial. A leftover sidecar or lock is swept up either way but does not count. `clear_cache` chains on this (`delete(proposedCacheKey)` falling back to `delete(localPath)`), so letting a stray lock report `true` would silently skip the fallback key.

**Stream teardown.** Both write paths now destroy the stream and wait for the descriptor to be released *before* the reservation is dropped — reader errors, write errors and progress-callback errors alike. Previously the `finally` released the lock while the stream was still open on the partial, so the next writer could take the key with a descriptor still on the file. The fallback path had the same shape, where an open handle also makes the temp-file unlink fail outright on Windows.

While there I added an `error` listener to both streams. An async write failure surfaces through the `write`/`close` callbacks, but with no listener attached the same failure also reached the process as an unhandled `'error'` event.

**Completion gate.** Now `loaded === total` rather than `loaded < total`. A short body still keeps its partial so the next attempt can resume; an oversized one is discarded, since it already runs past `total` and can never be resumed onto. Neither is published.

**Tests** — 13 added, 42 total in `file_cache.test.js`:

- deletion: completed file, partial + sidecar + stale lock, both together, a held reservation, nothing cached, and a bare leftover lock (which must return `false`)
- stream teardown: reader error mid-body, progress callback throwing, writer error, and the fallback path's temp-file cleanup
- oversized bodies: a 200 overshooting `Content-Length` and a 206 overshooting a valid `Content-Range`

Nine of the thirteen fail against the previous commit; the other four pin down behaviour that was already correct and is easy to break from here. The teardown tests assert against `/proc/self/fd` where it exists — POSIX unlinks an open file happily, so a removability check alone would have passed either way and proved nothing.

Local gate green: `file_cache`/`cache`/`custom_cache`/`hub` suites, `tsc --build`, prettier, `docs-api`, and the esbuild bundles.


I resolved the four threads from your first review, so only what you flag from here stays open. CI on `8b3f681d` has gone to `action_required` again — could you approve the runs so the matrix confirms the above?

nico-martin (changes requested):
Thanks, the cleanup and exact-size checks look good. I found one remaining concurrency issue: delete() currently removes the lock even if another writer is actively using it. That could let a second writer acquire the same key and write concurrently. Could you preserve live foreign locks and only remove locks we own or have confirmed are stale?
Also, stream.destroyed doesn’t necessarily mean the file descriptor is closed yet. destroyStream() should still wait for close unless stream.closed is already true.

renezander030:
@nico-martin Both fixed in `51ffe494`. You were right on each, and the lock one was a real hole — I had reasoned about `delete()` as "remove the entry" and not as something that races a live writer, which is exactly the case the reservation scheme exists to handle.

**Live foreign locks.** `delete()` now runs the same abandonment check `reserveResume` uses, rather than unlinking the lock outright. Absent, stale with a provably dead owner, or stamped by this process → ours to clear. Anything else is a writer downloading right now.

One decision I'd like you to sanity-check, since it goes slightly past what you asked. When the lock is live I also leave the `.incomplete` and its sidecar. They are that writer's working set: pulling the partial out from under it would leave the lock pointing at nothing and make the writer's final rename fail with `ENOENT`, so preserving only the lock felt half-done. The completed file is still removed in that case — it is the entry the caller asked to delete, and it is not what the other writer is holding. If you'd rather `delete()` were strictly lock-scoped and cleared the partial regardless, say so and I'll narrow it.

**`destroyed` vs `closed`.** Correct, and my test was weaker than I thought. `destroyStream` now short-circuits on `stream.closed` and waits for the `close` event otherwise, re-destroying only a stream that has not already begun tearing down.

Worth being precise about the impact, because it changes how the test had to be written: against a real `fs.WriteStream` the gap is invisible. The descriptor is released within the next event-loop turn, and the `unlink` in `releaseResume` yields long enough for that to land — which is why my `/proc/self/fd` assertions passed against the previous commit either way and proved nothing here. The guarantee was missing rather than the behaviour, so the new test drives `put` through a stream whose teardown is slow enough to observe and asserts `put` does not return while the descriptor is still open. It fails against `8b3f681d`.

**Tests** — 4 added, 46 total:

- a live foreign lock surviving `delete` while the completed file goes, and the key staying unavailable to anyone else afterwards
- a stale lock cleared along with its partial and sidecar once its owner is provably gone
- a lock this instance stamped but no longer tracks — the "we own it" branch, where the pid is alive because it is us
- `put` not returning while a mid-teardown stream still holds its descriptor

Two of the four fail against `8b3f681d`; the other two pin down the branches of the new abandonment check that happen to coincide with the old unconditional behaviour.

Local gate green: `file_cache`/`cache`/`custom_cache`/`hub`/`model_registry` (98 passed, 1 skipped), `tsc --build`, prettier, `docs-api`, esbuild bundles. Could you approve the runs on `51ffe494`?

PR #1716, opened 2026-07-06
Allow all `origins` for Cross-Origin Storage

tomayac:
This makes the resources available globally. Initially this wasn't implemented in the extension, but the `origins` check is implemented now, so we need to update the code to make use of it. (Internally, the extension as a one-off upgrades all resources that were stored already as `origins: '*'`.)

### DIFF
--- a/packages/transformers/src/utils/cache/CrossOriginStorageCache.js
+++ b/packages/transformers/src/utils/cache/CrossOriginStorageCache.js
@@ -112,6 +112,7 @@ export class CrossOriginStorage {
     _storeBlobInCOS = async (blob, hashHex) => {
         const handle = await navigator.crossOriginStorage.requestFileHandle(makeHashDescriptor(hashHex), {
             create: true,
+            origins: '*',
         });
         const writableStream = await handle.createWritable();
         await writableStream.write(blob);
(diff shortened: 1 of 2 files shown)


tomayac:
By the way, we also now offer [TypeScript types](https://www.npmjs.com/package/@types/wicg-cross-origin-storage).

nico-martin (approved):

xenova:
Thanks for the PR!

build fails with:
```
Error: src/utils/cache/CrossOriginStorageCache.js(115,13): error TS2353: Object literal may only specify known properties, and 'origins' does not exist in type 'CrossOriginStorageRequestFileHandleOptions'.
 ELIFECYCLE  Command failed with exit code 1.
/home/runner/_work/transformers.js/transformers.js/packages/transformers:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @huggingface/transformers@4.2.0 build: `node scripts/build.mjs && pnpm typegen`
Exit status 1
 ELIFECYCLE  Command failed with exit code 1.
Error: Process completed with exit code 1.
```

seems that we should update types?

nico-martin (approved):

xenova (approved):
LGTM ✅  thanks!

Issue #380, opened 2023-11-04
[Bug] Local Models Mode Not Usable in Web Extensions


MentalGear:
It seems that Transformer.js uses the CacheAPI to cache models. However, the CacheAPI can't be used with urls like `chrome-extension:` or `moz-extension:` which results in `Transformer.js` failing with:
```shell
Failed to execute 'put' on 'Cache': 
Error: no available backend found. ERR: [wasm] TypeError: Failed to fetch
```
when trying to use Local Models in Web Extensions.

**Describe the bug**
When trying to use local models in the example transformer.js chrome extension, the following error occurs.

**How to reproduce**
1. fork https://github.com/xenova/transformers.js/blob/main/examples/extension/src/background.js
2. Put the model / wasm files in a local folder
3. Use local mode in transformer.js

**Expected behavior**
Using local models.

**Logs/screenshots**
```shell
index.js:6 Unable to determine content-length from response headers. Will expand buffer when needed.
A @ index.js:6
3index.js:6 Unable to add response to browser cache: TypeError: Failed to execute 'put' on 'Cache': Request scheme 'chrome-extension' is unsupported.
(anonymous) @ index.js:6
index.js:6 Unable to determine content-length from response headers. Will expand buffer when needed.
A @ index.js:6
w @ index.js:6
index.js:6 Unable to add response to browser cache: TypeError: Failed to execute 'put' on 'Cache': Request scheme 'chrome-extension' is unsupported.
(anonymous) @ index.js:6
Promise.catch (async)
w @ index.js:6
index.js:1704 Uncaught (in promise) Error: no available backend found. ERR: [wasm] TypeError: Failed to fetch
```

**Environment**
- Transformers.js version: 2.7.0
- Browser (if applicable): Chrome
- Operating system (if applicable): macOS

MentalGear:
Never mind, the problem was that the wasm model was not made available in the manifest's 'web_accessible_resources'.

MaxFrax:
@MentalGear @xenova Would either of you be able to share a working example of using local models within a Chrome extension?

I'm attempting to embed a model in the extension, but I'm running into issues with resolving the path. Any guidance would be greatly appreciated!

MentalGear:
Don't know anymore.On 10 Sep 2024, at 16:03, Massimo ***@***.***> wrote:﻿
@MentalGear @xenova Would either of you be able to share a working example of using local models within a Chrome extension?
[...]

PR #1723, opened 2026-07-15
Skip browser cache writes for non-http(s) resources

patrickkettner:
This replaces #614, the original report of this problem. That PR patched a file that has since been rewritten and moved, so it is being closed as obsolete in favor of this one. Related background: #380.

When a model is bundled inside a browser extension (`env.localModelPath` pointing at a `chrome-extension://` URL) with the browser cache enabled, every file load attempts `cache.put()` with a `chrome-extension://` key. The browser Cache API only supports http(s) request keys, so the put always fails: one call site logs `Unable to add response to browser cache: TypeError: ...` on every load, and the streaming path has no error handling at all.

This skips the write when the cache is the browser Cache API and the key does not resolve to an http(s) URL. Keys are resolved against the page location first, so a relative `env.localModelPath` on an extension page is caught too. The other cache backends (FileCache, custom caches, Cross-Origin Storage) are untouched.

Resolving can itself fail, since nothing resolves against an opaque base like `about:blank`. `toAbsoluteURL` takes an opt-in `allowUnresolved` flag for that, and only this call site passes it, so the pre-existing `wasmPaths` callers still throw on a bad value.

Tests cover the cache behavior on extension, opaque-origin and regular pages, the remote case, and the `toAbsoluteURL` flag itself.

### DIFF
--- a/packages/transformers/src/utils/hub.js
+++ b/packages/transformers/src/utils/hub.js
@@ -15,6 +15,7 @@ import {
     isValidHfModelId,
     makePretrainedOptionsKey,
     readResponse,
+    toAbsoluteURL,
 } from './hub/utils.js';
 import { getCache, tryCache } from './cache.js';
 import { get_file_metadata } from './model_registry/get_file_metadata.js';
@@ -204,12 +205,12 @@ export async function storeCachedResource(path_or_repo_id, filename, cache, cach
     if (
         typeof Cache !== 'undefined' &&
         cache instanceof Cache &&
-        isValidUrl(cacheKey) &&
-        !isValidUrl(cacheKey, ['http:', 'https:'])
+        !isValidUrl(toAbsoluteURL(cacheKey), ['http:', 'https:'])
     ) {
         // The browser Cache API only supports http(s) URLs as keys, so do not attempt to cache
-        // responses for other schemes (e.g., files bundled within a browser extension).
-        // Relative keys are not skipped, since the Cache API resolves them against the page URL.
+        // responses for other schemes (e.g., files bundled within a browser extension). Relative
+        // keys are resolved against the page URL first, since a relative key on an extension page
+        // still resolves to a chrome-extension:// request.
         return;
     }
 
--- a/packages/transformers/src/utils/hub.js
+++ b/packages/transformers/src/utils/hub.js
@@ -201,6 +201,18 @@ export async function storeCachedResource(path_or_repo_id, filename, cache, cach
         return;
     }
 
+    if (
+        typeof Cache !== 'undefined' &&
+        cache instanceof Cache &&
+        isValidUrl(cacheKey) &&
+        !isValidUrl(cacheKey, ['http:', 'https:'])
+    ) {
+        // The browser Cache API only supports http(s) URLs as keys, so do not attempt to cache
+        // responses for other schemes (e.g., files bundled within a browser extension).
+        // Relative keys are not skipped, since the Cache API resolves them against the page URL.
+        return;
+    }
+
     if (!result) {
         // We haven't yet read the response body, so we need to do so now.
         // Ensure progress updates include consistent metadata.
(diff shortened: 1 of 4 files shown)


packages/transformers/src/utils/hub.js
```diff
@@ -201,6 +201,18 @@ export async function storeCachedResource(path_or_repo_id, filename, cache, cach
         return;
     }
 
+    if (
+        typeof Cache !== 'undefined' &&
+        cache instanceof Cache &&
+        isValidUrl(cacheKey) &&
```
nico-martin: This checks only absolute keys. On an extension page, `/models/foo` resolves to `chrome-extension://<id>/models/foo`, so the real Cache API still rejects it. Please resolve the key against `globalThis.location?.href` (possibly via the existing URL helper), inspect the effective protocol, and run the relative-path test against an extension-scheme base.

nico-martin (changes requested):
Hi @patrickkettner, thank you so much for looking into this! The narrow Cache API guard is a good idea. One case is still missing: relative model paths on an extension page resolve to `chrome-extension:` URLs and can still fail. Could you validate the resolved URL and add a test using an extension URL as the page base? I'm holding off on merging until that common extension path is covered.

packages/transformers/src/utils/hub.js
patrickkettner: good catch, thanks! pushed a fix that resolves the key against the
page location before checking the scheme, so a relative path on an
extension page gets caught too. also found toAbsoluteURL could throw
on an opaque-origin page like about:blank, so that's guarded now with
a test.

nico-martin (approved):
Thanks, lgtm :)

packages/transformers/src/utils/hub/utils.js
```diff
@@ -185,5 +186,10 @@ export function toAbsoluteURL(url) {
         return url;
     }
 
-    return new URL(url, baseURL).href;
+    try {
+        return new URL(url, baseURL).href;
+    } catch {
+        // baseURL has an opaque origin (e.g. "about:blank"), so return the URL unchanged
+        return url;
+    }
```
xenova, line 200: the only other usage of `toAbsoluteURL` (pre-PR) was resolving `wasmPaths` from ONNX env, and since this is user-configurable, I think that a user could specify some strange value here, and encounter an unexpected error. It would fall to the `catch` branch, propagating the incorrect value elsewhere... when the expected `Uncaught TypeError: Failed to construct 'URL': Invalid URL` error should be thrown here.

xenova: perhaps we can be more explicit here... (e.g., detecting for `"about:blank"` and other valid options specific to chrome extensions as the baseURL... or add a flag which enables this branch only for certain cases, like used below (not in the original uses)

xenova (changes requested):
Thanks for the PR! One quick comment 🤗

packages/transformers/src/utils/hub/utils.js
patrickkettner: good call, thanks! went with the flag option.

packages/transformers/src/utils/hub/utils.js
```diff
@@ -185,5 +189,14 @@ export function toAbsoluteURL(url) {
...
+    if (!allowUnresolved) {
+        return new URL(url, baseURL).href;
+    }
+
+    try {
+        return new URL(url, baseURL).href;
+    } catch {
+        // Nothing resolves against an opaque base, so hand back the original for the caller to check
+        return url;
+    }
```
xenova: minor style suggestion
```suggestion
    try {
        return new URL(url, baseURL).href;
    } catch (error) {
        if (!allowUnresolved) {
            throw error;
        }
        // Nothing resolves against an opaque base, so hand back the original for the caller to check
        return url;
    }
```

xenova (approved):
thanks!

PR #1732, opened 2026-07-28
Pluggable inference providers

nico-martin:
This PR decouples Transformers.js from ONNX Runtime by moving all ORT-specific code into a new workspace package, `@huggingface/transformers-onnxruntime`, and introducing a pluggable **inference provider** architecture. It is the groundwork for alternative runtimes (e.g. custom WebGPU kernels): a custom backend can now supply model loading and execution while Transformers.js keeps ownership of tokenization, processors, configs, hub access, and the generation loop.

### New package: `@huggingface/transformers-onnxruntime`
- Owns the `onnxruntime-node` / `onnxruntime-web` dependencies and all ORT-specific code: EP/device selection, dtype → file-suffix mapping, external-data loading, WASM caching, session construction and execution, and the ONNX tensor-op micro-graphs.
- Written in strict TypeScript with typed host contracts.
- Connected to the main package through an inversion-of-control **host seam** (`configureOnnxProviderHost`): the main package injects `env`, `apis`, `logger`, hub file loading, tensor factories, caching, and limits. `env`/`apis`/`logger` bind late via Proxies, and a `Symbol.for('transformers.js.onnxProviderHost')` global allows host registration before module evaluation.
- Loaded **lazily**: apps using only custom backends no longer bundle or initialize ONNX Runtime.

### Inference backend contract (`src/backends/`)
- Duck-typed `InferenceBackend` (`{ modelId, load(options) }`) accepted by `pipeline()`, `AutoModel.from_pretrained()`, and `PreTrainedModel.from_pretrained()` in place of a model ID string.
- Namespaced capability model (`capabilities.causalGeneration`, `forward`, reserved families) with early task validation; the flat `generationCapabilities` field remains supported as a deprecated alias.
- `Tensor` now uses runtime-neutral storage; providers wrap native tensors via `Tensor.fromBackendStorage()`. `TensorOpRegistry` is populated by the active provider.
- Model-file discovery (`utils/model_registry/*`) goes through an injectable registry-provider adapter instead of referencing ONNX directly.

### Generation: policy/runtime split (`src/generation/`)
- New `GenerationController` owns the CPU-side policy (logits processors, sampling, stopping criteria, streaming, finalization); `PreTrainedModel.generate()` is rewritten around it.
- New versioned runtime protocol for custom backends:
  - `PullAutoregressiveSessionV1` — per-step CPU logits leases (`prefill`/`decode`/`read`/`release`).
  - `PlanAutoregressiveSessionV1` — GPU-resident decoding via `generateWithPlan` with declarative plans (`{ sampler: { op: 'argmax' } }`), so a WebGPU runtime can pipeline tokens without CPU round-trips.
  - Strict capability validation, optional `sessionConcurrency` limits, and guaranteed lease/session cleanup on error or abort.

### Cancellation support
- New `signal` (AbortSignal) option is honored end-to-end for built-in models: metadata requests, hub downloads, ONNX model-file loading, and session construction, with rollback of partially constructed sessions.

### Docs & tests
- New guide: `docs/source/guides/custom-backends.md` (backend contract + generation protocol), wired into the toctree.
- New test coverage for the module-wiring layer: fresh-module host wiring, provider-object loading (including a real tiny-model integration test), hub abort behavior, generation controller/protocol suites, and a compile-tested TypeScript sample of the public backend types.

## Breaking / behavioral changes
- `OnnxInferenceProvider` is imported from `@huggingface/transformers-onnxruntime` (no longer re-exported by the main package). String model IDs continue to work unchanged.
- `env.backends` starts empty and `env.backends.onnx` is populated on first ONNX use (lazy) instead of at import time.
- `subfolder` option default changed from `'onnx'` to `null`; the ONNX provider applies its own `'onnx'` default, so effective behavior is unchanged.
- `@huggingface/transformers` now depends on `@huggingface/transformers-onnxruntime` (published from this repo; added to the publish workflow).


sroussey:
progress events on prefill would be nice too. sometimes that is where most of the time is and there is no feedback

xenova:
> progress events on prefill would be nice too. sometimes that is where most of the time is and there is no feedback

Thanks for the suggestion. Are you aware of any runtimes which support this?

sroussey:
> > progress events on prefill would be nice too. sometimes that is where most of the time is and there is no feedback
> 
> Thanks for the suggestion. Are you aware of any runtimes which support this?

No. It’s a nice to have. Refill can take a while if you have a large doc and the model supports it.

packages/transformers-onnx/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "@huggingface/transformers-onnx",
```
xenova: maybe
```
@huggingface/transformers-onnxruntime
```
? (onnx=spec, onnxruntime=runtime)

PR #1733, opened 2026-07-31
Add llguidance-based constrained generation (`@huggingface/transformers-llguidance`)

nico-martin:
Adds grammar-constrained generation to Transformers.js via a [pure JavaScript implementation](https://github.com/nico-martin/llguidance-js) of [llguidance](https://github.com/guidance-ai/llguidance), so model output can be forced to match a JSON schema, a regex, or generic JSON guaranteed at the token level, not by prompting.

The integration is split into a minimal core hook and a separate opt-in package, so the core stays free of any llguidance dependency:

### 1. Core: post-sampling hook (`@huggingface/transformers`)

- New optional `LogitsProcessorList.onTokensSampled(token_ids, input_ids)` hook, dispatched after each generation step to every processor that implements it. This lets a logits processor observe which token was actually sampled. Required for grammar engines that advance their state per committed token.
- The generation loop in `modeling_utils.js` appends the sampled tokens for the full batch, then fires the hook with the post-append state. Behavior is unchanged for existing users. Processors without the hook are unaffected.
- `logger` is now exported from the main entry point.
- New core tests: hook ordering/batching and custom `StoppingCriteria` support.

### 2. New package: `@huggingface/transformers-llguidance`

`LlguidanceConstraint.fromResponseFormat(tokenizer, response_format)` accepts `json_object`, `json_schema`, or `regex` formats and returns `{ logits_processor, stopping_criteria, stats, dispose }` ready to pass to `generate()`:

- Each step, the llguidance interpreter computes a packed token bitmask (zero-allocation via `computeMaskInto` into a reused `Uint32Array`), and disallowed tokens are banned by writing `-Infinity` into the logits. The mask application is optimized for skewed grammar masks: fully-allowed 32-token words are skipped, fully-banned words use `fill`, and only mixed words pay per-bit cost.
- Sampled tokens are committed back to the interpreter through the new hook. The stopping criteria ends generation once the grammar accepts.
- On acceptance reported by the mask, logits are constrained to the tokenizer's EOS token(s) so no unconstrained token can leak into the output.
- Fails closed everywhere: unsatisfiable grammars (`dead_end`), requested backtracking, missing vocab/EOS metadata, and batch sizes other than 1 all throw clear errors instead of silently producing invalid output.
- The llguidance interpreter is disposed automatically on terminal states and error paths; `dispose()` (idempotent) covers early exits such as `max_new_tokens`.
- `stats` exposes per-run timings (`computeMaskMs`, `applyMaskMs`, `commitTokenMs`), step count, stop reason, and llguidance mask-cache instrumentation.

## Demo

```js
import { LlguidanceConstraint } from "@huggingface/transformers-llguidance";

const pipe = await pipeline("text-generation", "onnx-community/gemma-4-E2B-it-ONNX", {
  device: "webgpu",
  dtype: "q4f16",
});

const { logits_processor, stopping_criteria } =
  await LlguidanceConstraint.fromResponseFormat(pipe.tokenizer, {
    type: "json_schema",
    json_schema: {
      type: "object",
      properties: {
        answer: { type: "string" },
      },
      required: ["answer"],
      additionalProperties: false,
    },
  });

const output = await pipe(
  [
    {
      role: "system",
      content: "You are a helpful assistant.",
    },
    { role: "user", content: "Answer with a short friendly greeting." },
  ],
  {
    max_new_tokens: 64,
    logits_processor,
    stopping_criteria,
  },
);

console.log("=== ANSWER ===");
console.log(JSON.parse(output[0].generated_text.pop().content));

await pipe.dispose();
```


nico-martin:
I tested two thing.

A rather long generation:
```javascript
const messages = [
  {
    role: "system",
    content: "You are a helpful cooking assistant. Return practical, realistic recipes that match the requested format.",
  },
  {
    role: "user",
    content: "Create a weeknight vegetarian pasta recipe with pantry ingredients."
  }
];
```
Screenshot 2026-07-31 at 10 05 54

And also a really short one:
```javascript
const messages = [
  {
    role: "system",
    content: "You are a helpful AI assistant.",
  },
  {
    role: "user",
    content: "Answer with a short friendly greeting."
  }
];
```
image

Yes, the constrains make the geneeation slower, but I'd say it's within an expected and acceptable range

packages/transformers/src/models/modeling_utils.js
```diff
@@ -1003,13 +1003,19 @@ export class PreTrainedModel extends Callable {
...
                     break;
                 }
             }
+            for (let batch_idx = 0; batch_idx < generated_input_ids.length; ++batch_idx) {
+                all_input_ids[batch_idx].push(generated_input_ids[batch_idx][0]);
+            }
+            prepared_logits_processor.onTokensSampled(
+                generated_input_ids.map(([token_id]) => Number(token_id)),
+                all_input_ids,
+            );
```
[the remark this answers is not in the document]
nico-martin: resolved with https://github.com/huggingface/transformers.js/pull/1733/commits/6331dd4e621c4c3ce849d750595527c6bb2811df

nico-martin:
@xenova, I've completely rewritten the response-constraints.
It does not require any dependencies and it is much faster than before. I also ran benchmarks with different tasks and models (ONNX). And as it stands now the constraints add suuuper little overhead:

## gemma-4-E2B

### simple JSON schema
````
{"answer": "Hello! How can I assist you today?"}
````

### regex
````
Hi there!
````

### complex JSON schema
````
{"answer": {"text": "Hello! How can I assist you today?", "tone": "friendly", "language": "en"}, "alternatives": ["Hi there!", "Greetings!", "Hey!"], "metadata": {"confidence": 98, "safe": true, "tags": ["greeting", "welcome", "english"]}}
````

### unconstrained
````
Hello! How can I help you today? 😊
````

## LFM2.5-2.6B

### simple JSON schema
````
{"answer": "string (minLength 1, maxLength 120)}"}
````

### regex
````
Hello.
````

### complex JSON schema
````
{"answer": {"text": "Hello! How can I help you today?", "tone": "friendly", "language": "en"}, "alternatives": ["Hi there!", "Greetings!", "Hey!"], "metadata": {"confidence": 95, "safe": true, "tags": ["greeting", "friendly"]}}
````

### unconstrained
````
The user wants a short friendly greeting. I should respond with a simple, warm greeting.</think>Hello! How can I help you today? 😊
````

## granite-4.0-h-350m-ONNX

### simple JSON schema
````
{"answer": "Hello!"}
````

### regex
````
Hello!
````

### complex JSON schema
````
{"answer": {"text": "Hello", "tone": "formal", "language": "en"}, "alternatives": ["friendly", "playful", "serious"], "metadata": {"confidence": 95, "safe": true, "tags": ["formal", "playful", "serious"]}}
````

### unconstrained
````
Hello!
````

## Timing (median ms/token, steady state)

| constraint          | gemma-4-E2B ms/tok | gemma-4-E2B engine | gemma-4-E2B overhead | LFM2.5-2.6B ms/tok | LFM2.5-2.6B engine | LFM2.5-2.6B overhead | granite-4.0-h-350m-ONNX ms/tok | granite-4.0-h-350m-ONNX engine | granite-4.0-h-350m-ONNX overhead |
| ------------------- | ------------------ | ------------------ | -------------------- | ------------------ | ------------------ | -------------------- | ------------------------------ | ------------------------------ | -------------------------------- |
| simple JSON schema  | 26.86              | 0.07               | 0.23                 | 9.48               | 0.06               | 0.16                 | 20.18                          | 0.1                            | -2.49                            |
| regex               | 26.82              | 0.215              | 0.19                 | 9.77               | 0.115              | 0.45                 | 21.28                          | 0.09                           | -1.39                            |
| complex JSON schema | 28.3               | 0.215              | 1.67                 | 9.35               | 0.11               | 0.03                 | 20.75                          | 0.105                          | -1.92                            |
| unconstrained       | 26.63              |                    |                      | 9.32               |                    |                      | 22.67                          |                                |                                  |

**Comment by jeromeetienne** [comment]
Thank you for this. We have taken the rewritten package end to end in a downstream project and it works, so this is a report of what we measured, including two things this pull request does not mention.

We are [WebAI@Home](https://github.com/webai-at-home/webai-at-home), a volunteer compute cluster whose workers run models in ordinary browser tabs. We vendored the built bundle of `@huggingface/transformers-response-constraint` at commit `7c4593c6b43ad5975865184b3f79e17535b99eeb`, against the released `@huggingface/transformers` 4.2.0, and now serve OpenAI-style `response_format` for `onnx-community/gemma-4-E2B-it-ONNX` at `q4f16` on WebGPU. Every number below was written by a live browser run on an `apple metal-3` adapter with `shader-f16`, greedy decoding (`do_sample: false`), which is what our stage uses.

## 1. Against the released 4.2.0 the package silently constrains nothing

`ConstraintLogitsProcessor` masks in `_call` and advances its grammar in `onTokensSampled`. That hook is added to `LogitsProcessorList` and called from the generation loop by this pull request. The package's peer dependency is `^4.2.0`, which npm satisfies with the released 4.2.0, where the hook does not exist and is never called.

Used exactly as the package README says, the grammar then masks every step from its opening state and never leaves it. There is no error. The same schema and prompt:

```text
{"          … 253 more tokens of whitespace (256 tokens, 18744 ms, cut off by the token limit)
```

versus, with the calls made:

```text
{"city": "Paris"}          (8 tokens, 608 ms)
```

This is the one failure mode that is not fail-closed, and it is the one a downstream user is most likely to hit, because npm will happily install the combination. Two things would help: a peer range that cannot be satisfied by a core without the hook, and a check at construction time that the runtime dispatches it.

Worth knowing for its own sake: **the hook can be emulated without any core change.** The sampled token is already in `input_ids` — `PreTrainedModel.generate` appends it to `all_input_ids` right after sampling and hands the same array to every logits processor on the next step and to every stopping criterion on this one. We read the tokens that appeared since we last looked and call `onTokensSampled` ourselves, from both places, in about 60 lines and with no change to the runtime. Announcing from the stopping criterion as well is not an optimisation: a logits processor runs before the sampler, so the token that completes the grammar would only ever be seen on a step that never happens. With both halves, the answer ends on `<eos>` on the step it completes.

## 2. Flexible whitespace is a trap for greedy decoding

This is the finding we did not expect. `transition` in `engine/json.ts` returns the same state for every whitespace byte wherever JSON allows whitespace, when `whitespaceFlexible` is at its default of `true`. Under greedy decoding that is a fixed point: whenever the highest-scoring allowed token is a space or a line break, the next step faces exactly the same choice and takes it again, until the token limit ends the answer.

Five of seven schemas we sent fell into it. It is not only a leading-whitespace problem — the nested case wrote a complete inner object and then looped, because a required key was still missing, the model would rather have written a key of its own, and whitespace was the only thing it preferred that the grammar allowed:

```text
\n{\n  "weather": {\n    "celsius": 15,\n    "sky": "Partly Cloudy"\n  }\n  \n  \n  …
```

| schema | as sent | with `whitespace_flexible: false` |
| --- | --- | --- |
| one required string | `{"city": "Paris"}`, 8 tokens | `{"city":"]Paris"}`, 7 tokens |
| an integer | 256 tokens of whitespace | `{"sides":6}`, 7 tokens |
| a boolean | 256 tokens of whitespace | `{"answer":true}`, 6 tokens |
| an enumeration | 256 tokens of whitespace | `{"sky":"cloudy"}`, 10 tokens |
| an array of three strings | 31 tokens | the same, compact, 16 tokens |
| an object inside an object | 256 tokens of whitespace after a complete inner object | `{"weather":{"sky":"","celsius":22.000},"city":"Paris"}`, 24 tokens |
| `minLength` and `maxLength` | 256 tokens of whitespace | `{"material":"]The Eiffel,"}`, 10 tokens |

`x-guidance: { whitespace_flexible: false }` on the root schema closes the loop — and it is not free. Denying the whitespace the model wanted makes it prefer a worse token: twice it opened a string with `]`, from token 1935 (`"]`), chosen after the space it wanted was refused. Both answers satisfy their schemas; both are worse than what the model would otherwise have written.

**The two separator options are what actually fixes it.** Setting `key_separator: ": "` and `item_separator: ", "` beside `whitespace_flexible: false` gives back exactly the whitespace the model was reaching for, as fixed bytes with nothing to loop on. With all three options the same one-string schema wrote `{"city": "Paris"}` rather than `{"city":"]Paris"}`, live through our worker. That combination is what we ship.

Two suggestions, both cheap: say in the README that `whitespaceFlexible: true` and greedy decoding do not mix, and consider making the three `x-guidance` options reachable for a `json_object` too — today they can only ride on a root schema, so a caller wanting them has to send the `json_schema` of `{"type":"object"}` instead of a `json_object`.

## 3. The cost, measured

- **62.6 ms per token unconstrained against 69.3 ms constrained**, about 11 per cent, on the same prompt. Your "suuuper little overhead" holds on our hardware and our model.
- Building one constraint: 0 to 3 ms.
- `ResponseConstraint.warmup` on this tokenizer: 1128 to 1534 ms, once per tokenizer. A caller that does not call it pays that inside its first request, which for us was worth moving to model load time.
- The built bundle: 92813 bytes of JavaScript, 21178 gzipped, no WebAssembly module, no runtime dependency. For a page a volunteer loads, that is nothing beside a 3111 MB model.

## 4. Smaller things, all of them good news

- **The real call shape works.** `do_sample: false`, a `TextStreamer`, and our own `InterruptableStoppingCriteria` all coexist with the constraint: `StoppingCriteriaList.extend` takes a criterion, a list, or an array, so both criteria go in one list. Byte for byte the same answers as without the streamer, each ending in `<eos>`.
- **The logical batch size is 1** on every step of a `pipeline('text-generation', …)` call, and the logits are `[1, 262144]`, so the stated limitation never bit us.
- **A schema it cannot enforce throws while the constraint is built**, with the offending reference in the message. That let us refuse such a request at submission by asking the package rather than keeping a keyword list of our own, which is exactly what we wanted. Refused, each by name: an external `$ref`, a local `$ref` that does not resolve, `$dynamicRef`, `unevaluatedProperties`, `unevaluatedItems`, any undefined keyword, and a schema that is neither a boolean nor an object. Accepted: everything an OpenAI-style request usually carries, including `description`, `title`, `default`, `examples`, `$schema`, `$id`, `additionalProperties`, `format`, a local `$ref` into `$defs`, and `minLength`/`maxLength`.
- **`22.000` under `{"type":"integer"}` is correct, not a defect**, and the reasoning in `engine/json.ts` is right: a prefix such as `0.9` would leave nothing viable that can close. `JSON.parse` returns `22`, and a consumer was going to parse the answer anyway.
- **`extractTokenizer` accepts `GemmaTokenizer`** through `_tokenizerJSON` and `eos_token_id`.

## 5. One request

The package is not published, and `"link:../transformers.js/packages/transformers-response-constraint/"` is not something a downstream project can repeat. We vendored the built bundle with a script that fetches the pinned commit and reproduces the checked-in files byte for byte, which works but pins us to one commit by hand. A published preview version, even `0.0.0-alpha`, would let people depend on this properly.

Happy to answer questions about any of the runs above, and happy to rerun anything against a newer commit of the branch.

packages/transformers/src/generation/logits_process.js
```diff
@@ -88,6 +88,19 @@ export class LogitsProcessorList extends Callable {
...
     }
 
+    /**
+     * Calls Transformers.js-specific post-sampling hooks on processors that need to update state after token selection.
+     * The hook observes the full batch after the current generation step has been appended.
+     *
+     * @param {number[]} token_ids The sampled token IDs for the current generation step.
+     * @param {bigint[][]} input_ids The input IDs after appending the sampled tokens.
+     */
+    onTokensSampled(token_ids, input_ids) {
```
xenova: do you think there's a way to avoid needing this additional `onTokensSampled` function? I don't think the python library has this, and it could cause an unnecessary divergence

nico-martin: Good point. It was initially added because the constraint parser must update its state after a token is selected. The logits processor runs before sampling, so it knows which tokens are allowed but not which token was chosen.

We can avoid adding the Transformers.js-specific onTokensSampled hook by committing the sampled token from the existing stopping criteria, which already receives the updated input_ids after sampling. I’ve updated the PR and added a cursor to ensure repeated stopping-criteria calls do not process the same token twice. This keeps the change local to ResponseConstraint and avoids divergence from the Python API.

nico-martin:
@jeromeetienne thanks for taking the time to test it! I really appreciate it.

**Whitespace problem**
I’ve seen the same behavior with Gemma 4 E2B. Constraints filter out invalid tokens, but they do not increase the probability of valid structural tokens. At the start of generation, both a newline and { are valid JSON. If the model ranks newline higher, it can keep selecting newlines because whitespace does not advance the JSON parser.

There are several ways to address this:                                                                        
                                                                                                                    
1. Add an instruction such as “Always respond with valid JSON” to the system prompt. This changes the model’s probability distribution and usually increases the score of {
2. Use a model that naturally ranks JSON structural tokens higher in this context
3. Set repetition_penalty. This makes repeatedly selecting the same newline token less likely, although it also affects other repeated tokens
4. Set whitespace_flexible: false. This forbids optional JSON whitespace, so newline tokens are filtered out
5. Apply a targeted whitespace repetition penalty inside the response constraint

I implemented 5. in `discourageRepeatedWhitespace` https://github.com/huggingface/transformers.js/pull/1733/changes/47e71193286d0c668c96a729ec05a8e1a95b3259. It progressively penalizes whitespace-only tokens that do not advance the JSON parser and eventually masks them to guarantee progress. Whitespace inside strings or tokens that also advance the parser is unaffected.

The separator workaround should no longer be necessary with the repeated-whitespace protection. Flexible whitespace remains enabled, allowing the model to produce its preferred formatting, while consecutive whitespace-only tokens are progressively penalized and eventually masked to guarantee progress.

**Publishing**
Agreed that vendoring a pinned bundle is not a good downstream experience. We will look into publishing an alpha version once the PR is ready.

nico-martin:
closed in favor of https://github.com/huggingface/transformers.js/pull/1758

PR #1737, opened 2026-08-02
fix: dispose KV cache after Chatterbox generation (#1734)

### LINKED ISSUE #1734: Chatterbox KV cache leak
### System Info

```
@huggingface/transformers  4.2.0
onnxruntime-web            1.26.0-dev.20260416-b7804b056c (as pinned by the above)

Browser                    Chrome 149.0.7827.200
OS                         Linux 7.0.14-arch1-1 (Arch Linux)
GPU                        NVIDIA GeForce RTX 5090, driver 610.43.02
WebGPU backend             Dawn / Vulkan

Model                      onnx-community/chatterbox-ONNX
device                     "webgpu"
dtype                      language_model q4; embed_tokens / speech_encoder /
                           conditional_decoder fp32
```
Observed through a browser application built with Vite 6, React 19 and TypeScript 5.7, running inference inside a Web Worker. Neither the framework nor the worker is required to reproduce it — the loop below is enough.

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension

### Description

Every call to `ChatterboxModel.generate` abandons a full KV cache of GPU buffers, so GPU memory grows without bound in any application that synthesizes repeatedly. A long-running session reached roughly **24 GB** before it was noticed.

`generate()` disposes the cache only when it is not returning it:

```js
const keepCacheAlive = "past_key_values" in kwargs || generation_config.return_dict_in_generate;
if (!keepCacheAlive) {
}
```

`ChatterboxModel.generate` always opts into the dict form, because it needs `audio_tokens`, `speaker_embeddings` and `speaker_features` alongside the sequences:

```js
async generate(params) {
  const { sequences, audio_tokens, speaker_embeddings, speaker_features } =
    await super.generate({
      ...params,
      return_dict_in_generate: true,
    });
  // ... builds speech_tokens, runs conditional_decoder, returns waveform
}
```

`keepCacheAlive` is therefore permanently true for this architecture and the disposal never runs. The wrapper then destructures the four values it needs and lets `past_key_values` go out of scope **without disposing it**, leaving its GPU buffers allocated. `DynamicCache.dispose()` already exists and does the right thing; nothing on this path calls it.

Each leak is up to max_new_tokens (256 in the runs below) of keys and values across every layer.

**Expected:** GPU memory stays flat across repeated generations, since the caller never asked to keep a cache and never receives one.

**Actual:** it rises monotonically, in proportion to the number of generations rather than to elapsed time, and is only reclaimed by reloading the page.

Measured across two phases of one session:

| phase | duration | GPU memory growth |
| --- | --- | --- |
| idle, nothing synthesized | ~4 h | 15 MB |
| continuous synthesis | ~1 h | to ~24 GB |

Lowering `max_new_tokens` reduces the size of each leak proportionally but does not stop the accumulation.

Reproduction Results

This was run in Chrome on WebGPU (RTX 5090, Vulkan backend), sampling the GPU
process once a second with `nvidia-smi`. Model files were served from
localhost, so nothing is competing for bandwidth.

With `max_new_tokens: 256`, GPU memory rises in exact steps, one
per generation, and **never falls back**:

```
      MiB     (one step per generate, ~4.8 s apart)
2983 -> 3103 -> 3223 -> 3345 -> 3465 -> 3585 -> 3705 -> 3825
     -> 3945 -> 4065 -> 4185 -> 4305 -> 4425 -> 4545 -> 4665
                                        ^ loop ends, flat from here
```

**+120 MiB per generation**, held until the page is closed.

Repeating with `max_new_tokens: 64` gives roughly **+65 MiB per generation**
(coarser: at 1.27 s per generation the 1 Hz sampling straddles calls).

That is not proportional to the token count: a quarter of the tokens still leaks over half the memory (65 of 120 MiB), where proportionality predicts 30 MiB. Fitting the two points gives roughly 47 MiB fixed plus 0.29 MiB per token.

**47 MiB fixed plus 0.29 MiB per token**. The token-proportional part is
consistent with the abandoned KV cache; the fixed part suggests something else
is also being kept per call, and identifying it is left to whoever picks this
up rather than guessed at here.

### Reproduction

```js
import { AutoProcessor, ChatterboxModel, Tensor } from "@huggingface/transformers";

const model_id = "onnx-community/chatterbox-ONNX";

const model = await ChatterboxModel.from_pretrained(model_id, {
  device: "webgpu",
  dtype: {
    embed_tokens: "fp32",
    speech_encoder: "fp32",
    language_model: "q4",
    conditional_decoder: "fp32",
  },
});
const processor = await AutoProcessor.from_pretrained(model_id);

// Any few seconds of mono reference audio works; silence is enough to
// exercise the path.
const reference = new Tensor("float32", new Float32Array(72000), [1, 72000]);
const speaker = await model.encode_speech(reference);

for (let i = 0; i < 200; i++) {
  const inputs = await processor("This is a short line of speech.");

  // The waveform is the only thing returned, so past_key_values is
  // unreachable from here and cannot be disposed by the caller.
  const waveform = await model.generate({
    ...inputs,
    ...speaker,
    exaggeration: 0.5,
    max_new_tokens: 256,
  });

  console.log(i, waveform.dims);
}
// GPU memory has climbed by several GB by now and never comes back down.
```

m96-chan:
## Summary

Fixes #1734.

`ChatterboxModel.generate` leaked a full KV cache of GPU buffers on every call, so GPU memory grew without bound in any application that synthesizes repeatedly.

## Root Cause

`PreTrainedModel.generate` only disposes the cache when it is *not* handing it back to the caller:

```js
const keepCacheAlive = 'past_key_values' in kwargs || generation_config.return_dict_in_generate;
if (!keepCacheAlive) {
}
```

ChatterboxModel.generate always opts into the dict form, because it needs audio_tokens, speaker_embeddings and speaker_features alongside sequences.
keepCacheAlive is therefore permanently true for this architecture, so the disposal in the base class never runs. The wrapper then destructured only the four values it needed and let past_key_values go out of scope without disposing it,
leaving its GPU buffers allocated — up to max_new_tokens worth of keys and values across every layer, per call. The caller of ChatterboxModel.generate receives only a waveform, so nothing downstream can free it either.

## Fix

Destructure past_key_values in ChatterboxModel.generate and dispose it once the
conditional_decoder run is finished:

await past_key_values?.dispose();

The dispose happens after conditional_decoder has run, so the cache stays alive for as long as generation actually needs it. DynamicCache.dispose() already exists and does the right thing; this simply calls it on the one path that never did.
The optional chaining keeps the change safe for configurations where no cache is returned.

## Impact

Measured in Chrome on WebGPU (RTX 5090, Dawn/Vulkan), onnx-community/chatterbox-ONNX, 20 generations at max_new_tokens: 256, GPU process sampled once a second:

growth over 20 generations
before:  
  +1682 MiB — twenty clean steps of ~120 MiB
after:
   +36 MiB — flat at 2422 MiB from the third generation onward

The remaining residue is one-off rather than per-call: once it settles it does not move again for the rest of the run. Generation speed is unchanged (5.3 s per call before, 5.3–5.7 s after), so nothing is traded away for it.

Scope is limited to ChatterboxModel.generate; no public API or behavior change — callers already only received the waveform.

### DIFF
--- a/packages/transformers/src/models/chatterbox/modeling_chatterbox.js
+++ b/packages/transformers/src/models/chatterbox/modeling_chatterbox.js
@@ -162,26 +162,37 @@ export class ChatterboxModel extends ChatterboxPreTrainedModel {
 
     /** @type {PreTrainedModel['generate']} */
     async generate(params) {
-        const { sequences, audio_tokens, speaker_embeddings, speaker_features } = /** @type {any} */ (
+        // A caller-supplied cache is updated in place and handed back as the same object,
+        // so keep a reference to tell it apart from a cache created for this call.
+        const supplied_past_key_values = params.past_key_values;
+
+        const { sequences, audio_tokens, speaker_embeddings, speaker_features, past_key_values } = /** @type {any} */ (
             await super.generate({
                 ...params,
                 return_dict_in_generate: true,
             })
         );
 
-        const new_tokens = sequences.slice(null, [
-            /** @type {Tensor} */ (params.input_ids).dims[1], // Exclude start of speech token
-            -1, // Exclude end of speech token
-        ]);
+        try {
+            const new_tokens = sequences.slice(null, [
+                /** @type {Tensor} */ (params.input_ids).dims[1], // Exclude start of speech token
+                -1, // Exclude end of speech token
+            ]);
 
-        const silence_tokens = full([new_tokens.dims[0], 3], SILENCE_TOKEN); // Add 3 silence tokens
-        const speech_tokens = cat([audio_tokens, new_tokens, silence_tokens], 1);
+            const silence_tokens = full([new_tokens.dims[0], 3], SILENCE_TOKEN); // Add 3 silence tokens
+            const speech_tokens = cat([audio_tokens, new_tokens, silence_tokens], 1);
 
-        const { waveform } = await sessionRun(this.sessions['conditional_decoder'], {
-            speech_tokens,
-            speaker_features,
-            speaker_embeddings,
-        });
-        return waveform;
+            const { waveform } = await sessionRun(this.sessions['conditional_decoder'], {
+                speech_tokens,
+                speaker_features,
+                speaker_embeddings,
+            });
+            return waveform;
+        } finally {
+            // Only dispose caches created here: a caller-supplied cache stays caller-owned.
+            if (past_key_values !== supplied_past_key_values) {
+                await past_key_values?.dispose();
+            }
+        }
     }
 }
--- a/packages/transformers/src/models/chatterbox/modeling_chatterbox.js
+++ b/packages/transformers/src/models/chatterbox/modeling_chatterbox.js
@@ -162,7 +162,7 @@ export class ChatterboxModel extends ChatterboxPreTrainedModel {
 
     /** @type {PreTrainedModel['generate']} */
     async generate(params) {
-        const { sequences, audio_tokens, speaker_embeddings, speaker_features } = /** @type {any} */ (
+        const { sequences, audio_tokens, speaker_embeddings, speaker_features, past_key_values } = /** @type {any} */ (
             await super.generate({
                 ...params,
                 return_dict_in_generate: true,
@@ -182,6 +182,7 @@ export class ChatterboxModel extends ChatterboxPreTrainedModel {
             speaker_features,
             speaker_embeddings,
         });
+        await past_key_values?.dispose();
         return waveform;
     }
 }
--- a/packages/transformers/tests/models/chatterbox/test_modeling_chatterbox.js
+++ b/packages/transformers/tests/models/chatterbox/test_modeling_chatterbox.js
@@ -1,0 +1,111 @@
+import { jest } from "@jest/globals";
+
+import { ChatterboxModel, PreTrainedModel, Tensor, env, LogLevel } from "../../../src/transformers.js";
+
+import { MAX_TEST_EXECUTION_TIME } from "../../init.js";
+
+// Helper function to create a model with only a fake `conditional_decoder` session, so no weights are loaded
+function createModel(run) {
+  const model = Object.create(ChatterboxModel.prototype);
+  model.sessions = {
+    conditional_decoder: { inputNames: ["speech_tokens", "speaker_features", "speaker_embeddings"], run },
+  };
+  return model;
+}
+
+// Helper function to stub the outputs of `super.generate`, returning the KV cache it creates internally.
+// A supplied cache is updated in place and handed back as the same object, so mirror that here.
+function mockSuperGenerate() {
+  const past_key_values = { dispose: jest.fn(async () => {}) };
+  jest.spyOn(PreTrainedModel.prototype, "generate").mockImplementation(async (params) => ({
+    sequences: new Tensor("int64", [0n, 1n, 2n, 3n, 4n], [1, 5]),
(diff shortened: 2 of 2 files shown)


packages/transformers/src/models/chatterbox/modeling_chatterbox.js
```diff
@@ -182,6 +182,7 @@ export class ChatterboxModel extends ChatterboxPreTrainedModel {
             speaker_features,
             speaker_embeddings,
         });
+        await past_key_values?.dispose();
```
nico-martin: This runs only when all waveform-building work succeeds. If the conditional decoder rejects, the KV cache is abandoned and repeated failed calls can reproduce the GPU leak. Please move disposal into a `finally` block and add a rejecting-decoder test that asserts disposal runs exactly once.

nico-martin (changes requested):
Hi @m96-chan, thank you so much for looking into this! The memory measurements are really helpful, and the fix works for the normal path. Could you move disposal into a `finally` block and add a rejecting-decoder test? I'm holding off on merging because a failed waveform build can still leak the KV cache.

packages/transformers/src/models/chatterbox/modeling_chatterbox.js
m96-chan: Thank you for the review.
I planed to fix!

nico-martin: Thanks! I saw [402c9f2](https://github.com/m96-chan/transformers.js/commit/402c9f2c658249f9fd8b49bc3186dac0788bf028). The try/finally change and rejecting-decoder test address this thread. Please push that commit to the PR branch so CI can run on it, and I’ll take another look.

nico-martin:
One additional lifecycle case: `past_key_values` can be supplied by the caller, and generation updates that same `DynamicCache` in place. Unconditionally disposing it here would invalidate caller-owned state. Could you dispose only caches created internally and add a test confirming that a supplied cache is not disposed?

m96-chan:
> One additional lifecycle case: `past_key_values` can be supplied by the caller, and generation updates that same `DynamicCache` in place. Unconditionally disposing it here would invalidate caller-owned state. Could you dispose only caches created internally and add a test confirming that a supplied cache is not disposed?

Thank you for review!
I am going to fix and run test. If everything was ok, I will push.

m96-chan:
@nico-martin 

Thanks for the follow-up — you're right, and it's fixed in d05e7ae4.

ChatterboxModel.generate now only disposes a cache that was created for the call:

```javascript
const supplied_past_key_values = params.past_key_values;
...
} finally {
    if (past_key_values !== supplied_past_key_values) {
        await past_key_values?.dispose();
    }
}
```
Identity rather than 'past_key_values' in params, because that's the contract getPastKeyValues documents: a supplied cache is updated in place and returned as the same object, a new DynamicCache only when none was given. It also covers an explicit past_key_values: null. Happy to switch to the in form if you'd rather it mirror PreTrainedModel.generate literally.

Tests assert a supplied cache is never disposed, on both the success and rejecting-decoder paths. The branch is squashed, so d05e7ae4 also carries the finally change from the earlier thread. Full suite green locally.

nico-martin:
@m96-chan Thanks, this addresses both lifecycle cases.

nico-martin (approved):

xenova (approved):
thanks! very clean PR

PR #1738, opened 2026-08-03
Fix Moonshine ASR token decoding

### LINKED ISSUE #1735: Moonshine ASR pipeline throws 'Unable to decode without a tokenizer' — _call_moonshine calls processor.batch_decode but processor has no tokenizer
## Bug

The `automatic-speech-recognition` pipeline crashes at decode time for **all** moonshine models (e.g. `onnx-community/moonshine-tiny-ar-ONNX`, `onnx-community/moonshine-tiny-en-ONNX`), in **both 3.8.1 and 4.2.0**.

## Exact error

```
Error: Unable to decode without a tokenizer.
    at Function.batch_decode (transformers.node.mjs:15763)
    at Function._call_moonshine (transformers.node.mjs:32216)
```

## Reproduce (Node, v4.2.0)

```js
import { pipeline, env } from "@huggingface/transformers";
env.allowLocalModels = false;
const transcriber = await pipeline("automatic-speech-recognition", "onnx-community/moonshine-tiny-ar-ONNX", { dtype: "q8", device: "cpu" });
// model loads fine (~3s)
const out = await transcriber(float32Audio16k); // throws at decode
```

## Root cause

`_call_moonshine` calls the **processor's** `batch_decode`:
```js
// src/pipelines/automatic-speech-recognition.js:328 (v4.2.0)
const text = this.processor.batch_decode(outputs, { skip_special_tokens: true })[0];
```

But the moonshine processor is constructed **without a tokenizer** — only a feature_extractor. Verified directly:
```js
const proc = await AutoProcessor.from_pretrained("onnx-community/moonshine-tiny-ar-ONNX");
// proc.components => ['feature_extractor']   (no 'tokenizer')
// proc.config.processor_class => undefined
// proc.config.auto_map => undefined
```
So `processor.batch_decode` hits its `if (!this.tokenizer) throw` guard (`processing_utils.js:94`).

The pipeline-level tokenizer **is** correctly populated:
```js
const p = await pipeline("automatic-speech-recognition", "onnx-community/moonshine-tiny-ar-ONNX", {...});
// p.tokenizer => PreTrainedTokenizer  (loaded from tokenizer.json) ✓
// p.processor.tokenizer => undefined   ✗
```

## Fix

`_call_moonshine` should use the **pipeline's** tokenizer, not the processor's:
```js
const text = this.tokenizer.batch_decode(outputs, { skip_special_tokens: true })[0];
```

This matches every other ASR path — `_call_whisper` uses `this.tokenizer._decode_asr`, and `_call_wav2vec2` uses `this.tokenizer.decode`. `_call_moonshine` is the only one that routes through `this.processor`, and the processor legitimately may not carry a tokenizer.

## Environment
- `@huggingface/transformers`: 3.8.1 (reproduced) and 4.2.0 (reproduced)
- Model: `onnx-community/moonshine-tiny-ar-ONNX` (and the en variant)
- Runtime: Node 22 (onnxruntime-node cpu) and browser (onnxruntime-web wasm)

Thank you.

Mr-Neutr0n:
## What does this PR do?

Fixes Moonshine automatic speech recognition pipelines that fail during token decoding.

Moonshine processors can contain only a feature extractor, while the tokenizer is loaded separately on the pipeline. `_call_moonshine` currently asks the processor to decode generated token IDs, which raises `Unable to decode without a tokenizer`. This change uses the pipeline tokenizer, matching the other ASR paths.

The regression test constructs a processor without a tokenizer and verifies that generated tokens are decoded through the pipeline tokenizer.

Fixes #1735.

## Tests

- focused Moonshine Jest test passes
- Prettier checks pass for both changed files
- `pnpm build` passes, including TypeScript type generation

### DIFF
--- a/packages/transformers/src/pipelines/automatic-speech-recognition.js
+++ b/packages/transformers/src/pipelines/automatic-speech-recognition.js
@@ -325,7 +325,7 @@ export class AutomaticSpeechRecognitionPipeline
             const max_new_tokens = Math.floor(aud.length / sampling_rate) * 6;
             const outputs = await this.model.generate({ max_new_tokens, ...kwargs, ...inputs });
 
-            const text = this.processor.batch_decode(/** @type {Tensor} */ (outputs), { skip_special_tokens: true })[0];
+            const text = this.tokenizer.batch_decode(/** @type {Tensor} */ (outputs), { skip_special_tokens: true })[0];
             toReturn.push({ text });
         }
         return single ? toReturn[0] : toReturn;
--- a/packages/transformers/tests/pipelines/test_pipelines_automatic_speech_recognition.js
+++ b/packages/transformers/tests/pipelines/test_pipelines_automatic_speech_recognition.js
@@ -458,5 +458,43 @@ export default () => {
         await pipe?.dispose();
       }, MAX_MODEL_DISPOSE_TIME);
     });
+
+    describe("moonshine", () => {
+      it("decodes generated tokens with the pipeline tokenizer", async () => {
+        const sampling_rate = 16000;
+        const generated_tokens = [[1n, 2n, 3n]];
+        let decode_args;
+
+        const processor = Object.assign(async (audio) => ({ input_values: audio }), {
+          feature_extractor: { config: { sampling_rate } },
+        });
+        const tokenizer = {
+          batch_decode(outputs, options) {
+            decode_args = { outputs, options };
+            return ["decoded transcript"];
+          },
+        };
+        const model = {
+          config: { model_type: "moonshine" },
+          async generate() {
+            return generated_tokens;
+          },
+        };
+        const pipe = new AutomaticSpeechRecognitionPipeline({
+          task: PIPELINE_ID,
+          model,
+          tokenizer,
+          processor,
+        });
+
+        const output = await pipe(new Float32Array(sampling_rate));
+
+        expect(output).toEqual({ text: "decoded transcript" });
+        expect(decode_args).toEqual({
+          outputs: generated_tokens,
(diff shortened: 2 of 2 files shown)


nico-martin (approved):
Hi @Mr-Neutr0n, thank you so much for looking into this! This looks good to me. Using the pipeline tokenizer is the right fix, and the test covers the failing Moonshine setup through the public path. I'd merge this one over the later duplicate #1746.

xenova:
Hi! 👋 thanks for the PR. I looked into this a bit and was initially confused because `onnx-community/moonshine-tiny-ONNX` works, but `onnx-community/moonshine-tiny-ar-ONNX` didn't. After a bit of digging, I saw it was because the `ar` version didn't have `processor_class` set:

https://huggingface.co/onnx-community/moonshine-tiny-ONNX/blob/main/preprocessor_config.json:
```json
{
  "feature_extractor_type": "MoonshineFeatureExtractor",
  "processor_class": "MoonshineProcessor",
  "sampling_rate": 16000
}
```

https://huggingface.co/onnx-community/moonshine-tiny-ar-ONNX/blob/main/preprocessor_config.json:
```json
{
  "do_normalize": false,
  "feature_extractor_type": "Wav2Vec2FeatureExtractor",
  "feature_size": 1,
  "padding_side": "right",
  "padding_value": 0.0,
  "return_attention_mask": true,
  "sampling_rate": 16000
}
```

and technically moonshine doesn't require an audio preprocessor because it uses raw audio waveform as input, so we don't need to load the `preprocessor_config.json`.

However, the change we have here is sound, so I'll merge. Thanks!

PR #1594, opened 2026-03-18
Overdue Whisper fixes

xenova:
Did a deep dive recently to uncover and fix issues with whisper models.

- Closes https://github.com/huggingface/transformers.js/issues/551
- Closes https://github.com/huggingface/transformers.js/issues/805
- Closes https://github.com/huggingface/transformers.js/issues/1357 (tested this one specifically and output is much better than described in the issue).
- Closes https://github.com/huggingface/transformers.js/issues/1358
- Closes https://github.com/huggingface/transformers.js/issues/1590

Also add spectrogram unit tests
- Closes https://github.com/huggingface/transformers.js/issues/1387

PR #1742, opened 2026-08-12
Fix Whisper ASR timestamp regression

### LINKED ISSUE #1684: ASR (Whisper) regression for detected timestamps
### System Info

transformers.js 4.2.0

### Environment/Platform

- [x] Website/web-app
- [ ] Browser extension

### Description

Using transformers.js 4.2.0 `automatic-speech-recognition` and standard `onnx-community/whisper-base` model, produces timestamps that are less accurate compared to v3.8.1

### Reproduction

Run the following demo.html in browser:

```html
<script type="module">
//import { pipeline } from "https://cdn.jsdelivr.net/npm/@huggingface/transformers@3.8.1";
import { pipeline } from "https://cdn.jsdelivr.net/npm/@huggingface/transformers@4.2.0";
const audio = new Float32Array(await (await fetch("input.pcm")).arrayBuffer());
const pipe = await pipeline("automatic-speech-recognition",
	"onnx-community/whisper-base",
	{dtype:{encoder_model:"fp32", decoder_model_merged:"q4"},
	device:"webgpu"});

	return_timestamps: true});

console.log(result.chunks.map(chunk => `${chunk.timestamp[0]} -> ${chunk.timestamp[1]} ${chunk.text}`))
</script>
```

which prints for 3.8.1:

```
0:"0 -> 7  This is a day I've been looking forward to for two and a half years."
1:"7 -> 13  [applause]"
2:"13 -> 21  Every once in a while, a revolutionary product comes along that changes everything."
3:"21 -> 29  And Apple has been, well first of all, one's very fortunate if you get to work on just one of these,"
```

while using the very same model running with 4.2.0 prints:

```
0:"0 -> 7  This is a day I've been looking forward to for two and a half years."
1:"7 -> 22  Every once in a while, a revolutionary product comes along that changes everything."
2:"22 -> 28.92  And Apple has been, well first of all, one's very fortunate if you get to work on just one"
3:"28.92 -> 29.92  of these."
```

Notice the difference in timing the sentence `Every once in a while...` which actually starts around `13` second mark (hearing attached input audio [input.zip](https://github.com/user-attachments/files/27711520/input.zip)). It is correctly recognized by 3.8.1 but 4.2.0 output is `7`.

- This is not a model problem, as both outputs are run on the very same model.
- The issue looks similar to the one reported on https://github.com/huggingface/transformers.js/issues/1590 which exposes regression in timing issue with tjs v4 as well (while using another model)

Could you please look into the issue or provide a workaround?

anshusaurav:
## Summary

Fixes #1684 — Whisper ASR timestamp regression between v3.8.1 and v4.2.0.

Three fixes targeting the seek loop and logits processor introduced in commit `43b2662` ("Overdue Whisper fixes #1594"):

- **Use actual audio length for seek loop bounds** (`modeling_whisper.js`): The seek loop used `input_features.dims[2]` (always 3000 = padded 30s) as `total_frames` instead of `generation_config.num_frames` (set by the ASR pipeline to the real mel frame count). For clips shorter than 30 seconds, this caused the loop to process silence/padding as real audio, producing hallucinated text and incorrect timestamps.

- **Apply `max_initial_timestamp_index` constraint** (`logits_process.js`): A `continue` statement in `WhisperTimeStampLogitsProcessor._call()` skipped the `max_initial_timestamp_index` check at line 326, making it dead code. This allowed the model to generate any timestamp as its first token instead of being constrained to the allowed range. The Python reference implementation has no such early exit — both blocks execute sequentially.

- **Guard against infinite seek loop** (`modeling_whisper.js`): If `segment_offset` computes to zero (e.g., from a degenerate timestamp pair), the seek loop would run forever. Added a `break` when `segment_offset <= 0`.

## Test plan

- [ ] Verify with the reproduction case from #1684 (Steve Jobs keynote audio) — timestamps should match v3.8.1 output
- [ ] Verify `[applause]` segment is no longer dropped
- [ ] Existing Whisper prefix token tests (`test_modeling_whisper.js`) still pass — they use `max_new_tokens` which bypasses the seek loop
- [ ] Test with audio shorter than 30s — seek loop should terminate after one iteration
- [ ] Test with audio longer than 30s — multi-segment seek should still work correctly

### DIFF
--- a/packages/transformers/src/models/whisper/modeling_whisper.js
+++ b/packages/transformers/src/models/whisper/modeling_whisper.js
@@ -201,7 +201,7 @@ export class WhisperForConditionalGeneration extends WhisperPreTrainedModel {
 
         // input_features shape: [batch=1, n_mels, total_frames]
         const input_features = inputs;
-        const total_frames = generation_config.num_frames
+        const total_frames = generation_config.num_frames != null
             ? Math.min(input_features.dims[2], generation_config.num_frames)
             : input_features.dims[2];
 
--- a/packages/transformers/src/models/whisper/modeling_whisper.js
+++ b/packages/transformers/src/models/whisper/modeling_whisper.js
@@ -201,7 +201,9 @@ export class WhisperForConditionalGeneration extends WhisperPreTrainedModel {
 
         // input_features shape: [batch=1, n_mels, total_frames]
         const input_features = inputs;
-        const total_frames = input_features.dims[2];
+        const total_frames = generation_config.num_frames
+            ? Math.min(input_features.dims[2], generation_config.num_frames)
+            : input_features.dims[2];
 
         // The encoder downsamples by input_stride (=2 for whisper), so:
         //   num_segment_frames = input_stride * max_source_positions = 3000 mel frames per segment
@@ -343,6 +345,7 @@ export class WhisperForConditionalGeneration extends WhisperPreTrainedModel {
             if (seek_token_timestamps) {
                 allTokenTimestamps.push(...seek_token_timestamps.slice(0, tokens_to_keep));
             }
+            if (segment_offset <= 0) break;
             seek += segment_offset;
         }
 
--- a/packages/transformers/src/generation/logits_process.js
+++ b/packages/transformers/src/generation/logits_process.js
@@ -304,7 +304,6 @@ export class WhisperTimeStampLogitsProcessor extends LogitsProcessor {
 
             if (input_ids[i].length === this.begin_index) {
                 batch_logits_data.subarray(0, this.timestamp_begin).fill(-Infinity);
-                continue;
             }
 
             // timestamps have to appear in pairs, except directly before eos_token; mask logits accordingly
--- a/packages/transformers/tests/models/whisper/test_modeling_whisper.js
+++ b/packages/transformers/tests/models/whisper/test_modeling_whisper.js
@@ -1,4 +1,4 @@
-import { WhisperTokenizer, WhisperForConditionalGeneration, full } from "../../../src/transformers.js";
+import { WhisperTokenizer, WhisperForConditionalGeneration, full, Tensor } from "../../../src/transformers.js";
 
 import { MAX_MODEL_LOAD_TIME, MAX_TEST_EXECUTION_TIME, MAX_MODEL_DISPOSE_TIME, DEFAULT_MODEL_OPTIONS } from "../../init.js";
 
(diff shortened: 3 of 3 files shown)


packages/transformers/src/models/whisper/modeling_whisper.js
```diff
@@ -201,7 +201,9 @@ export class WhisperForConditionalGeneration extends WhisperPreTrainedModel {
 
         // input_features shape: [batch=1, n_mels, total_frames]
         const input_features = inputs;
-        const total_frames = input_features.dims[2];
+        const total_frames = generation_config.num_frames
```
nico-martin: `num_frames` can legitimately be `0`. This truthiness check then falls back to the padded 3,000-frame tensor and reintroduces padding processing. Please distinguish an omitted value from zero, for example with `generation_config.num_frames != null`, and add a zero/short-frame regression test.

nico-martin (changes requested):
Hi @anshusaurav, thank you so much for looking into this! The diagnosis and overall direction make sense. One edge case remains: `num_frames` can be `0`, so the truthy check falls back to the padded tensor. Could you use a nullish check and add direct tests for short audio, timestamp masking, and the non-progress case? I'm holding off on merging until the zero-frame bug and untested seek-loop changes are covered.

anshusaurav:
Thanks for the thorough review @nico-martin!

Just pushed the requested changes:

1. **Nullish check** — changed `generation_config.num_frames` (truthy) to `generation_config.num_frames != null` so that `num_frames=0` is correctly treated as zero frames rather than falling back to the padded tensor length.

2. **Tests added:**
   - `num_frames=0` — verifies the seek loop does not execute on zero frames
   - Short audio (500 frames < 3000 segment) with timestamps — verifies no crash on sub-segment input
   - Timestamp token presence — verifies `return_timestamps=true` produces timestamp tokens in the output

nico-martin:
Thanks, the nullish check correctly fixes the num_frames=0 case. The added tests still do not exercise the changed seek-loop behavior, however: each passes max_new_tokens, and generate() explicitly bypasses _generate_with_seek() whenever that option is present.

Could you adjust the coverage so that:

- zero frames and padded short audio run through the seek path without max_new_tokens
- the short-audio case uses a padded 3000-frame tensor with num_frames: 500
- timestamp masking asserts that initial timestamp IDs above timestamp_begin + max_initial_timestamp_index are masked, preferably with a direct WhisperTimeStampLogitsProcessor unit test
- a mocked degenerate timestamp pair produces segment_offset === 0 and terminates after one pass
- normal multi-segment output advances through the expected passes?

The current timestamp-presence assertion and short-input smoke test would pass with the corresponding fixes reverted. Also, the added Tensor import is currently unused.

PR #1745, opened 2026-08-13
Add StaticCache: fixed-capacity in-place KV cache for WebGPU

### LINKED ISSUE #1741: [WebGPU] WebGPU decode permanently degrades ~3x at S=2048/4096/6144
## System info

- Transformers.js 4.2.0 (code paths re-verified on `main` @ `353007be`)
- onnxruntime-web 1.29, Chrome / Windows 11, Intel iGPU (Panther Lake)
- Phi-4-mini-instruct ONNX q4f16 (merged decoder, 64 KV tensors); cross-checked with Qwen3.5-4B

## Observation

Per-token decode latency jumps as **permanent steps at fixed absolute positions**, reproducible bit-exactly across runs:

| position S | ms/token |
| --- | --- |
| < 2048 | ~46 |
| 2048–4096 | ~138 |
| 4096–6144 | ~233 |
| > 6144 | ~298 |

With prefill=2048, decode drops 25.6 → 7.2 TPS.

### Root cause

Each decode step, DynamicCache + `preferredOutputLocation: 'gpu-buffer'` makes ORT allocate 64 fresh `present.*` GPU buffers and release the previous 64. The WebGPU EP's `BucketCacheManager` (`buffer_manager.cc`) rounds sizes up to buckets (4 / 8 / **12** / 16 / 25 MiB…) and retains at most 10–20 buffers per bucket. 64 ≫ cap, so `64 − cap` buffers are truly created (alloc + zero-fill) and destroyed **every step**. Each bucket boundary the KV size crosses shifts the per-step byte churn → one permanent step each at S=2048/4096/6144 (KV tensor = `2048·S` bytes).

Supporting evidence:

- Step at **S=6144** = the 12→16 MiB boundary. 12 MiB is not a power of two; only the bucket table predicts a step there.
- **Qwen3.5-4B control**: only 16 growing KV tensors → fits under the caps, no comparable staircase. The table even predicts a *negative* step at its 16→25 MiB boundary (cap 10→15): measured −6.0 ms.
- **Zero-churn control**: same loaded session driven with a pre-allocated static KV cache (`present.*` bound in-place onto past buffers via `session.run(feeds, fetches)`) → staircase gone, decode 6.6 → **19.6 TPS**, within 1.9% of a native ORT static-KV baseline.
-  **Correctness**: static-KV greedy output is token-identical to DynamicCache across all three boundaries (2× dynamic + 2× static runs, with self-consistency controls).

Allocator-side report filed against onnxruntime-web: https://github.com/microsoft/onnxruntime/issues/32017 . Both sides are worth fixing: caps that low hurt any dynamic-KV workload, but a static cache is also the standard cure elsewhere (Python `StaticCache`, ORT GenAI).

### Proposal: opt-in `StaticCache`

#1638 already routes user-provided `past_key_values` through `generate()`, so this needs no new options:

```js
const past_key_values = new StaticCache({ max_cache_len: 4096 });
await generator(messages, { max_new_tokens: 256, past_key_values });
```

Allocate each cache entry once at `max_cache_len`, bind `present.*` outputs in-place onto `past_*` inputs via fetches → zero per-step allocation. Composes with #1638: later turns skip recomputation *and* reallocation.

First-version scope:

- Opt-in only; `DynamicCache` stays the default, zero change otherwise
- WebGPU, decoder-only, batch=1; clear errors elsewhere
- In-place binding assumes the graph tolerates shared past/present buffers (true for GQA exports; verified token-exact on Phi-4-mini). A ping-pong double-buffer variant avoids the assumption at 2× cache memory if preferred.

Implementation is ready to submit as two PRs: (1) optional `fetches` pass-through in `sessionRun` (~15 lines, no behavior change), (2) `StaticCache` + `instanceof` branches in `modeling_utils.js`. **Does the direction and API shape look right?**

<details>
<summary>Repro</summary>

1. Load Phi-4-mini-instruct ONNX q4f16, `device: "webgpu"`.
2. Prompt ~1900 tokens, generate 256 greedily, timestamp each `TextStreamer.put`.
3. Per-token latency jumps ~46 → ~138 ms permanently at position ~2050. Prefill ~3900 / ~6000 shows the 4096 / 6144 steps.

</details>

kylo5aby:
Adds an opt-in `StaticCache` that eliminates per-step KV cache reallocation on WebGPU. 

Fixes https://github.com/huggingface/transformers.js/issues/1741

With `DynamicCache` (the default), every decode step replaces every KV tensor with a slightly larger one. On WebGPU each replacement destroys and re-creates a zero-filled GPU buffer, so the per-step cost grows with context length and jumps permanently at the buffer-pool bucket boundaries,  producing a staircase-shaped latency curve.

`StaticCache` allocates each cache entry **once** at a fixed `max_cache_len`, and binds the model's `present.*` outputs in place onto the same GPU buffers via `session.run(feeds, fetches)`. Decoding performs zero cache (re-)allocations per step.

## Usage

```js
const past_key_values = new StaticCache({ max_cache_len: 4096 });
const output = await generator(messages, { max_new_tokens: 256, past_key_values });
await past_key_values.dispose(); // caller owns the cache
```

## Performance

Decode throughput, q4f16 on WebGPU (TPS), 256 new tokens:

| Model | Prefill | DynamicCache | StaticCache | Δ |
|---|---|---|---|---|
| Phi-4-mini-instruct | 2048 | 10.1 | 25.4 | +151% |
| Phi-4-mini-instruct | 4096 | 5.6 | 20.8 | +271% |
| Qwen3.5-4B | 2048 | 17.7 | 19.2 | +8% |
| Qwen3.5-4B | 4096 | 15.1 | 18.2 | +21% |

The gain grows with context length: `DynamicCache` throughput degrades as buffers get larger while `StaticCache` stays nearly flat. Qwen3.5 (hybrid attention) benefits less because only its full-attention layers have growing KV entries. In our measurements `StaticCache` matches a hand-written ORT static-KV baseline within 2%.

## Correctness

Token-identical output vs `DynamicCache` (greedy, 256 new tokens, crossing the 2048/4096 bucket boundaries), on Phi-4-mini-instructand Qwen3.5-4B, Both paths are self-deterministic.

## Constraints

Enforced with explicit errors, `DynamicCache` remains the default everywhere:

- WebGPU only (the problem and the mechanism are WebGPU-specific).
- Decoder-only models, batch size 1.
- The exported graph must support past/present sharing one buffer (`past_present_share_buffer` semantics — true for GQA-based exports, which current WebGPU decoder exports use). This is why the feature is opt-in.

### DIFF
--- a/packages/transformers/src/cache_utils.js
+++ b/packages/transformers/src/cache_utils.js
@@ -1,4 +1,26 @@
+/// <reference types="@webgpu/types" />
 import { Tensor } from './utils/tensor.js';
+import { createGpuBufferTensor, isONNXProxy } from './backends/onnx.js';
+
+/**
+ * Map a cache output name to the corresponding cache input name.
+ * @param {string} name The output name.
+ * @returns {string} The matching input name.
+ */
+export function presentNameToPastName(name) {
+    return (
+        name
+            // Hybrid cache architecture
+            .replace('present_ssm', 'past_ssm') // Mamba
+            .replace('present_conv', 'past_conv') // LFM2
+            .replace('present_recurrent', 'past_recurrent') // Qwen3.5
+            .replace('present_compressor', 'past_compressor') // Deepseek V4
+            .replace('present_indexer', 'past_indexer') // Deepseek V4
+
+            // Standard cache architecture
+            .replace('present', 'past_key_values')
+    );
+}
 
 /**
  * A cache class that stores past key values as named tensors.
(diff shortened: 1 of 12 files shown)


packages/transformers/src/models/modeling_utils.js
```diff
@@ -1037,9 +1037,12 @@ export class PreTrainedModel extends Callable {
...
         const past_key_values = getPastKeyValues(outputs, model_inputs.past_key_values);
 
-        // Dispose output tensors not held by the cache
+        // Dispose output tensors not held by the cache. StaticCache outputs are
+        // fresh wrappers aliasing the cache's buffers, so match them by name.
+        const is_static_cache = past_key_values instanceof StaticCache;
         const cachedTensors = new Set(Object.values(past_key_values));
-        for (const tensor of Object.values(outputs)) {
+        for (const [name, tensor] of Object.entries(outputs)) {
+            if (is_static_cache && name.startsWith('present')) continue;
```
nico-martin: These are fresh ORT tensor wrappers each run even when they alias the static GPU buffers. Skipping disposal leaves per-step runtime handles to GC. Please explicitly release alias wrappers after each run while preserving the caller-owned buffers, and add a long-generation resource test.

packages/transformers/src/models/modeling_utils.js
```diff
@@ -1350,11 +1367,29 @@ export async function decoder_forward(self, model_inputs, is_encoder_decoder = f
...
+ *
+ * @param {Object} session The decoder InferenceSession.
+ * @param {Record<string, Tensor>} inputs The session inputs.
+ * @param {DynamicCache|StaticCache|null} past_key_values The cache in use.
+ * @returns {Promise<Object>} The session outputs.
+ */
+export async function runDecoderSession(session, inputs, past_key_values) {
+    if (past_key_values instanceof StaticCache) {
+        const num_new_tokens = (inputs.input_ids ?? inputs.inputs_embeds).dims[1];
+        past_key_values._reserve(num_new_tokens);
```
nico-martin: This advances sequence length before `sessionRun()` succeeds. A rejected run leaves the cache logically ahead of its contents and breaks retries. Please commit length only after success or roll back on failure.

packages/transformers/src/cache_utils.js
```diff
@@ -80,3 +101,158 @@ class _DynamicCache {
...
+                `StaticCache is only supported on the 'webgpu' device (got '${session.config?.device}'). ` +
+                    'Use DynamicCache (the default) instead.',
+            );
+        }
+        if (isONNXProxy()) {
+            throw new Error('StaticCache is not supported when the ONNX backend is proxied.');
+        }
+
+        for (const meta of session.inputMetadata) {
+            if (!cacheInputNames.has(meta.name)) continue;
```
nico-martin: If a later allocation throws, earlier buffers/properties remain while `allocated` is false, so retry allocates over partial state. Please allocate into local structures, commit atomically, and destroy temporary buffers on failure.

nico-martin (changes requested):
Hi @kylo5aby, thank you so much for looking into this! The performance results are compelling and the overall approach makes sense. Before merging, we need to release per-step tensor wrappers, advance cache length only after successful inference, and make allocation transactional on failure. Could you also add focused lifecycle and WebGPU tests? I'm holding off on merging because the current failure paths can leak resources and corrupt cache state.

packages/transformers/src/models/modeling_utils.js
kylo5aby: > These are fresh ORT tensor wrappers each run even when they alias the static GPU buffers. Skipping disposal leaves per-step runtime handles to GC. Please explicitly release alias wrappers after each run while preserving the caller-owned buffers, and add a long-generation resource test.

Fixed, `sessionRun()` now reuses the caller's wrappers for pre-allocated outputs, so no per-step alias wrappers exist to leak, and the name-based dispose skip is gone. Covered by the `sessionRun` identity tests and a 300-step long-generation resource test.

packages/transformers/src/models/modeling_utils.js
kylo5aby: > This advances sequence length before `sessionRun()` succeeds. A rejected run leaves the cache logically ahead of its contents and breaks retries. Please commit length only after success or roll back on failure.

Fixed, `_reserve()` is split into ` _checkCapacity()`  (before the run) and _commit() (only after the run succeeded and an in-place-write assertion passed). A failed run leaves the cache length untouched and retryable.

packages/transformers/src/cache_utils.js
kylo5aby: > If a later allocation throws, earlier buffers/properties remain while `allocated` is false, so retry allocates over partial state. Please allocate into local structures, commit atomically, and destroy temporary buffers on failure.

Fixed,  allocation now goes into local structures and is committed atomically at the end; on failure all created wrappers are disposed and temporary GPU buffers destroyed before rethrowing, so the cache returns to a clean unallocated state and retry works

kylo5aby:
> Hi @kylo5aby, thank you so much for looking into this! The performance results are compelling and the overall approach makes sense. Before merging, we need to release per-step tensor wrappers, advance cache length only after successful inference, and make allocation transactional on failure. Could you also add focused lifecycle and WebGPU tests? I'm holding off on merging because the current failure paths can leak resources and corrupt cache state.

Hi @nico-martin, thanks for the thorough review, all three issues were  fixed, plus a few related hardening changes and a dedicated test suite. Added ` tests/cache_utils.test.js` : state-machine lifecycle, transactional allocation with a mocked WebGPU device, ` generate()` integration on a tiny model, `sessionRun ` fetches identity, and the long-generation resource test (300 decode steps: stable wrapper identity, transient outputs disposed, zero GPU buffer churn).

nico-martin (changes requested):
Thanks for the updates! The wrapper cleanup and cache commit ordering look good now, and all 23 focused tests pass locally. There’s just one allocation edge case left: if `createBuffer()` succeeds but `Tensor.fromGpuBuffer()` throws, that new buffer never reaches the rollback logic and leaks. Could you destroy it inside `createGpuBufferTensor()` when wrapping fails and add a test for that path?

kylo5aby:
> Thanks for the updates! The wrapper cleanup and cache commit ordering look good now, and all 23 focused tests pass locally. There’s just one allocation edge case left: if `createBuffer()` succeeds but `Tensor.fromGpuBuffer()` throws, that new buffer never reaches the rollback logic and leaks. Could you destroy it inside `createGpuBufferTensor()` when wrapping fails and add a test for that path?

good catch, that buffer was indeed unreachable by `_allocate's ` rollback since it had not been handed back yet.  `createGpuBufferTensor()` now destroys the buffer and rethrows when ` Tensor.fromGpuBuffer()` fails, and there's a new test covering exactly that path. Also added a "static KV cache" section to the WebGPU guide with usage.

nico-martin (approved):
Thanks for the updates! LGTM

kylo5aby:
resolved conflict.

kylo5aby:
@nico-martin hi, anything blocking it, or could you help merge when you get a chance? thanks.

PR #1747, opened 2026-08-14
Load external data without copying it onto the JS heap

astefanutti:
## Description

Loading a model's external data currently routes every byte through the JS heap. On a cache hit, `loadResourceFile` reads the `Response` into a `Uint8Array` and `getModelDataFiles` passes that buffer to onnxruntime-web, which allocates its own — so a multi-gigabyte model is resident twice at the same moment. On a cache miss the same buffer is built in order to report download progress. And `getModelDataFiles` starts every chunk **concurrently**, so the peak is the sum of the chunks, not the largest one.

For a model with eight ~2 GB chunks that is ~16 GB of `Uint8Array` before the runtime has allocated anything, and it fails.

A Cache Storage `Response.blob()` is a file reference rather than a copy, and onnxruntime-web accepts one for external data as of [#29477](https://github.com/microsoft/onnxruntime/pull/29477) (merged 2026-07-14). `cache.put()` likewise takes a `Response` and streams it to disk without materialising it. Between them the bytes can go network → disk → runtime and never touch the heap.

### Measured

**Warm load** — same model, same runtime, same session, peak RSS sampled at 200 ms, external data passed as `Uint8Array` vs `Blob`, on a 2.887 GB model:

| | peak RSS | × the weights |
|---|---|---|
| before | 6,056 MB | 2.10× |
| after | **4,221 MB** | 1.46× |

The 1,835 MB saved is one full copy of the weights, so it scales with model size. (Caveats: one session rather than the two an app may load concurrently, wasm EP so the GPU's own residency is not counted, and summed RSS double-counts pages shared between processes. The ratio is the finding; the absolutes are indicative.)

**Cold load** — a 17 GB model (8 external-data files, largest 1.996 GB), empty cache, first attempt:

| | before | after |
|---|---|---|
| `Array buffer allocation failed` | 2 | **0** |
| downloaded | stalled at 16.2 / 17.0 GB | **17.0 / 17.0 GB** |
| outcome | 13 GB cached, load failed, needed a retry | **loaded and generated** |

Before this, a first load of a model that size fails and only completes on a later attempt, because each file is cached as it completes so the retry has less left to buffer. Progress reporting still works throughout (1.6 → 4.2 → 6.8 → 9.2 → 11.8 → 14.3 → 16.1 → 17.0 GB).

That model has also run three full conversation suites through the new path — 84 generated answers, no failures.

### Design notes

**Progress is why the buffer existed**, so the cold path keeps it with a pass-through `TransformStream` that counts bytes as they go past. It holds one chunk, not the file, and backpressure keeps it that way.

**It is opt-in**, threaded from the one caller that knows it is looking at external data. `loadResourceFile` is generic — it also serves `config.json` and `tokenizer.json`, whose callers parse the result as text — so returning a `Blob` unconditionally would break them. Hence the `as_blob` parameter rather than a change in behaviour. Node is unaffected and keeps returning a path.

**If the cache refuses the write** (`QuotaExceededError` being the expected one) it falls back to the buffered path. That has to re-fetch rather than reuse the response, whose body the failed attempt consumed; the re-fetch goes through `getFile()` so a gated repo keeps its `Authorization` header.

**The onnxruntime-web bump is load-bearing, not incidental.** The current pin (2026-04-16) predates #29477, so the runtime materialises any Blob handed to it — which is the peak this removes, and is also why it throws over the 2 GiB `ArrayBuffer` limit. The new pin is dated after that merge and is what the numbers above were measured against. The package already pins a dev build, so this is a bump rather than a new kind of dependency.

### Not covered

`getModelDataFiles` still starts every chunk concurrently. That no longer costs heap, but it does mean N simultaneous connections; serialising them is a separate question and not addressed here.

### DIFF
--- a/packages/transformers/package.json
+++ b/packages/transformers/package.json
@@ -58,7 +58,7 @@
     "@huggingface/jinja": "^0.5.6",
     "@huggingface/tokenizers": "^0.1.3",
     "onnxruntime-node": "1.24.3",
-    "onnxruntime-web": "1.26.0-dev.20260416-b7804b056c",
+    "onnxruntime-web": "1.29.0-dev.20260811-e415ef9afd",
     "sharp": "^0.34.5"
   },
   "devDependencies": {
--- a/packages/transformers/src/utils/hub.js
+++ b/packages/transformers/src/utils/hub.js
@@ -245,9 +245,11 @@ export async function storeCachedResource(path_or_repo_id, filename, cache, cach
  * @param {PretrainedOptions} [options] An object containing optional parameters.
  * @param {boolean} [return_path=false] Whether to return the path of the file instead of the file content.
  * @param {import('./cache.js').CacheInterface | null} [cache] The cache instance to use.
+ * @param {boolean} [as_blob=false] Whether to return a `Blob` when the file is served from the cache,
+ * rather than reading it into a `Uint8Array`. Only honoured on a cache hit — see the note at the use site.
  *
  * @throws Will throw an error if the file is not found and `fatal` is true.
- * @returns {Promise<string|Uint8Array|null>} A Promise that resolves with the file content as a Uint8Array if `return_path` is false, or the file path as a string if `return_path` is true.
+ * @returns {Promise<string|Uint8Array|Blob|null>} A Promise that resolves with the file content as a Uint8Array if `return_path` is false, or the file path as a string if `return_path` is true. Resolves with a `Blob` when `as_blob` is set and the file came from the cache.
  */
 export async function loadResourceFile(
     path_or_repo_id,
@@ -256,6 +258,7 @@ export async function loadResourceFile(
     options = {},
     return_path = false,
     cache = null,
+    as_blob = false,
 ) {
     const { requestURL, localPath, remoteURL, proposedCacheKey, validModelId } = buildResourcePaths(
         path_or_repo_id,
@@ -366,7 +369,26 @@ export async function loadResourceFile(
         let buffer;
 
         if (typeof response !== 'string') {
-            if (!options.progress_callback) {
+            if (as_blob && cacheHit) {
+                // The whole point of `as_blob`, and it is one call: a Cache Storage `Response` hands back a
+                // Blob that is a FILE REFERENCE rather than a copy, so the bytes never enter the JS heap.
+                //
+                // Gated on `cacheHit`, which is a deliberate trade rather than caution. Reading the stream to
+                // report progress would defeat the whole thing — the chunks would be resident twice, once as
+                // buffers and once in Blob storage — and on a cold download progress is worth more than peak
+                // memory, because the user is watching several gigabytes arrive. Every load AFTER the first
+                // takes this branch, is instantaneous, and is where the peak actually matters.
+                buffer = /** @type {any} */ (await response.blob());
+
+                dispatchCallback(options.progress_callback, {
+                    status: 'progress',
+                    name: path_or_repo_id,
+                    file: filename,
+                    progress: 100,
+                    loaded: buffer.size,
+                    total: buffer.size,
+                });
+            } else if (!options.progress_callback) {
                 // If no progress callback is specified, we can use the `.arrayBuffer()`
                 // method to read the response.
                 buffer = new Uint8Array(await response.arrayBuffer());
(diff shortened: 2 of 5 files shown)


packages/transformers/package.json
```diff
@@ -58,7 +58,7 @@
     "@huggingface/jinja": "^0.5.6",
     "@huggingface/tokenizers": "^0.1.3",
     "onnxruntime-node": "1.24.3",
-    "onnxruntime-web": "1.26.0-dev.20260416-b7804b056c",
+    "onnxruntime-web": "1.29.0-dev.20260811-e415ef9afd",
```
nico-martin, line 61: Please regenerate `pnpm-lock.yaml`. The lockfile still resolves the old runtime, so frozen installs fail and validation may not exercise this load-bearing change.

packages/transformers/src/utils/model-loader.js
```diff
@@ -91,8 +91,11 @@ export async function getModelDataFiles(
                         true,
                         options,
                         return_path,
+                        // In the browser, hand onnxruntime-web a Blob rather than a materialised buffer.
+                        // Node keeps returning a path, which is cheaper still.
+                        !return_path,
```
nico-martin, line 105: Upstream on-demand Blob loading is JSPI-only, while Transformers.js defaults to asyncify, whose fallback fully materializes the Blob. Please select/gate a supported JSPI path or clearly separate and test the asyncify behavior from the claimed memory result.

packages/transformers/src/utils/hub.js
```diff
@@ -245,9 +245,11 @@ export async function storeCachedResource(path_or_repo_id, filename, cache, cach
  * @param {PretrainedOptions} [options] An object containing optional parameters.
  * @param {boolean} [return_path=false] Whether to return the path of the file instead of the file content.
  * @param {import('./cache.js').CacheInterface | null} [cache] The cache instance to use.
+ * @param {boolean} [as_blob=false] Whether to return a `Blob` when the file is served from the cache,
```
nico-martin: This is no longer cache-hit-only: the cold path streams to cache and returns the new Blob in the same call. Please update the parameter and return documentation.

packages/transformers/src/utils/hub.js
```diff
@@ -494,11 +569,20 @@ const INFLIGHT_LOADS = new Map();
...
+ * @returns {Promise<string|Uint8Array|Blob>} A Promise that resolves with the file content as a Uint8Array if `return_path` is false, or the file path as a string if `return_path` is true. Resolves with a `Blob` when `as_blob` is set and the file came from the cache.
  */
-export async function getModelFile(path_or_repo_id, filename, fatal = true, options = {}, return_path = false) {
+export async function getModelFile(
+    path_or_repo_id,
+    filename,
+    fatal = true,
+    options = {},
+    return_path = false,
+    as_blob = false,
```
nico-martin, line 607: `as_blob` changes the promise result type and must be part of this key. Mixed concurrent callers currently share whichever representation was requested first.

nico-martin (commented):
Hi @astefanutti, thank you so much for looking into this! I like the streaming direction, but the memory benefit is JSPI-only while the default runtime is still asyncify. The ORT lockfile also needs updating, and `as_blob` should be part of the in-flight key. Could you address those points and add focused warm/cold cache tests? I'm holding off on merging because the default path does not yet match the memory claim and concurrent callers can receive the wrong type.

astefanutti:
Thanks a lot @nico-martin for the review. I've addressed your comments as extra commits rather than a rewrite so the review stays readable:

**Lockfile** (`96281e5`). The ORT bump left `pnpm-lock.yaml` pinning `1.26.0-dev.20260416`, so `pnpm install --frozen-lockfile` was dying eight seconds in on Node 18, 20 and 22. Regenerated with `--lockfile-only`, so the change is the resolution and nothing else.

**`as_blob` in the in-flight key** (`60e92d5`). Agreed, and I think the sharp version of it is that `as_blob` changes the *resolved type*, not just how the bytes are fetched — so two callers passing different values aren't asking the same question, which is why `return_path` is in the key already. Two of the new tests cover both orderings of a concurrent Blob/bytes pair; I confirmed they fail with the key change reverted and pass with it.

**Warm/cold tests** (`60e92d5`). Six, against an in-memory cache and a stubbed `env.fetch`, so they touch no network and assert which branch ran: cold streams into the cache, stores exactly once and still reports progress up to the total; warm resolves from the cache without re-fetching; a cache that refuses the write falls back to a buffer rather than failing the load; and `as_blob` unset still returns a `Uint8Array`.

**The memory claim vs the default runtime** (`a731c00b`). You're right that they don't match, and I think the PR was overstating rather than wrong — there are two savings here and my comments ran them together:

- the **download** no longer buffers. That happens before onnxruntime-web sees anything and holds whichever build is in use. It's the half that stops a cold load of a model with several multi-gigabyte chunks dying in `getModelDataFiles`, where every chunk starts concurrently and the peak is their sum.
- the **session** may still copy. Only the JSPI build reads external data from a Blob without materialising it; the default asyncify build copies it in.

So "the bytes never sit on the JS heap at all" was true of the download and overstated end to end. The comments now say which is which, and say plainly that on the default build this buys the download and not the session. No behaviour change. If you'd prefer the PR claim only the download half, I'm happy to retitle it accordingly.

**One more** (`a04faae2`). `tsc --build` has been failing on this branch since the Blob was introduced — five errors, invisible because CI never got past `pnpm install`. Fixing the lockfile on its own would have moved the red from install to build rather than clearing it. So: `buffer` widened to `Uint8Array|Blob` rather than cast away at each use, the `cache.match` result cast to `Response` instead of a blanket `any`, `INFLIGHT_LOADS` widened, and two places narrowed *because a Blob genuinely cannot reach them* — a warm read is a cache hit and never sets `toCacheResponse`, a cold stream writes the entry itself and then clears it — with that reasoning written down so a future branch which breaks the assumption fails at build.

packages/transformers/package.json
astefanutti: Fixed in 96281e5ead5ef6a69e2c0d6904f33fc79a404553

packages/transformers/src/utils/hub.js
astefanutti: Addressed in 60e92d5b99cbd6ffcaa790dc4542843176310a7e.

packages/transformers/src/utils/hub.js
astefanutti: Addressed in face029db94d67de5428fefa27c346b84f8f046c.

nico-martin (approved):
Thanks for the thorough follow-up! LGTM!

PR #1755, opened 2026-08-22
Fix Whisper WebGPU memory leak during generation

### LINKED ISSUE #1739: Whisper ASR pipeline leaks ~650 MB of GPU memory per 30 s chunk on WebGPU (both `return_timestamps` modes)
### System Info

```
@huggingface/transformers  4.2.0
onnxruntime-web            1.26.0 (as pinned)
Browser                    Chrome 150, macOS 15, Apple silicon (24 GB unified memory)
Also observed on           Chrome 150/151, Windows 11 (Iris Xe) and Linux
                           (RTX 3060 12 GB, driver 580, Vulkan) — see downstream report
Model                      onnx-community/whisper-small_timestamped
device                     "webgpu", dtype {encoder_model: "fp32", decoder_model_merged: "q4"}
```

### Description

Transcribing long audio through the ASR pipeline grows GPU memory monotonically — roughly **650 MB per 30-second chunk** with `whisper-small` — and it is only reclaimed when the page is closed or reloaded. Sampling Chrome's GPU-process memory footprint every 3 s during the minimal repro below: ~1.6 GB after pipeline init → **~8.1 GB after transcribing 300 s of audio** (10 chunks, 134 s wall clock), still ~7.8 GB at rest afterwards. On a 12 GB RTX 3060 a longer file dies mid-run with `failed to call OrtRun ... [Invalid Buffer]` (downstream: hyperaudio/hyperaudio-lite-editor#462).

`return_timestamps: "word"` and `return_timestamps: true` leak at the same rate, so this is not specific to the token-timestamps/DTW path.

### Reproduction

Self-contained page — the leak occurs inside the single `pipe()` call:

```html
<script type="module">
const { pipeline } = await import('https://cdn.jsdelivr.net/npm/@huggingface/transformers@4.2.0');
const buf = await fetch('./audio.mp4').then(r => r.arrayBuffer());
const ctx = new AudioContext({ sampleRate: 16000 });
const audio = (await ctx.decodeAudioData(buf)).getChannelData(0).slice(0, 300 * 16000);
const pipe = await pipeline('automatic-speech-recognition', 'onnx-community/whisper-small_timestamped', {
  device: 'webgpu',
  dtype: { encoder_model: 'fp32', decoder_model_merged: 'q4' },
});
await pipe(audio, { return_timestamps: 'word', chunk_length_s: 30, stride_length_s: 5, language: 'en' });
// GPU process: ~1.6 GB before this call, ~8 GB after — watch it climb per chunk
</script>
```

(Any ~5-minute speech recording reproduces it. On macOS, watch the Chrome GPU process's *memory footprint* — plain RSS does not show these allocations.)

### Relation to #1734

Looks like the same family — GPU buffers from generation never disposed — but the `keepCacheAlive` mechanism there only covers `"word"` mode here (chunk mode takes the non-dict path, whose `past_key_values.dispose()` does run) and both modes leak equally, so there appears to be an additional mechanism. PR #1737 fixes Chatterbox's wrapper only.

### Possibly-relevant observations from reading 4.2.0 (unverified)

- In the seq2seq path, `encoder_outputs` is created per `generate()` call and never disposed in either mode.
- In `getPastKeyValues` (`modeling_utils.js`), on the `is_encoder_pkv && pastKeyValues` reuse path, each step's freshly returned `decoderResults[name]` buffers are dropped in favour of the cached ones without being disposed.

Happy to run further instrumented tests on macOS/Chrome if useful.

ahsan722505:
## What does this PR do?

Fixes #1739.

- Disposes the duplicate constant encoder KV tensors returned by the decoder when the existing Optimum-style encoder cache is reused.
- Disposes encoder output created internally by encoder-decoder generation once generation completes.
- Disposes Whisper attention and cache tensors after token timestamps have been extracted.
- Adds regression coverage for both duplicate and already-reused encoder cache tensors.

## Validation

- `pnpm --filter @huggingface/transformers test -- ./tests/utils/generation_cache.test.js`
- `pnpm --filter @huggingface/transformers test -t "WhisperForConditionalGeneration"`
- `pnpm --filter @huggingface/transformers typegen`
- Prettier check on all changed files

A downstream Electron/WebGPU smoke test using `onnx-community/whisper-small_timestamped`, an fp32 encoder, a q4 decoder, 30-second chunks, 5-second stride, and word timestamps completed 300 seconds of audio with renderer memory staying around 1.5-1.9 GB instead of growing per chunk.

### DIFF
--- a/packages/transformers/src/models/modeling_utils.js
+++ b/packages/transformers/src/models/modeling_utils.js
@@ -856,6 +856,7 @@ export class PreTrainedModel extends Callable {
         });
 
         const is_encoder_decoder = this.config.is_encoder_decoder;
+        let encoderOutputsToDispose = null;
 
         // 4. Define other model kwargs
         if (!is_encoder_decoder) {
@@ -869,6 +870,7 @@ export class PreTrainedModel extends Callable {
                 model_input_name,
                 generation_config,
             });
+            encoderOutputsToDispose = model_inputs.encoder_outputs;
         }
 
         // 5. Prepare `input_ids` which will be used for auto-regressive generation
@@ -1051,6 +1053,8 @@ export class PreTrainedModel extends Callable {
             await past_key_values.dispose();
         }
 
+        encoderOutputsToDispose?.dispose();
+
         if (generation_config.return_dict_in_generate) {
             return {
                 sequences,
@@ -1199,6 +1203,9 @@ export function getPastKeyValues(decoderResults, pastKeyValues) {
                 // Optimization introduced by optimum to reuse past key values.
                 // So, we just replace the constant outputs (`decoderResults[name]`) with the previous past key values.
                 // https://github.com/huggingface/optimum/blob/0bf2c05fb7e1182b52d21b703cfc95fd9e4ea3dc/optimum/onnxruntime/base.py#L677-L704
+                if (decoderResults[name] !== pastKeyValues[newName]) {
+                    decoderResults[name].dispose();
+                }
                 pkvs[newName] = pastKeyValues[newName];
             } else {
                 pkvs[newName] = decoderResults[name];
--- a/packages/transformers/src/models/whisper/modeling_whisper.js
+++ b/packages/transformers/src/models/whisper/modeling_whisper.js
@@ -13,6 +13,24 @@ import { mergeArrays } from '../../utils/core.js';
 import { ModelOutput } from '../modeling_outputs.js';
 import { logger } from '../../utils/logger.js';
 
+async function disposeTimestampGenerationOutputs(outputs) {
+    const disposed = new Set();
+    for (const name of ['cross_attentions', 'encoder_attentions', 'decoder_attentions']) {
+        const tokenAttentions = outputs[name];
+        if (!tokenAttentions) continue;
+
+        for (const attentions of tokenAttentions) {
+            for (const tensor of attentions) {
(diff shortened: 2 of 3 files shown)


maboa:
Tested this on an M4 Pro (24 GB, unified memory), Chrome 153, WebGPU, running `onnx-community/whisper-base.en_timestamped` with `encoder_model: fp32`, `decoder_model_merged: q4`, `return_timestamps: 'word'`, `chunk_length_s: 30`, `stride_length_s: 5`, three times in a row over the same 137 s recording. I compared this branch against its merge base (353007b), measuring the physical footprint of Chrome's GPU process:

- **merge base:** grows about 600 MB per run (2.03 → 2.63 → 3.22 GB, and 2.95 → 3.55 → 4.11 GB on a second pass)
- **this PR:** flat (1.50 → 1.59 → 1.59 GB, and 4.12 → 4.20 → 4.12 GB)

Words and timestamps are byte-identical between the two builds, and the run time is the same.

For context: on Apple silicon the leak doesn't end in a GPU out-of-memory error, as it does on a discrete card. With `whisper-small.en_timestamped` on a 10-minute file, Chrome's memory reached about 56 GB and the whole machine locked up, needing a forced restart. We've withdrawn the larger Whisper models from our editor until this lands. Thanks for the fix. It would be great to see it reviewed.

PR #1757, opened 2026-08-24
Pin WebGPU KV cache for ImageAudioTextToText models (Gemma3n, Gemma4)

shoemoney:
`[MODEL_TYPES.ImageAudioTextToText]` (src/models/session_config.js:77 on main) is the only generating model type in `MODEL_SESSION_CONFIG` with no `cache_sessions` declaration. `constructSessions` reads `cache_sessions?.[name] ?? false` (src/models/session.js:149), so `cache_config` is `false` for `decoder_model_merged`, and the `if (cache_config && selectedDevice === 'webgpu')` gate in `getSession` (src/models/session.js:110) never sets `preferredOutputLocation: 'gpu-buffer'` for the `present.*` outputs.

The result: on WebGPU, every decode step of `Gemma3nForConditionalGeneration` and `Gemma4ForConditionalGeneration` (src/models/registry.js:627-636) downloads the full KV cache to the CPU and re-uploads it as `past_key_values`, with cost growing per token. The same weights loaded text-only keep the cache on-GPU, since `Gemma3nForCausalLM`/`Gemma4ForCausalLM` route through `MODEL_TYPES.DecoderOnly`, which declares `cache_sessions: { model: true }` (session_config.js:24). The pinning machinery even has a dedicated `gemma4`/`gemma4_text` branch in `getCacheNames` (src/configs.js:403-415) that the multimodal path never reached.

Fix: add `cache_sessions: { decoder_model_merged: true }` to the `ImageAudioTextToText` entry, matching the `AudioTextToText` (session_config.js:74) and `ImageTextToText` (session_config.js:64) siblings exactly.

Also adds `tests/utils/session_config.test.js`: a regression test that the `ImageAudioTextToText` decoder is cache-pinned, plus an invariant that every model type declaring a `generation_config` also declares at least one cache session, so the next model type added cannot silently repeat this. Without the source change both tests fail (`missing: ["ImageAudioTextToText"]`); with it, the affected slice is green (17 suites, 294 passed, 1 skipped). `pnpm format:check` and `tsc --build` pass.

One thing deliberately not changed: no encoder pinning. `VoxtralRealtime` pins its `audio_encoder` (session_config.js:137) because streaming re-invokes the encoder; Gemma3n/Gemma4 run their vision and audio encoders once per input, so only the decoder KV cache benefits.

### DIFF
--- a/packages/transformers/tests/utils/session_config.test.js
+++ b/packages/transformers/tests/utils/session_config.test.js
@@ -1,0 +1,22 @@
+import { MODEL_TYPES, MODEL_SESSION_CONFIG, getSessionsConfig } from "../../src/models/session_config.js";
+
+describe("MODEL_SESSION_CONFIG", () => {
+  it("should pin the KV cache of the ImageAudioTextToText decoder (regression: missing cache_sessions)", () => {
+    const { cache_sessions } = getSessionsConfig(MODEL_TYPES.ImageAudioTextToText, {});
+    expect(cache_sessions?.decoder_model_merged).toBe(true);
+  });
+
+  it("should declare cache_sessions for every generating model type", () => {
+    // Every model type that loads a generation config runs an autoregressive decode loop,
+    // so at least one of its sessions must have its KV-cache outputs marked as cacheable
+    // (used on WebGPU to keep `present.*` outputs on-GPU via preferredOutputLocation).
+    const missing = [];
+    for (const [name, type] of Object.entries(MODEL_TYPES)) {
+      const typeConfig = MODEL_SESSION_CONFIG[type];
+      if (!typeConfig?.optional_configs?.generation_config) continue;
+      const hasCacheSession = Object.values(typeConfig.cache_sessions ?? {}).some(Boolean);
+      if (!hasCacheSession) missing.push(name);
+    }
+    expect(missing).toEqual([]);
+  });
+});
--- a/packages/transformers/src/models/session_config.js
+++ b/packages/transformers/src/models/session_config.js
@@ -84,6 +84,7 @@ export const MODEL_SESSION_CONFIG = {
             }
             return s;
         },
+        cache_sessions: { decoder_model_merged: true },
         optional_configs: { generation_config: 'generation_config.json' },
     },
     [MODEL_TYPES.Phi3V]: {


packages/transformers/tests/utils/session_config.test.js
```diff
@@ -0,0 +1,22 @@
+import { MODEL_TYPES, MODEL_SESSION_CONFIG, getSessionsConfig } from "../../src/models/session_config.js";
+
+describe("MODEL_SESSION_CONFIG", () => {
+  it("should pin the KV cache of the ImageAudioTextToText decoder (regression: missing cache_sessions)", () => {
+    const { cache_sessions } = getSessionsConfig(MODEL_TYPES.ImageAudioTextToText, {});
+    expect(cache_sessions?.decoder_model_merged).toBe(true);
+  });
+
+  it("should declare cache_sessions for every generating model type", () => {
```
nico-martin, line 9: Can we confirm that “has generation config” is an intentional project-wide contract meaning “must expose a cacheable ONNX session”? If this guard is specifically for decoder/KV-cache architectures, narrowing or documenting the predicate would avoid misleading failures for a future cacheless generator.

nico-martin (approved):
Hi @shoemoney, thank you so much for looking into this! The config change looks right and the focused regression test is good. My only question is whether the broader test should assume every generating model must have a cacheable session; that may be too broad for future architectures.

shoemoney:
Thanks for the review. On the broad test: fair concern. It encodes what is true of every current generating type after this fix, and its purpose is to be a tripwire so the next model type does not silently ship without cache pinning the way ImageAudioTextToText did. If a future architecture legitimately has no cacheable session, the test fails loudly at the config level and the fix is a one line exemption. That said, I am happy to add an explicit exemptions list now, or drop the broad invariant and keep only the focused regression test if you prefer. Your call.

On CI: the build (22) failure is the musicgen batch_size tests returning garbage magnitudes (5.5e24 vs 0.1664) on CPU. This PR only adds a cache_sessions flag for ImageAudioTextToText, which is consulted only for WebGPU preferredOutputLocation, so it cannot reach the musicgen path. Other PRs today pass the same suite, so it looks like a flaky run. A re-run should clear it.

nico-martin:
@shoemoney Thanks, that rationale makes sense. I’m fine keeping the broad invariant as a deliberate tripwire. No need to add an exemptions list until there is a legitimate exception. The MusicGen failure also looks unrelated given that this path is WebGPU-only.

packages/transformers/tests/utils/session_config.test.js
xenova, line 1: nice test!

xenova (approved):
thanks! clean PR. honestly not sure how this was missed before!

PR #1758, opened 2026-08-24
Add dependency-free response constraints

nico-martin:
Adds a small, dependency-free package for JSON Schema, JSON object, and regex constrained generation. It also hooks token sampling into the generation loop and includes tests and docs for the new API.

```js
import { ResponseConstraint } from "@huggingface/transformers-response-constraint";

const constraint = ResponseConstraint.fromResponseFormat(tokenizer, {
  type: "json_schema",
  json_schema: {
    type: "object",
    properties: {
      answer: { type: "string" },
    },
    required: ["answer"],
    additionalProperties: false,
  },
});

await model.generate({
  ...inputs,
  logits_processor: constraint.logits_processor,
  stopping_criteria: constraint.stopping_criteria,
});
```


packages/transformers/src/models/modeling_utils.js
```diff
@@ -999,17 +999,18 @@ export class PreTrainedModel extends Callable {
 
                 const sampledTokens = await sampler(logs);
                 for (const [newTokenId, logProb] of sampledTokens) {
-                    const bigint = BigInt(newTokenId);
                     // TODO: If branching, use previous beam as a starting point
                     // update generated ids, model inputs, and length for next step
                     scores[batch_idx] += logProb;
-                    all_input_ids[batch_idx].push(bigint);
-                    generated_input_ids.push([bigint]);
+                    generated_input_ids.push([newTokenId]);
```
xenova: is `BigInt(newTokenId)` no longer needed? 👀 may be a while since I visited this code snippet, but are sampled tokens `number` or `bigint`?

packages/transformers/src/models/modeling_utils.js
```diff
@@ -999,17 +999,18 @@ export class PreTrainedModel extends Callable {
...
 
                     // TODO: Support beam search
                     break;
                 }
             }
+            for (let batch_idx = 0; batch_idx < generated_input_ids.length; ++batch_idx) {
+                all_input_ids[batch_idx].push(generated_input_ids[batch_idx][0]);
+            }
```
xenova: reason for doing this afterwards instead?

packages/transformers/src/transformers.js
```diff
@@ -52,6 +52,7 @@ export { load_video, RawVideo, RawVideoFrame } from './utils/video.js';
 export * from './utils/tensor.js';
 export { softmax, log_softmax, dot, cos_sim } from './utils/maths.js';
 export { random } from './utils/random.js';
+export { logger } from './utils/logger.js';
```
xenova: do we need to export this?

packages/transformers-structured-output/src/engine/json.ts
```diff
@@ -0,0 +1,1853 @@
...
+        if (authority.indexOf(':') !== colon || !/^\d*$/.test(authority.slice(colon + 1))) return false;
+        host = authority.slice(0, colon);
+    }
+    return uriRegName.test(host);
+}
+
+function checkSchema(schema: Schema, path: string): void {
+    if (typeof schema === 'boolean') return;
+    if (!isRecord(schema)) throw new TypeError(`${path} must be a boolean or object schema.`);
+    const supported = new Set([
```
xenova, line 1571: do you have a reference for where this constant list of strings comes from?

packages/transformers-response-constraint/README.md
```diff
@@ -0,0 +1,58 @@
...
+const constraint = ResponseConstraint.fromResponseFormat(tokenizer, {
+  type: "regex",
+  regex: "\\d{4}-\\d{2}-\\d{2}",
+});
+
+const output = await model.generate({
+  ...inputs,
+  logits_processor: constraint.logits_processor,
+  stopping_criteria: constraint.stopping_criteria,
+});
```
xenova: pretty slick API! 👍

packages/transformers-structured-output/package.json
```diff
@@ -0,0 +1,68 @@
...
+  ],
+  "author": "Hugging Face",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/huggingface/transformers.js/issues"
+  },
+  "homepage": "https://github.com/huggingface/transformers.js#readme",
+  "peerDependencies": {
+    "@huggingface/transformers": "^4.2.0"
+  },
```
xenova, line 45: can you just explain the benefits of specifying as a `peerDependencies`? Can anyone (or should anyone) be allowed to use the library without transformers.js? More asking out of curiosity. This is probably the best choice anyway.

xenova (commented):
very nice! Just a few comments.

Regarding package naming, do you maybe have a list of good package names we can compare `transformers-response-constraint` to?

nico-martin:
> Regarding package naming, do you maybe have a list of good package names we can compare transformers-response-constraint to?

I actually spent quite some time thinking aobut names.  Here is the list of names I played around with:

* `@huggingface/transformers-constraints`
* `@huggingface/transformers-constrained-generation`
* `@huggingface/transformers-structured-output`
* `@huggingface/transformers-response-constraints`
* `@huggingface/transformers-guidance`

`@huggingface/transformers-response-constraint` is long, but it best describes the package.

If it should be shorter, we could my preference is `@huggingface/transformers-constraints`, otherwise `@huggingface/transformers-constrained-generation` is the most explicit alternative. Happy to rename it to whichever you prefer.

packages/transformers/src/models/modeling_utils.js
nico-martin: They are already `bigint`: the sampler contract is `Promise<[bigint, number][]>`, and all sampler implementations return either `BigInt(argmax)` or IDs from an `int64` tensor. So the conversion was redundant, but this change is unrelated to the feature. I reverted it to keep the PR focused.

packages/transformers/src/models/modeling_utils.js
nico-martin: No remaining reason. This was left over from an earlier implementation while I was changing how sampled tokens reached the constraint. The current stopping criteria receives the updated `all_input_ids` either way, so I restored the original placement and removed the unrelated diff.

packages/transformers/src/transformers.js
nico-martin: No. That export was left over from an earlier implementation and is not used by this package. I removed it.

packages/transformers-response-constraint/README.md
nico-martin: Thanks!

packages/transformers-structured-output/package.json
nico-martin: The package should not be used without Transformers.js. It imports and extends `LogitsProcessor`, `LogitsProcessorList`, and `StoppingCriteria` at runtime. Making Transformers.js a peer dependency says that this package plugs into the caller's Transformers.js installation, lets the package manager check version compatibility, and avoids a second nested copy of Transformers.js. The workspace dev dependency supplies it for this package's build and tests. So I think keeping it as a peer dependency is the right choice.

packages/transformers-structured-output/src/engine/json.ts
nico-martin: It’s a hand-maintained allowlist based on the JSON Schema 2020-12 Core and Validation specifications. It includes the keywords this engine implements, annotations it can safely ignore, a few draft-07 compatibility keywords, and the custom x-guidance extension. The allowlist ensures unsupported constraints are rejected rather than silently ignored. I’ll add links to the relevant specifications in the code.
It basically grew with the implementation and wit the demo-cases I used with the package.

sroussey:
I personally like transformers-structured-output as it relates to the intent and why rather than the how.

Any name you use will be find, I am looking forward to this!

xenova:
I actually agree with @sroussey -- I like `@huggingface/transformers-structured-output` a bit more too.

It kind of explains what you will achieve with the library instead of the internals (more dev focused)

> Structured outputs are the goal (the desired schema-compliant result), while constrained generation is the underlying mechanical technique used to achieve it.

It also looks like docs typically refer to this name (and is probably what users wills search for when they try themselves).
- https://developers.openai.com/api/docs/guides/structured-outputs
- https://platform.claude.com/docs/en/build-with-claude/structured-outputs
- https://ai.google.dev/gemini-api/docs/structured-output

astefanutti:
That's a great feature, thanks!

I've tested it for some of my use cases and I can see two additions that could be useful:

## 1. `allowedSpecialTokenIds`

`createTokenConstraint` refuses every id the tokenizer reports as special, in both places it can:

```ts
// engine/constraint.ts, in the trie walk that fills the mask
for (const tokenId of node.tokenIds) {
    if (tokenizer.data.specialTokenIds.has(tokenId)) continue;
    ...
// engine/constraint.ts, in commit()
if (tokenizer.data.specialTokenIds.has(tokenId)) {
    throw new Error(`Token ${tokenId} does not satisfy the constraint.`);
}
```

For JSON and prose that is exactly right — a schema-constrained answer should not be able to emit `<pad>` or `<bos>`, and an accidental EOS mid-object would be a bug.

An interesting use of a *regex* constraint on a chat model is a grammar over the model's **own wire format**: force a well-formed tool call, force a channel to open and close, force a turn to end where it should. Every marker in such a grammar is an added token, and every added token is `special: true`. So the constraint that most needs to name these tokens is the one that categorically cannot.

### Repro — no model, ~12 lines

```js
import { createTokenConstraint } from './src/engine';

const enc = new TextEncoder();
const strings = ['<eos>', '<|tool_call>', '<', '|', 'tool', '_', 'call', '>', 'a'];
const tokenizer = { tokens: strings.map((s) => enc.encode(s)), eosTokenId: 0, specialTokenIds: [0, 1] };
const format = { type: 'regex', regex: '<\\|tool_call>a' };

const mask = new Uint32Array(1);
const bits = (c) => (c.fillMask(mask), strings.filter((_, i) => mask[0] & (1 << i)));

bits(createTokenConstraint(tokenizer, format));
// → [ '<' ]                     the grammar demands a marker the model may not emit as itself

bits(createTokenConstraint(tokenizer, format, { allowedSpecialTokenIds: [1] }));
// → [ '<|tool_call>', '<' ]     with the patch, both paths are open and the sampler chooses
```

### The change

Two conditions, plus an option to carry the set:

```diff
-    if (tokenizer.data.specialTokenIds.has(tokenId)) continue;
+    if (tokenizer.data.specialTokenIds.has(tokenId) && !allowedSpecialTokenIds.has(tokenId)) continue;
```

## 2. `terminatorTokenIds`

Once the grammar is satisfied, the mask exposes exactly one terminator:

```ts
if (machine.accepting(state)) {
    setBit(target, tokenizer.data.eosTokenId);
```

That is a completion model's stop. A chat model's is its own end-of-turn marker — `<turn|>` on Gemma, `<|im_end|>` on the ChatML family, `<|eot_id|>` on Llama 3 — and `eos_token_id` in the tokenizer config is frequently *not* that id. When the two differ, a satisfied constraint puts the model in front of a choice between a token it does not emit in that position and no stop at all, and the observable result is generation running to `max_new_tokens`.

The change makes the terminator a caller-supplied **set**, defaulting to `[eos_token_id]` so nothing moves unless asked:

```js
ResponseConstraint.fromResponseFormat(tokenizer, format, {
  terminatorTokenIds: [endOfTurnId, tokenizer.eos_token_id],
});
```

A set rather than a single id for a practical reason: a caller who is not sure which stop a given checkpoint prefers can offer both and let the sampler answer, which is a much cheaper experiment than guessing and finding out from a runaway. An empty set is rejected at construction — it is not "the default", it is a constraint that can never terminate, and that failure mode is invisible until the token budget runs out.

These two options interact and the order matters: a terminator is very often a special token itself, so `commit()` has to test the terminator set **before** the special-token ban, or naming an end-of-turn marker as a terminator would be refused by the very check the first patch exists to relax.

nico-martin:
@astefanutti Thanks for testing this and for the clear reproductions. I like both oadditions!

I think the API direction makes sense, but the implementation needs to account for a few details beyond the two checks:

- The special-token and terminator sets need to be part of the mask cache key so constraints with different options don’t accidentally share masks.
- Terminators should be excluded from normal trie traversal, even when they aren’t special tokens, and only exposed once the constraint is accepting.
- As you mentioned, `commit()` should check terminators before applying the special-token restriction, while still rejecting them before the constraint is accepting.
- We should validate the IDs, reject an empty terminator set, and define what happens if an ID appears in both sets. I’d lean toward terminator behavior taking precedence.
- We should also document that terminators sit outside the constrained text, like EOS currently does. Structural special tokens can appear in the regex through `allowedSpecialTokenIds`, while a token configured as a terminator would normally be omitted from the regex itself.

With those details covered and a few tests around caching, early termination, and multiple terminators, this sounds like a useful follow-up.

nico-martin:
I renamed the package to `@huggingface/transformers-structured-output` and simplified the API around a `new StructuredOutputProcessor`.
I also removed the separate `stopping_criteria`, since generation can rely on the built-in EOS handling. Benchmarks across three models, three constraint types, and two context sizes showed no meaningful performance change.

The new API is now

```js
import { StructuredOutputProcessor } from "@huggingface/transformers-structured-output";

const processor = new StructuredOutputProcessor(tokenizer, {
  type: "json_schema",
  json_schema: {
    type: "object",
    properties: {
      answer: { type: "string" },
    },
    required: ["answer"],
    additionalProperties: false,
  },
});

await model.generate({
  ...inputs,
  logits_processor: processor,
});
```

xenova (approved):
excited for everyone to try this out!

PR #1762, opened 2026-08-31
Support custom file names for multi-session models

yushuosun:
Closes #1754.

`model_file_name` was only applied by single-session model configurations, so
encoder-decoder and other multi-session models always fell back to their
hard-coded file names.

This adds object mappings for multi-session models while preserving string
overrides for single-session models:

```js
model_file_name: {
  encoder_model: "custom_encoder",
  decoder_model_merged: "custom_decoder",
}
```

The resolved session map is now shared by the runtime loader, model registry,
dtype discovery, and pipeline progress-file discovery. Mapping keys may use
either the runtime session name or its default file name; unspecified sessions
retain their defaults.

Validation:

- Added unit coverage for mapped, partial, and existing string overrides
- Added a file-registry regression test for Seq2Seq models
- Ran isolated runtime-loader and file-registry assertions for the custom paths
- Ran syntax checks and Prettier checks over all changed JavaScript files

PR #1769, opened 2026-09-09
Support custom file names for multi-session models (refresh)

yushuosun:
This is a refresh of #1762 against the current main branch.


yushuosun:
Superseded by #1770.

